PR #9, opened 2015-04-09
Added holidays for New Zealand

linuxsoftware:
This adds New Zealand as another holiday country class.  Includes provincial anniversary days.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+
+
+- Added holidays for New Zealand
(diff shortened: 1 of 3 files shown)


linuxsoftware:
I will fix the flake8 errors.  But not tonight.

linuxsoftware:
There's another change I want to make, so I will close this request and open a new one when NZ holidays is ready.

PR #10, opened 2015-04-09
Added holidays for New Zealand

linuxsoftware:
This adds New Zealand as another holiday country class. Includes provincial anniversary days.


ryanss:
This is terrific, thank you!

PR #11, opened 2015-04-10
Allow '*' to fetch holidays for all New Zealand provinces. 

linuxsoftware:
* is a wildcard meaning get all holidays.


ryanss:
This is a really good idea, but I have a few concerns:

1. I'm not sure how I feel about the value being a list when multiple provincial holidays fall on the same date. If you are actually using the names of the holidays it would be nice to not need to test whether it is a string or list. What would you think about the values being a concatenated string instead of a list. Ex. instead of `datetime.date(2014, 12, 1): ['Westland Anniversary Day', 'Chatham Islands Anniversary Day']` it would be `datetime.date(2014, 12, 1): 'Chatham Islands Anniversary Day, Westland Anniversary Day'`?

2. Instead of having to go through each provincial holiday and adding the `'*'` option it might be better if you could just join all the provinces together using the overridden `__add__` method.
If each country had it's list of provinces stored in a constant you could do something like this to get the complete set: `sum([holidays.NZ(prov=p) for p in holidays.PROVINCES['NZ']])`
Another idea is to change the `if` statements to look something like `filter(re.compile(self.prov).search, ('WTC', 'WST', 'Westland', 'Greymouth')` which would allow you to use regular expressions in the `prov` string so to get the holidays from two different provinces you could do `holidays.NZ(prov="AUK|WTC")` and `holidays.NZ(prov=".*")` to get all of them.

Any thoughts?

linuxsoftware:
Yes, even as I wrote that code I was thinking it was ugly, and now I realize it is broken too (when holidays are added together).  Thanks for the suggestion of summing a list of provinces together to get a complete set of holidays for a country.  I have some code which I think is much cleaner based upon that.  I will submit it once I have some test cases working.  And thanks for this excellent library too!

linuxsoftware:
I have removed my province-wildcard code.  Instead I propose a couple of simple backwards compatible changes to HolidayBase:
1. There is new accessor get_names(<date>) which will return all the names of a holiday.  If two sets of holidays are added together and they have different names for the holiday on a certain date, then a call to get_names on the resulting HolidaySum will return both names.  (Is get_names a good name for this function?).  get and __getitem__ continue to work as before.
2. Similar to what you suggested, I've added a constant ALL_PROVINCES to the HolidayBase class of all the provinces (if any) an Holiday class supports.  So, now based on your suggestion I can go 
sum((holidays.NZ(prov=p) for p in holidays.NZ.ALL_PROVINCES), holidays.NZ()) to get an object with all the holidays for anywhere in New Zealand.
What do you think?

PR #13, opened 2015-10-05
Fix example script, importing date missing

mwharrison:


ryanss:
I agree the missing import should probably be added, but I think it should follow PEP8 standards.

```python
from datetime import date

import holidays

us_holidays = ...
```

PR #14, opened 2015-12-16
Adds German holidays

martinth:
This adds all German holidays for all provinces since it's unification in 1990.
There are some special edge cases since some province specific have complicated
sub-rules but those are document in the docstring of the class.


ryanss:
This is terrific, thank you for contributing!

PR #16, opened 2016-01-30
Colombian holidays

agmarrugo:
There are three kinds of Holidays in Colombia.
- Fixed dates holidays: They are always observed on the exact date and are never observed on a weekday if they happen to fall on the weekend; e.g. christmas, new year's eve, etc.
- Emiliani law holidays: These are mostly religious holidays that are always observed on the following Monday, unless they happen to fall on a Monday.
- Easter based: These are also religious holidays, but the date is based on Easter. Some like "Good Friday" always occur on a Friday, but others like Corpus Christi are observed on the following Monday, unless it falls on a Monday.

Observed feature supported
- For the fixed date holidays, if ``observed=True`` and they fall on a weekend they are not observed.
- If ``observed=False`` there are 18 holidays in total.

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -824,8 +824,8 @@ class UnitedStates(HolidayBase):
             self[date(year, 4, 22)] = "Arbor Day"
 
         # Primary Election Day
-        if self.state == 'IN' and ((year >= 2006 and year % 2 == 0)
-                                   or year >= 2015):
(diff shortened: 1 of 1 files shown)


agmarrugo:
No idea why it stills fails for Python 3.2.

ryanss:
This is great. Thank you!

PR #19, opened 2016-04-12
Explicit UTF-8 encoding for README.rst resolves ryanss/holidays.py#18

lukovkin:
Resolves installation issue under Python 3.5


ryanss:
Used a different solution that was compatible with Python2 as well.

 aa19794

PR #21, opened 2016-06-06
adding UK holidays

anthonykrose:
I've added the holidays (and associated tests) for the United Kingdom, as listed here:

https://en.wikipedia.org/wiki/Public_holidays_in_the_United_Kingdom

All of my tests pass but I do see a fail in the test_timestamp function (I see this independent of my changes though).

======================================================================
FAIL: test_timestamp (__main__.TestKeyTransforms)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests.py", line 357, in test_timestamp
    self.assertTrue(1388552400 in self.holidays)
AssertionError: False is not true

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -1607,3 +1607,91 @@ class Austria(HolidayBase):
 
 class AT(Austria):
     pass
+
+
+class UnitedKingdom(HolidayBase):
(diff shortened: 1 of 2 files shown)


anthonykrose:
I fixed this timestamp issue, will submit a second PR.

PR #22, opened 2016-06-07
adding UK holidays and enforcing UTC timestamp

anthonykrose:


ryanss:
This is excellent, thank you!

PR #23, opened 2016-06-07
Add Danish holidays and tests

thomasboevith:
On second thought, I think it would be more correct if I remove the holidays which are not officially real holidays: Grundlovsdag ("Constitution Day"), Juleaftensdag ("Christmas Eve's Day"), and Nytårsaftensdag ("New Year's Eve's Day"). On these days most stores and offices are closing but again, they are not real holidays.


ryanss:
Thanks!

PR #27, opened 2016-09-05
Pull version from `holidays.py` without importing it

MinchinWeb:
Fixes #5

This is designed as a more elegant fix than hardcoding the version of the library in `setup.py`. It read the version from `holidays.py` (using regex) without needing to import the file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ MANIFEST
 .coverage
 *.egg-info
 *.pyc
+env*/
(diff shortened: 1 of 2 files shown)


ryanss:
This would be useful if the `setup()` command was the only place with the version hardcoded but I've got it in the header comment of every file as well. I'll take a look at some popular python projects and see how they handle it. Thanks.

PR #28, opened 2016-09-05
Update Canadian Holidays

MinchinWeb:
Working mostly from this list from the CRA -- http://www.cra-arc.gc.ca/tx/hldys/menu-eng.html -- and details from Wikipedia


ryanss:
Awesome, thank you! I know it can be a lot of work to trudge through government web pages and wikipedia to find all the little details of each province. Much appreciated.

PR #29, opened 2016-09-18
TARGET holidays

sonntagsgesicht:
Added holidays for the European TARGET (Trans-European Automated Real-time Gross Settlement Express Transfer System).


ryanss:
I wasn't aware of official holidays for the whole EU, thanks! 905585bb296cf5cc9e2905dba743314b8a29083e

sonntagsgesicht:
Hi thanks, the holidays are rather the ECB (European Central Bank) than EU holidays. May be we should rename EU calls to ECB and/or EURO?

PR #30, opened 2016-09-20
renamed EU calendar to ECB 

sonntagsgesicht:
TAR holidays are published by European Central Bank.
Naming EU might be ambiguous.


ryanss:
Agreed, thanks.

PR #34, opened 2016-11-28
Adding PT Holidays

ajscarvalho:
Adding Extended Days that most people have as a bonus from their
company's or that take from the vacation days for theirselves


ryanss:
Looks good, thanks!

Issue #35, opened 2016-12-12
setup.py fails in non-utf8 locales


kvdveer:
When setup.py us run in non-unicode locales, it crashes with the error below. This is caused by the implicit use of codecs. To find the version number, holidays.py is read using the codecs module, but no codec is specified. Because of this, the codec used depends on the locale, which is ascii for non-unicode locales. Holidays.py contains many non-ascii characters in the holiday names, so this fails.

Non-unicode locales are the default in most minimal setups (such as docker images), so this has quite a bit of impact. Fortunately the fix is trivial; I'll add a merge request.

```
Collecting holidays (from -r requirements.txt (line 22))
  Downloading holidays-0.6.tar.gz (40kB)
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-a2dby65n/holidays/setup.py", line 21, in <module>
        fd.read(), re.MULTILINE).group(1)
      File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 16417: ordinal not in range(128)
```

Workaround for those who depend on the pypi version (requires the `C.UTF-8` locale to be generated on your system. Ubuntu docker image has that by default.)

```
LC_CTYPE=C.UTF-8 python setup.py install  # or whatever pip command you need
```

PR #36, opened 2016-12-12
Specify encoding when parsing for version in holidays.py

kvdveer:
This allows the package to be installed in non-unicode locales.


kvdveer:
This should close issue #35

ryanss:
Good catch. Thank you!

PR #41, opened 2017-02-05
Add Norwegian holidays

gablank:
In Norway, all Sundays are considered holidays. I can see that a lot of users would want Sunday to not be considered a holiday, so I added a parameter to Norway.\_\_init\_\_ (*include_sundays*) to allow them to control whether they should be considered a holiday or not. Currently, it defaults to True.

I've also used asserts in the code to check that the days of easter always fall on the same weekdays. I believe this is good practice, as something is horribly wrong (for example in dateutils.easter) if this is not true.


ryanss:
This looks excellent, thanks! Great test coverage.

PR #43, opened 2017-03-20
Changes to Australia holiday class

linuxsoftware:
I am the one who created the Australia country class with it taking a prov argument.  Now that the library is much more developed and HolidayBase can also take a state argument I think Australians would prefer to use that (even though ACT and NT are territories, not states).  I have changed this, and have also removed ACT as a default, and fixed the handling of Canberra Day and Family & Community Day for ACT.

NOTE:  I have not made any attempt at maintaining backwards compatibility.  Is this required?  If so I could add some code for that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ MANIFEST
 .coverage
 *.egg-info
 *.pyc
+.DS_Store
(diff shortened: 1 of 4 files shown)


linuxsoftware:
Sorry I should have done the PR as a branch

PR #46, opened 2017-03-29
Add Polish holidays

tadeoos:
Hey!
I am adding support for Polish holidays.

Docstring and implementation for including Sundays as Holidays was copied from the Norway class (I hope this won't be considered as intelectual theft or anything, if so, please let me know I'll change that). The above results in having (twice) of around 11 lines of much the same code, so maybe there could be a base class method for that in the future?

This is my first open source contribution so I am quite excited ;) 
All the best!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ MANIFEST
 .coverage
 *.egg-info
 *.pyc
+.DS_Store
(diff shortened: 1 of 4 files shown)


dr-prodigy:
Merged in **1452cc3**

PR #54, opened 2017-05-14
Add one-off national Reformation day (DE/2017)

willhardy:
This library is missing a national holiday in Germany that is coming up soon.

Germany is celebrating the 500th anniversary of Reformation day by making it a national public holiday on 2017-10-31. Normally it is only a public holiday in a few states. See eg https://en.wikipedia.org/wiki/Reformation_Day#500th_anniversary


dr-prodigy:
Superseded by **1452cc3**

PR #61, opened 2017-06-27
Add Polish holidays

mbartosi:
Added Polish national holidays.


mbartosi:
Please use earlier contribution.

PR #78, opened 2017-10-29
Update Python versions: add 3.6, remove EOL

hugovk:
Fixes https://github.com/ryanss/python-holidays/issues/56.

Python 3.2 is causing build problems (#56) and it's been decided it's easiest to drop it as it's EOL and no longer receiving security updates.

Now is also a good time to add Python 3.6 and remove the other EOL versions (2.5, 2.6, 3.2 and 3.3), which allows the code to be modernised.

Here's the pip installs for holidays from PyPI for the last month:
```
$ pypinfo --percent --pip holidays pyversion
python_version percent download_count
-------------- ------- --------------
2.7              70.7%         11,946
3.6              19.7%          3,332
3.5               7.3%          1,239
3.4               2.2%            376
2.6               0.0%              3
3.3               0.0%              1
```

Those versions are barely used if at all.

PR #80, opened 2017-10-29
Fix E722 do not use bare 'except'

hugovk:
Fixes https://github.com/ryanss/python-holidays/issues/77 and fixes the flake8 build.

I guess the general idea of the warning is to catch more specific exceptions, but it can be hard to know what the intention was with old code so this is fine. In the future let's be more specific with new ones.

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -68,7 +68,7 @@ class HolidayBase(dict):
         elif isinstance(key, six.string_types):
             try:
                 key = parse(key).date()
-            except:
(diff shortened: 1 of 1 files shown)


hugovk:
Note: the 3.2 build still fails, see https://github.com/ryanss/python-holidays/issues/56, will be fixed by #78.

dr-prodigy:
Fix included in **1452cc3**

PR #82, opened 2017-11-17
Added italian holidays

dr-prodigy:
Added support for italian holidays. It could be completed with more regional holidays (province saints, which I provided for Milan and Rome) but the most important part of the work is there.


dr-prodigy:
Working on a better commit: closing this PR.

PR #87, opened 2017-12-22
Add Finnish holidays

Ardetus:


dr-prodigy:
Please check and re-submit. Thx!

Ardetus:
I agree that eves are kind of holidays in Finland and day offs for some people, but I understood that this library was for counting holidays, not day offs or flag days.

[Public holidays in Finland](https://en.wikipedia.org/wiki/Public_holidays_in_Finland)

hugovk:
Flag days are definitely not eligible.

I think those two eves are important enough to be listed as holidays. That Wikipedia page agrees:

> Christmas Eve and Midsummer Eve might very well be the single most important holidays during the entire year for Finns. Surprisingly they are not officially called holidays and are not so marked in calendars, but for most people, are not working days, and in practice they differ from official holidays only in that most shops are open on those days from early morning till noon. They hold this de facto status partly due to some statements in legislation but also because most employment contracts provides for these days as full holidays.

And lists them:

image


From a practical point of view, they should be included.

Likewise, even though "All Sundays" are official holidays, from a practical point of view they should not be included.

Ardetus:
It's true that those eves are even more important holidays, than some of official holidays for majority of Finns. I just made new PR including those two eves.

PR #92, opened 2018-01-14
Drop support for EOL Python 2.5, 2.6, 3.2 and 3.3

hugovk:
Rebased version of https://github.com/ryanss/python-holidays/pull/78.

This removes those Python versions which are EOL (2.5, 2.6, 3.2 and 3.3) and no longer receiving security updates, and are little used, and then update the code to ease the maintenance burden and make use of more modern Python.

Here's updated pip installs for holidays from PyPI for the last month (via `pypinfo --percent --pip --markdown holidays pyversion`):

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   76.0% |         16,846 |
| 3.6            |   18.9% |          4,194 |
| 3.5            |    3.6% |            796 |
| 3.4            |    1.5% |            332 |
| 2.6            |    0.0% |              2 |
| 3.7            |    0.0% |              1 |

Of the EOL versions, this shows only two installs for 2.6, and nothing for 2.5, 3.2 and 3.3.


dr-prodigy:
Okay sorry, at the beginning I couldn't understand your point (I've faced some library building issues related to old Python versions adoption, which I then fixed, so I thought yours was just an alternative way of fixing)!
I agree with you, Python EOL versions can be abandoned; on the other hand it's not really a urgent task (especially because code supports them now), so I'm going to merge your rebased PR for the next release.

Thank you very much, cheers!
dr.p

PR #97, opened 2018-01-21
Various inspection fixes

hugovk:
Code cleanup:
* Remove redundant parentheses
* Simplify chained comparison
* Make variables lowercase
* Fix typos

List/dict upgrades:
* Rewrite unnecessary generator as a dict comprehension
* Rewrite unnecessary list call as literal
* Replace unnecessary list comprehension - 'sum' can take a generator

Tests:
* Update asserts to give descriptive feedback instead of only using `assertTrue`/`assertFalse`
* Fix some broken tests: `"name"` in `assertTrue(holiday, "name")` is just a message shown on failure, and will never fail. These should really be `assertEqual(holiday, "name")` which can fail.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -164,12 +164,12 @@ Methods:
 get(key, default=None)
     Returns a string containing the name of the holiday(s) in date `key`, which
     can be of date, datetime, string, unicode, bytes, integer or float type. If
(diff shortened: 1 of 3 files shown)


dr-prodigy:
Thank you Hugo!

PR #136, opened 2018-09-24
add prov/state parameters to CountryHoliday()

### LINKED ISSUE #128: setting state/province using CountryHoliday
Assuming that, `CountryHoliday` is used as described in the README and the country is not know until runtime (e.g. it is read from a config or user input):

```python
import holidays

country = "DE"  # from sys.argv, for example
ch = holidays.CountryHoliday(country)
```

This of course yields a perfectly fine instance. The problem comes up when the state or province is also relevant. Since `CountryHoliday` returns an instance, there is no way to add a state, which is set using the constructor.

A couple of workarounds:

* use `type()` or `.__class__` to get the class; create a new instance
* use `getattr(holidays, country)` to get the class directly, skips the nice error handling

Neither of those seem like a sleek and easy to understand solution to me.

To resolve this, I propose that all keyword arguments to `CountryHoliday` should be passed onto the class, like so (omitting error handling):

```python
def CountryHoliday(country, **kwargs):
    country_holiday = globals()[country](**kwargs)
    return country_holiday

holidays_de_he = CountryHoliday("DE", prov="HE")
```

This enables setting the state/province and of course all other options like `expand` or `observed`, even when using the very friendly `CountryHoliday`. The external API stays stable, since arguments are only added and the single existing one just stays in place.

What do you think? :) I am happy to code up a quick PR including docs and a test, if you'd like.

luto:
fixes #128 

see issue for details :)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,10 @@
+Version 0.10.0
+==============
+
+Released ?
+
+- Added state and prov parameters to CountryHoliday() method
+
 Version 0.9.7
 =============
 
--- a/README.rst
+++ b/README.rst
@@ -41,6 +41,12 @@ Example Usage
 
     us_holidays.get('2014-01-01')  # "New Year's Day"
 
+    us_pr_holidays = holidays.UnitedStates(state='PR')  # or holidays.US(...), or holidays.CountryHoliday('US', state='PR')
+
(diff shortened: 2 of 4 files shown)


luto:
Thank you :)

PR #144, opened 2018-10-19
remove magic numbers

elln2:
I noticed that there are many "magic" numbers used throughout the library. Initially, I considered it awkward that days of the week would be declared and used throughout some areas of the library but not others. I've went through and found as many locations as I could where a numeric day of week was used, and replaced it with its corresponding weekday variable.

Additionally, I mirrored this behaviour for months, and replaced numeric month values with the respective month variable.

I expect that this was started for days of the week to reduce the risk of errors due to confusion between python's `weekday` and `isoweekday` methods, and hope to avoid a similar confusion with months of the year.


elln2:
I may have figured out why months are kept as integers! Apologies for not observing the preferred 80-character line length limit. While I've fixed this style violation, I understand if it would be preferred to stick with month digits to improve code style.

dr-prodigy:
Hi,
thank you for your contribution!

Even if project overall code style has not been set by me (first releases are by @ryanss ), until now I've never really felt a need for a change on that (when talking dates, simple numbers should not probably considered "magic".. instead they are quite self-explanatory, and moreover require minimum space to store code, which in interpreted languages like Python has its weight).

That said, I still agree with you that widespread constant usage brings structure and tidiness to code (in fact, if it had been me, I admit I would have probably limited this intervention to days-of-week only, but still your work is very considerable), so I'm going to adopt all of your changes.

It will take a while to merge it, because the huge number of changes could possibly lead to "insidious" regressions, so your new code has to be checked thoroughly.
Anyway, not being a feature upgrade, it won't be a big issue to delay it a little I think.

Thank you, cheers!

elln2:
Completely understandable. Even though all the tests passed, I'm not sure of any possible side effects or "gotchas". I'm more than happy to provide any stylistic or other changes!

PR #146, opened 2018-11-12
Fix Brazil's Carnaval holiday

victorpluna:
The Carnaval holiday on Brazil should is on Tuesday before "Quaresma".


dr-prodigy:
Well done.. thank you very much.
Cheers!

Issue #110, opened 2018-05-14
Easter Sunday is a holiday in all of Germany, not only in Brandenburg.


Achimh3011:
For some reason I don't understand `python-holidays` only defines Easter Sunday for `country == 'DE'` as holiday in `prov = 'BB'` which I guess means _Brandenburg_. Easter Monday and Good Friday are (correctly) set up for entire Germany. 

I'm not aware of any special regulation in _Brandenburg_, so I'd say this is a bug. Of course, Easter Sunday is always a Sunday, so the consequences aren't that great from a work time perspective, but anyhow...

P.S.: https://www.youtube.com/watch?v=kdwKbiMyvu0

P.P.S: Same issue with Pentecost (_Pfingstsonntag_)...

dr-prodigy:
Thank you: it will be fixed in the next release.

dr-prodigy:
Released in v.0.9.6. Thank you

karlsebal:
This is not true. Easter Sunday and Whit Sunday are Holidays only in BB (Berlin-Brandenburg). For HE (Hessen) every Sunday is a holiday.

https://de.wikipedia.org/wiki/Gesetzliche_Feiertage_in_Deutschland

PR #149, opened 2018-11-19
fix German holidays

vlt:
In #110 a problem with the German public Easter and Pentecost holidays was introduced.

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -2023,10 +2023,11 @@ class Germany(HolidayBase):
 
             self[easter(year) - rd(days=2)] = 'Karfreitag'
 
-            # will always be a Sunday and we have no "observed" rule so
-            # this is pretty pointless but it's nonetheless an official
-            # holiday by law
-            self[easter(year)] = 'Ostersonntag'
+            if self.prov == "BB":
+                # will always be a Sunday and we have no "observed" rule so
+                # this is pretty pointless but it's nonetheless an official
+                # holiday by law
+                self[easter(year)] = "Ostersonntag"
 
             self[easter(year) + rd(days=1)] = 'Ostermontag'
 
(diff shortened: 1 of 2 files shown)


Achimh3011:
No, the problem was fixed. What gives you the idea that Easter and Pentecost are not holidays in entire Germany?

vlt:
> I'd still suggest to keep them as holidays for the entire
> Germany.

You might suggest that to the legislators ;-)
Currently only in the "BB" province these are public
holidays and -- by law -- paid as such, for example.

> Regarding the Reformationstag, you can see the full list
> at https://www.ferienwiki.de/feiertage/reformationstag -
> so the current list is not long enough.

I don't see which region would be missing. Can you specify
that?

vlt:
I moved the "Reformationstag" issue to a separate PR #150.

Achimh3011:
Well, Sundays are paid the same as Holidays, that's why they are not listed in the other federal states. So in terms of workday/free day it does not make a difference. I'm still wondering why there is an explicit definition only in this one state.

If you are looking for holidays to be generated (so some kind of calendaring), it does make a difference. Maybe a differentiation is needed?

dr-prodigy:
Hi, 
thank you both for your contribution.
Unfortunately, I don't read german and have no specific knowledge about holiday laws in DE provinces, so I can't really choose which point of view is best (still, at first thought, I would agree with @Achimh3011 that a Sunday can be considered holiday anytime with no particular harm :) ).
@vlt : given that there is no practical difference on salaries or work contracts (or so Achimh says), why is RS excluded explicitly from holidays in some provinces (or, the largest part of them)? If you could explain this, I'd surely consider to merge your PR (which btw is technically 100% flawless).
Thanks again to both.

Achimh3011:
I dug up an article on this weirdness: if you are forced to work on Pentecoast Sunday you will get normal Sunday premium (50%) where as on Pentecoast Monday you will get the Holiday premium (125%). Easter should be the same, but is not mentioned in the article.

> Wo ist er gesetzlicher Feiertag?
> 
> Einzig das Bundesland Brandenburg macht eine Ausnahme: Hier gilt der Pfingstsonntag als gesetzlich anerkannter Feiertag. Das wirkt sich in diesem Fall tatsächlich aber kaum aus, da Sonntage gesetzlich ohnehin mit Feiertagen gleichgestellt sind.
> 
> Einen wichtigen Unterschied gibt es allerdings: Müssen Sie am Pfingstsonntag arbeiten, bekommen Sie laut eines Urteils des Bundesarbeitsgerichts aus dem Jahre 2011 keinen sogenannten steuerfreien Feiertagszuschlag – es sei denn, Sie arbeiten in Brandenburg. Der Zuschlag beträgt dann 125 Prozent. Der normale Sonntagszuschlag liegt dagegen nur bei 50 Prozent.

https://www.t-online.de/leben/familie/id_69224230/warum-der-pfingstsonntag-kein-gesetzlicher-feiertag-ist.html

dr-prodigy:
Merged.. thx to both! 👍

PR #154, opened 2018-11-28
Fixed one more forgot bug in Canada Calendar

justinasfour04:
There was one more line of code that I forgot to change that needed fixing


justinasfour04:
@dr-prodigy I forgot one fix for Canada

PR #158, opened 2019-01-02
Adding once-off and historic holidays for South Africa

Gordonei:
# Overview
This PR makes two changes to the library:
1. Adds once-off holidays which have occurred over the past 20 years. These are mostly various elections, which are made holidays so as to make it easier to vote.
2. Adds historic holidays which are no longer observed. The current holidays are also time-bounded, so that they don't get returned in years that they shouldn't.

# Testing
## Automated
* Linting
* Unit Tests (including new ones)
## Ad-hoc
Testing workers day in 2018 vs 1988:
```python
import holidays

za_holidays = holidays.SouthAfrica()

for date in ('2018-05-01', '1988-05-01', '1988-05-06'):
    print(date in za_holidays)
```
result:
```
True
False
True
```


dr-prodigy:
All done.. Thank you very much! ;-)

Issue #113, opened 2018-06-11
pull date range


samuel-1985:
Hi

I am a newbie to this library and I tried to find the holidays in a date range that started from date A to date B in Canada.
 for example, this is what I would do with the U.S. holidays:

from pandas.tseries.holiday import USFederalHolidayCalendar as calendar
cal = calendar()
holidays = cal.holidays(start='05-13-2014, end=05-13-2016)
print(holidays)

I would appreciate if you help me with that.

Regards

philwech:
I solved the problem outside. But it could be a nice function for the library.

def holidays_from_range(lower_date, upper_date, prov):
    # smallest and biggest year
    smallest_year = lower_date.year
    biggest_year = upper_date.year
    # list_holidays has all holidays
    list_holidays = sorted(holidays.CH(prov=prov, years=range(smallest_year, biggest_year+1)))
    df_holidays = pd.DataFrame(index=list_holidays)
    # return just the holidays between lower- and upper-date
    df_holidays = df_holidays[lower_date:upper_date]
    ls_holidays = []
    for row_index, row in df_holidays.iterrows():
        ls_holidays.append(row_index.strftime('%Y%m%d'))
    return ls_holidays

dr-prodigy:
I will release this solution after some more analysis: thank you very much for your contribution.

elln2:
Hey @dr-prodigy I've got an idea to tackle this, mind if I take a shot and get some feedback if you haven't had a chance to get started?

dr-prodigy:
Hey elln2,
in fact unfortunately I couldn't work on this until now, but still it's an interesting feature.
So, of course, I'll surely be happy to consider/include your work, if you want to contribute in any way.

Thank you very much, cheers!

dr-prodigy:
Thank you very much for your contribution @elln2 .. much appreciated! 👍

PR #164, opened 2019-01-06
Holidays in date range

elln2:
RE #113 Allow users to get the dates for which any holiday falls within a date range. Implemented using slice notation; users simply give [start, stop, step], where step is optional. Sample usage:

```
>>> holidays.US()[date(2013, 12, 31): date(2014, 1, 2)]
[datetime.date(2014, 1, 1)]
>>> holidays.US()[date(2014, 1, 1): date(2013, 12, 24)]
[datetime.date(2014, 1, 1), datetime.date(2013, 12, 25)]
>>> holidays.CA()[date(2013, 12, 25): date(2014, 1, 2): 3]
[datetime.date(2013, 12, 25)]
```

Uses the existing `__keytransform__` functionality so users may use dates, datetimes, strings, etc. Step field may be given as a positive or negative integer or timedelta (only days interpreted), or use default step of one day.

Infinite sequences, such as `[date(2014, 1, 1):]` are not implemented and are protected against, as the need was not previously described, and the use case seems limited. If this use case would be useful to have, further implementation of such a generator could be done.

Let me know if you'd like any further tests or tweaks!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,6 +47,8 @@ Example Usage
 
     us_holidays.get('2014-01-01')  # "New Year's Day"
 
+    us_holidays['2014-01-01': '2014-01-03']  # [date(2014, 1, 1)]
+
     us_pr_holidays = holidays.UnitedStates(state='PR')  # or holidays.US(...), or holidays.CountryHoliday('US', state='PR')
 
     # some holidays are only present in parts of a country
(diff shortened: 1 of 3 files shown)


dr-prodigy:
Thank you very much! Very good work! 👍

PR #172, opened 2019-01-11
Update README.rst

mashawings:
add Russia


dr-prodigy:
Already there with Russia code addition. Thx

PR #186, opened 2019-03-04
Make Czech deprecated, introduce Czechia instead

janpipek:
It should work like this.

You're usually testing against the abbreviated names, e.g. CZ, IT, ... which leaves the test unchanged :-)

I am still having one issue - if I put the semantically correct deprecation warning in the code, it is by default not printed, as Python ignores it (see https://docs.python.org/3.6/library/warnings.html#default-warning-filters ).

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -2672,7 +2672,8 @@ class CZ(Czechia):
 
 class Czech(Czechia):
     def __init__(self):
-        warnings.warn("Czech is deprecated, use Czechia instead.", DeprecationWarning)
+        warnings.warn("Czech is deprecated, use Czechia instead.",
+                      DeprecationWarning)
         super(Czech, self).__init__()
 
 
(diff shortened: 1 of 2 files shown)


janpipek:
One suggestion regarding code brevity:

Why do you inherit the "abbreviated" classes like `class IT(Italy): pass` and not just assign `IT = Italy`? It is not just shorter, but perhaps even semantically cleaner...

dr-prodigy:
> One suggestion regarding code brevity:
> 
> Why do you inherit the "abbreviated" classes like `class IT(Italy): pass` and not just assign `IT = Italy`? It is not just shorter, but perhaps even semantically cleaner...

Hi,
that's just sth which was like that when I started maintaining the prj.. simple like that :-)

PS: very well done! Thank you very much 👍

janpipek:
You are welcome.

PR #191, opened 2019-04-04
Add Japanese holidays on the new Emperor

kokinomura:
- Add new holidays regarding the ceremonies of the new Emperor in Japan

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -3244,10 +3244,14 @@ class Japan(HolidayBase):
         # Foundation Day
         self[date(year, FEB, 11)] = "建国記念の日"
 
+        # Reiwa Emperor's Birthday
+        if year >= 2020:
+            self[date(year, FEB, 23)] = '天皇誕生日'
+
         # Vernal Equinox Day
         self[self._vernal_equinox_day(year)] = "春分の日"
 
-        # The former Emperor's Birthday, Greenery Day or Showa Day
+        # Showa Emperor's Birthday, Greenery Day or Showa Day
         if year <= 1988:
             self[date(year, APR, 29)] = "天皇誕生日"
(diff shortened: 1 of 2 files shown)


dr-prodigy:
Hi,
thank you very much for your contribution!

Question about it: I gave a look to your code, and noticed you've provided logic for 2019 only.
To explain my lifecycle principle, I use to say "we aim always to providing a final release library, even if we already know it won't be so".. so isn't there the possibility to manage/support future years too (eg: `if year **>=** 2019:` ) ?
Of course, if something in the future will change, there will be a new release to fix it.. and so on :-)

I rely on you for this because I'm not aware of how holidays in Japan work.. thanks again!

NOTE: I moved the target branch to "beta", which is the branch we use to include new PR.

kokinomura:
The two sections with `if year == 2019` are both related to the change of the Emperor, which is planned to perform this year, and there will no holidays like them from next year. I'm sorry I didn't explain enough with comments, but the latter section, which has two holidays,  is also related to Enthronement Day. These days become holidays only this year.

kokinomura:
Let me also explain some hard-coded functions.

`_add_national_holidays` and `_add_substitute_holidays` can be calculated with programs because they are established based on a series of weekdays and holidays, (a weekday between national holidays becomes a holiday, and a Monday after a national holiday on Sunday becomes a holiday), but the code can be long and these holidays are changed by Government sometimes. This is why I chose to hard code.

dr-prodigy:
Hey no problem at all, I just wanted to double check .. moreover I noticed now there's a `if year >= 2020:` block, so it's my fault, in the end ;-)

..Going to merge, thank you very much again!

PR #202, opened 2019-05-31
Added Iceland Holiday Support

justinasfour04:
Added Iceland holiday support and ran Black formatter on holidays.py

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -4433,3 +4433,32 @@ class Estonia(HolidayBase):
 
 class EE(Estonia):
     pass
+
+
+class Iceland(HolidayBase):
+    # https://en.wikipedia.org/wiki/Public_holidays_in_Iceland
+    # https://www.officeholidays.com/countries/iceland/index.php
+    def __init__(self, **kwargs):
+        self.country = "IS"
+        HolidayBase.__init__(self, **kwargs)
+
+    def _populate(self, year):
+        # Public holidays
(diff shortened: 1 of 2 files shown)


dr-prodigy:
Hi @justinasfour04, thank you for your contribution.

I've made a review on this new work of yours but, even if I consider it very good, and fully understand the pros of code auto-formatting (it makes some sense), I have to point out that:
* even if the new code often looks cleaner and more standard, we've lost some code format "intention" and readability on the way, in some places
* the huge number of not "crucial" changes would over-complicate future analysis ( = "broken" history)
* finally, to get the most out of your effort, we should probably adopt black formatter as a standard step for all future developments / commits (like we do for flake8 and automatic tests, to explain), in order to carry on the same code format. Unfortunately, black is available under Python 3 envs only, and this makes the resulting process trickier / less obvious (holidays is 100% Py2 compatible too, so its whole development / testing process should be too IMHO)

For these considerations, I think we have to (with some regret, I assure) leave the re-formatting task in standby for now, until at least some of these issues are removed / mitigated.

Do you think you'll be able to re-pack your modifications into a new PR excluding the format refactoring (some git cherry-picking should suffice)?
If needed, I can help you out with it.

Please let me know, thanks again
dr-p

justinasfour04:
@dr-prodigy I created a new pull request and I completely understand your point of view

PR #208, opened 2019-06-27
Add support for Hong Kong

polyfaces:
Add support for Hong Kong, and the code should be future proof. 

Some codes are a little bit complicated as it requires conversions from Lunar calendar to Gregorian calendar.

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -4353,3 +4353,309 @@ class Lithuania(HolidayBase):
 
 class LT(Lithuania):
     pass
+
+
+class HongKong(HolidayBase):
+
+    #https://www.gov.hk/en/about/abouthk/holiday/2020.htm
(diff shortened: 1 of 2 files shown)


dr-prodigy:
Hi,
first of all thank you very much for your rich contribution!

I took a look at it, and have some point of refinement to ask you before merging:
- its base branch was master: I switched it to beta but now it requires resolving some conflicts
- there is a very large number of PEP-8 issues (pls refer to flake8 usage: https://github.com/dr-prodigy/python-holidays#running-tests for checking). These should be fixed, otherwise build phase will fail
- tests are not very well structured (a large number of repeated blocks). I would suggest to re-engineer it in a more maintainable way (eg: tests for Germany)

Please let me know what you can do, thank you very much

polyfaces:
Thanks for your comments.  I have fixed the issues and re-opened another pull request.

PR #215, opened 2019-07-11
Make Polish deprecated, introduce Poland instead

nicoa:
stumbled upon this when trying to use and thought it should be aligned with #186 .
What do you think?

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -2861,7 +2861,7 @@ class Poland(HolidayBase):
 
 
 class Polish(Poland):
-    def __init__(self):
+    def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


dr-prodigy:
Hi
thank for your work.
Unfortunately all your changes/fixes are already implemented and release under "beta" branch.

I would ask you to upgrade your repo to it, and possibly double-check everything is already there as expected.

Thanks again, cheers
dr-p

nicoa:
Thanks, will do!

nicoa:
Now it's only aligning the deprecated class' super call with czech and slovak

PR #219, opened 2019-08-08
Correctly handle United Kingdom "May Day" holiday in 2020.

robfraz:
This holiday is normally on the first Monday in May, but in 2020
it's being moved to Friday 8th May in order to mark the 75th
anniversary of VE Day.  This is as published on the following
official government webpages for England and Wales, Scotland, and
Northern Ireland:

 * https://www.gov.uk/bank-holidays#england-and-wales
 * https://www.gov.uk/bank-holidays#scotland
 * https://www.gov.uk/bank-holidays#northern-ireland

However, note that Ireland (which also inherits from the
UnitedKingdom class in a not very politically correct manner)
seems to be following the normal 'first Monday in May' rules in
2020 to the best of my knowledge at time of writing:

 * https://publicholidays.ie/2020-dates/

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -2457,24 +2457,28 @@ class UnitedKingdom(HolidayBase):
         # May Day bank holiday (first Monday in May)
         if year >= 1978:
             name = "May Day"
-            if year == 1995:
-                dt = date(year, MAY, 8)
+            if year == 2020 and self.country != 'Ireland':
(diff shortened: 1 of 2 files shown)


robfraz:
Force pushes above were purely to correct two typos in the commit message...

dr-prodigy:
Good work, thank you very much!
Going to merge it.

PR #230, opened 2019-09-09
Support nigeria

ioluwayo:
Adds support for fixed holidays in Nigeria


dr-prodigy:
Thank you very much mate! 👍

PR #231, opened 2019-09-11
Telangana State (TS ) was added

kiranbeethoju:
Added Telangana state(TS) and its holidays.


dr-prodigy:
Thank you very much! 👍

PR #233, opened 2019-09-16
Hungarian calendar is flavoured by bit of history & swapped days off.

gypapp:
In the last century, Hungary switched its governmental system several
times backed by drastically different ideologies and, of course, it came
with change of officially endorsed holidays, too. I also did my best to
figure out which currently endorsed national holiday had been introduced
in which year.

In the last couple of decades, another trend emerged, the swap days
around official national holidays. Basically, if a holiday falls on
Tuesday or Thursday, that is, only one weekday is between the holiday
and the closest weekend, that weekday is a day off work usually. These
days are also captured if observed is True.

So far so good but these rest days come at cost. Saturday within at most
2 week distance of that weekday will be a working day. I do not know how
to represent those days in this package because, to my best knowledge,
`holidays` has no notion about such a construct. Anyway since large
number of people do not work on those swap days either the current setup
is good enough.

Unfortunately, there is a hidden caveat with swap days. Nowadays, one
may count on them for sure but technically those are regulated by yearly
governmental edicts, renewed year by year.

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -4129,6 +4129,11 @@ class HND(Honduras):
 
 class Hungary(HolidayBase):
     # https://en.wikipedia.org/wiki/Public_holidays_in_Hungary
+    # observed days off work around national holidays in the last 10 years:
+    # https://www.munkaugyiforum.hu/munkaugyi-segedanyagok/2018-evi-munkaszuneti-napok-koruli-munkarend-9-2017-ngm-rendelet
+    # codification dates:
+    # - https://hvg.hu/gazdasag/20170307_Megszavaztak_munkaszuneti_nap_lett_a_nagypentek
+    # - https://www.tankonyvtar.hu/hu/tartalom/historia/92-10/ch01.html#id496839
 
     def __init__(self, **kwargs):
         self.country = "HU"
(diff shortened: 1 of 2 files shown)


gypapp:
@dr-prodigy should I introduce a separate flag for swapped days in Hungary calendar or we could reuse/abuse `observed` flag for that purpose? (See my explanation about swapped days in the PR description.) What do you think?

dr-prodigy:
Hi @gypapp thank you very much for your brilliant work!
I must admit these hungarian rules are really peculiar ..
Given that this is going to be a singular case, yes, for simplicity's sake I would control this behaviour with observed flag.

Don't hesitate to ask for any support.. thanks again!

gypapp:
> Given that this is going to be a singular case, yes, for simplicity's sake I would control this behaviour with observed flag.

OK, let's leave it as is now then. I just asked it because I saw unique flags in the implementation and did not know what the overall intent is.

> Don't hesitate to ask for any support.. thanks again!

Just one more thing. I do not see the test coverage report in this PR and don't know the reason why. Can you check it and let me know when you are willing to release it.

dr-prodigy:
Hi @gypapp ! Just merged your work.

In the meanwhile, I noticed some PEP8 / tests garbage left on the way by someone before you :) that's probably why you were not seeing test coverage.

Anyway, all fixed and back to 100% coverage.

Thank you very much, cheers! 👍

PR #248, opened 2019-11-27
Deprecated Czech class was not backwards compatible.

dfrank8:
Hello! Wonderful library, thanks so much for maintaining it! I found that the **Czech** class (although deprecated) was not backwards compatible. It's definitely not a huge issue, but I was forced to develop a workaround for this in my use-case, so I figured I'd create a PR for it. 

Deprecated Czech class was not backwards compatible by failing to pass `**kwargs` through to parent's `__init__`.

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -2766,7 +2766,7 @@ class Czech(Czechia):
     def __init__(self, **kwargs):
         warnings.warn("Czech is deprecated, use Czechia instead.",
                       DeprecationWarning)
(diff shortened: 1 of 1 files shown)


dr-prodigy:
Very good point, and thank you for reaching out with your fix!
I'm releasing it on beta branch (the branch dedicated to new features / fixes before PyPi releases), and applying it to the other same cases (ie: Polish, Slovak).

Thanks again, cheers
dr-p

PR #257, opened 2019-12-28
Belarus (BY): add January, 2nd as a national holiday

mpolyakovsky:
Hurray!

From 2020 the second of January is the new national holiday in Belarus (New Year) according to the [presidential decree #464](http://president.gov.by/uploads/documents/2019/464uk.pdf):

"New Year - 1 **and 2 of January**"


dr-prodigy:
Merged into beta.. Thank you very much!
Cheers
dr-p

PR #263, opened 2020-01-08
Added support for Greece.

superconvergence:
Hello!
This is a pull request to add support for Greek holidays.

### DIFF
--- a/holidays.py
+++ b/holidays.py
@@ -5552,9 +5552,6 @@ class Greece(HolidayBase):
         # Independence Day
         self[date(year, MAR, 25)] = "Εικοστή Πέμπτη Μαρτίου [Independence Day]"
 
-        # Good Friday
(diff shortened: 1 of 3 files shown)


holidays.py
```diff
@@ -5528,3 +5528,60 @@ class RS(Serbia):
...
+        self[date(year, JAN, 6)] = "Θεοφάνεια [Epiphany]"
+
+        # Clean Monday
+        self[eday - rd(days=48)] = "Καθαρά Δευτέρα [Clean Monday]"
+
+        # Independence Day
+        self[date(year, MAR, 25)] = "Εικοστή Πέμπτη Μαρτίου [Independence Day]"
+
+        # Good Friday
+        self[eday - rd(days=2)] = "Μεγάλη Παρασκευή [Good Friday]"
```
ppanagiotis: This isn't an official holiday, unless you are working at public services.

superconvergence: Thanks, fixed.

PR #268, opened 2020-01-13
Fixed month for Día de Asturias

jbroudou:
It appears from https://www.boe.es/diario_boe/txt.php?id=BOE-A-2019-14552 that in Asturias, "Día de Asturias" is not on 8th March, but instead is on 8th September.


dr-prodigy:
Project structure has changed, so I had to re-do this fix. Released on beta: closing.

Thx, cheers!
dr-p

PR #272, opened 2020-01-18
Adding Vanilla Support for Egypt

gaberm:
The PR is meant to add support for Egypt.
Key notes:
* Weekend Days are (Friday, Saturday)
* The module adds vanilla support for Egypt Holidays 
* Support for shifting holidays due to weekends are there but not active since there is no formal rule.


dr-prodigy:
Thank you very much! 👍

PR #289, opened 2020-02-11
Corrected some misspellings and one string that was still in English

reinaldoramosarxi:
A few of the strings were misspelled and "Christmas Day" was still in English.


dr-prodigy:
Thank you very much, cheers 👍

PR #302, opened 2020-04-19
Add example for listing holidays of multiple years

joe42:
Add an example for listing all holidays in a range of multiple years.


dr-prodigy:
Thank you!
dr-p

PR #304, opened 2020-04-23
Updates netherlands.py

RooieRakkert:
Adds Good Friday to the Dutch holidays. 
Good friday is always the first friday prior to easter.


dr-prodigy:
Thank you! 👍

PR #315, opened 2020-05-13
Fixes for Switzerland GE

cgrigis:
A couple of fixes for the Geneva canton in Switzerland:
* "Berchtoldstag" is not a holiday in Geneva.
* "L'Escalade", while a celebration in Geneva, is not a holiday.
* Unlike the rest of Switzerland where the Federal fast is a holiday (3rd Sunday in September), Geneva has the Genevan fast on the Thursday following the 1st Sunday in September.

References:
https://en.wikipedia.org/wiki/Public_holidays_in_Switzerland
https://www.ge.ch/legislation/rsg/f/s/rsg_j1_45.html (in French)

### DIFF
--- a/holidays/countries/switzerland.py
+++ b/holidays/countries/switzerland.py
@@ -33,7 +33,7 @@ class Switzerland(HolidayBase):
         # public holidays
         self[date(year, JAN, 1)] = 'Neujahrestag'
 
-        if self.prov in ('AG', 'BE', 'FR', 'GE', 'GL', 'GR', 'JU', 'LU',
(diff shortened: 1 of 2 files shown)


cgrigis:
A couple questions that came to mind:
* The handling of holidays that always fall on a Sunday seems inconsistent: Easter and Pentecost are marked as holidays in the Switzerland file, but the Federal fast is not. Also, Easter is e.g. not listed in the France file. What is the general policy?
* The language used for the holiday names seems to be the language of the country. But what about countries with multiple languages (such as Switzerland ;) )?

Thank you :-)

dr-prodigy:
Thank you very much for your work! 👍 
Cheers
dr-p

PR #306, opened 2020-04-24
Update korea.py

seriousran:
It must be processed as a hotfix to solve #305 

And it is from following problems.
Issue [#1437](https://github.com/facebook/prophet/issues/1438) in prophet by @chongdae and this PR [#1438](https://github.com/facebook/prophet/pull/1438) in prophet

PR #317, opened 2020-05-17
Fix AttributeError: 'Korea' and 'Vietnam' object has no attribute 'korean_cal'

pelennor:
It is a supplement for PR #306  
Not only korea.py, but vietnam.py uses 'korean_lunar_calendar'. So this issue must deal with korea.py and vietnam.py together.


dr-prodigy:
Thank you! 👍

PR #320, opened 2020-05-25
Add pop_named(), to make removing holidays easier.  No need to look up or calculate the date.

samtregar:
I found it surprisingly hard to remove "Columbus Day" from the holiday list - the example code in the README works but it requires that the caller know when the holiday falls, but that's why I'm using the library! :)  So this PR adds a pop_named() call that does the reverse lookup and pop.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -226,6 +226,9 @@ get_list(key)
 pop(key, default=None)
     Same as `get` except the key is removed from the holiday object
 
+pop_name(name)
+    Same as `pop` but takes the name of the holiday rather than the date
+
 update/append
     Accepts dictionary of {date: name} pairs, a list of dates, or even singular
     date/string/timestamp objects and adds them to the list of holidays
@@ -364,13 +367,12 @@ More Examples
     # "Ninja Turtle Day" on July 13th. We can create a new class that inherits
     # the UnitedStates class and the only method we need to override is _populate()
 
-    >>> from dateutil.relativedelta import relativedelta
     >>> class CorporateHolidays(holidays.UnitedStates):
     >>>     def _populate(self, year):
     >>>         # Populate the holiday list with the default US holidays
     >>>         holidays.UnitedStates._populate(self, year)
     >>>         # Remove Columbus Day
-    >>>         self.pop(date(year, 10, 1) + relativedelta(weekday=MO(+2)), None)
+    >>>         self.pop_name("Columbus Day")
     >>>         # Add Ninja Turtle Day
     >>>         self[date(year, 7, 13)] = "Ninja Turtle Day"
     >>> date(2014, 10, 14) in Holidays(country="US")
--- a/README.rst
+++ b/README.rst
@@ -226,7 +226,7 @@ get_list(key)
 pop(key, default=None)
     Same as `get` except the key is removed from the holiday object
 
-pop_name(name)
+pop_named(name)
     Same as `pop` but takes the name of the holiday rather than the date
 
 update/append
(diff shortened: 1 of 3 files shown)


samtregar:
Let me know if you'd prefer that pop() were extended to remove items by holiday name - considered doing that but thought that a distinct method might be clearer.

dr-prodigy:
Hi,
thank you very much in your interest and support.

Question: what is the goal to pop a specific holiday out of the list?
Why do you consider this to be useful in the library itself?

Thanks in advance, cheers
dr-p

samtregar:
My company doesn't observe that (racist) holiday.  I'm using the module to power a bot that needs to know when the holidays are, so it won't bug us with support tickets while we're on a break.  I imagine many people need to make adjustments to the official list of holidays for their country, many businesses has a slightly different list of which ones they observe.

samtregar:
Another way to look at it - it makes the example code in your README much easier to follow!  The old way of removing Columbus Day is super complicated and requires an external library.

samtregar:
@dr-prodigy What do you think - will you take the patch?  I was thinking I might break it out into a wrapper library if not.

dr-prodigy:
Hi,
thank you for your explanation and sorry for my late reply.
Apart from the very specific case of yours (the library aims to return all official holidays, so it's a little non-sense to provide a direct method to modify them programmatically just after), personally I don't like too much the idea of relying on the holiday's description (clearly prone to errors and breaks, due to slightly different text formats, accents etc.) to perform the removal.
Anyway, I'll do some more checking/reviewing and let you know.
I'll keep you posted, thanks again!
dr-p

dr-prodigy:
I'm sorry, even with your explanation, I still don't see a point in providing specific methods to remove/modify official holidays from the lists the library returns: if we talk about issues (a holiday is wrongly returned), the code is just to be updated/fixed. Any other case is a personal exception, therefore is to be managed by custom code. Or even with a forked version of the library, at worst.
Ultimately, being completely agnostic to your specific goals (moreover, I don't want to express a political opinion through my work, especially for sth I don't fight for: I know in the US Columbus Day is somehow controversial nowadays, but that's not "work" for my code) I can't really accept that specific holiday removal as an example in python-holidays documentation.
Closing this PR, thank you for your understanding.

samtregar:
> I'm sorry, even with your explanation, I still don't see a point in providing specific methods to remove/modify official holidays from the lists the library returns: if we talk about issues (a holiday is wrongly returned), the code is just to be updated/fixed. Any other case is a personal exception, therefore is to be managed by custom code. Or even with a forked version of the library, at worst.
> Ultimately, being completely agnostic to your specific goals (moreover, I don't want to express a political opinion through my work, especially for sth I don't fight for: I know in the US Columbus Day is somehow controversial nowadays, but that's not "work" for my code) I can't really accept that specific holiday removal as an example in python-holidays documentation.
> Closing this PR, thank you for your understanding.

You don't see a reason to do something that you demonstrate how to do (in a complicated way) in your README?  That's surprising, but of course it's your call.  Thanks for taking a look.

samtregar:
FYI: https://pypi.org/project/holidays-plus/0.0.1/

dr-prodigy:
Hi @samtregar sorry for my late reply.
On a second thought, I realized that the specific example you've been talking about was already there in the official README (it's been written before I started my activity on the project: y'know, the project is pretty large now, so some small things falls out of my control sometimes): my bad!
After saying this, I've been completely wrong in my previous evaluations and therefore, if you're still okay with it, I'll be adding it to holiday's main stream.

dr-prodigy:
Hi @samtregar ,
I've finally merged this in beta, applying some updates retrieved from your wrapper's lib code (ie: using get_named(..) to locate holidays to be popped. This, along with my version of get_named(), allows popping out a number of holidays through partial string match too).
Best regards,
dr-p

PR #322, opened 2020-06-03
Fix incorrect abbreviations for Arkansas

patrick-nicholson:
Two holidays that should apply to AR (Arkansas) incorrectly refer to the AS (American Samoa).


dr-prodigy:
Hi,
thank you very much for your contribution. Added some PEP8 and tests fixes + CHANGES addition, then merged to beta.
Cheers!
dr-p

PR #327, opened 2020-06-21
Update germany.py -  Added PROVINCES BYP

MikeTsenatek:
I added BYP for Bavaria with more protestants than catholic.

This package is used by home assistance for house automation. With this change it is possible to get woken up on Maria Himmelfahrt, so I can go to work


dr-prodigy:
Okay, merged in beta.. now you'll be safe on Maria Himmelfahrt too :-)
Cheers
dr-p

PR #346, opened 2020-08-08
Add UAE holidays

marcomasulli:
Hi @dr-prodigy , took a shoot at adding holidays for UAE after reviewing official calendars for the last 5-6 years. Apart from the usual issues with the islamic calendar, the tests proved accurate.

The UAE government also seems to extend public holidays duration arbitrarily by 1 or 2 days depending on the year, this is not something I implemented, would be a best guess situation.

Feel free to let me know if anything with this needs fixing, I hope this can be of help :-)

Cheers

### DIFF
--- a/holidays/countries/__init__.py
+++ b/holidays/countries/__init__.py
@@ -69,6 +69,7 @@ from .sweden import Sweden, SE, SWE
 from .switzerland import Switzerland, CH, CHE
 from .turkey import Turkey, TR, TUR
 from .ukraine import Ukraine, UA, UKR
+from .united_arab_emirates import UnitedArabEmirates, AE, ARE
 from .united_kingdom import UnitedKingdom, UK, GB, England, Wales, Scotland,\
     IsleOfMan, NorthernIreland, GBR
 from .united_states import UnitedStates, US, USA
--- a/holidays/countries/united_arab_emirates.py
+++ b/holidays/countries/united_arab_emirates.py
@@ -21,18 +21,22 @@ from holidays.utils import get_gre_date
 
 WEEKEND = (FRI, SAT)
 
+
 class UnitedArabEmirates(HolidayBase):
 
-    # Holidays are regulated by the Article 74 of Federal Law No. 08 for the year 1980: 
-    # https://www.ilo.org/dyn/natlex/docs/ELECTRONIC/11956/69376/F417089305/ARE11956.pdf 
-    # However the law is not applied literally, and was amended often in the past few years. 
-    # 2017 holidays: https://www.khaleejtimes.com/nation/uae-official-public-holidays-list-2017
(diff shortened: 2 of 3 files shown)


mborsetti:
Nice work.

A suggestion: you may consider shadowing https://github.com/dr-prodigy/python-holidays/blob/master/holidays/countries/singapore.py and make it explicitly clear when the holiday returned is estimated vs. an official date.  I don't think `holiday` should return estimates without being explicit about it.

If you have the time you can also load the actual dates for the past decade or so like for the Singapore module.

marcomasulli:
Good point, I will edit and resubmit. I had shadowed Egypt for the holiday pattern similarities, but did not have estimates there. About Singapore - do you have a source I can get the official dates from?

mborsetti:
The actual dates for Singapore I got directly from the website of the
Ministry of Manpower, which is the official source. To memory this is
documented in the code.

I assume the UAE has a similar official source of information.

marcomasulli:
Hi, I added sources and the actual dates since 2017. The only governmental source which I found only has the last holiday dates for 2019 and all 2020; for the other years I had to rely on national newspaper websites, cross checking each holiday (as they are all officially announced/confirmed separately). 

I will be able to add more information about the previous years further on, please let me know if anything needs fixing. 

Cheers!

mborsetti:
Code looks great to me, excellent!

One last bit: you also need to add testing code in [tests.py](https://github.com/dr-prodigy/python-holidays/blob/master/tests.py). Of course this is as easy as just adapting some other country's test code to the UAE holidays. Please ensure that the tests ideally have 100% [coverage](https://github.com/dr-prodigy/python-holidays#coverage) of the code you wrote, and you can run them yourself before pushing as per [here](https://github.com/dr-prodigy/python-holidays#running-tests). 

Tests not only ensure that the new code works, but also that future changes (including changes made by libraries that are being imported) don't cause it to stop working.

P.S. for the record I did not independently validate the dates.

marcomasulli:
Hi, thanks a lot for the comments and guidance - ran flake8 and tests, rewrote a few things to fit lines within 80 chars (apart from source links). Should something still need fixing please let me know. Cheers!

mborsetti:
Thank you!

I have no further comments, it's up to @dr-prodigy

dr-prodigy:
Hey there,
great collaboration work, brilliant! .. thanks a lot to you both! 👍 

Just for the sake of completeness, last things I made:
- moved target branch to "beta"
- made some more code updates to avoid flake8 complaining for lines too long
- had to revert `f'{aaa}'` strings (very elegant and effective, btw) to good old `'{}'.format(aaa)` to preserve Python2 compat

Merged and available in beta.. cheers! :-)

PR #354, opened 2020-08-30
fixed MLK comma issue

raffg:
This PR fixes https://github.com/dr-prodigy/python-holidays/issues/351 by removing the comma in the Idaho-specific MLK, Jr. holiday. I searched all countries and it appears that this is the only remaining comma included in a holiday name. This change removes any confusion with holidays falling on the same day being comma-separated.


dr-prodigy:
Hi, I've found 2 issues:
- target branch has always to be "beta"
- automatic tests were not updated ( => test fail)
Issues fixed, change merged, thx
dr-p

PR #358, opened 2020-09-11
Added day after Thanksgiving for California

bgmiles:
The day after Thanksgiving is a state holiday as noted here https://www.calhr.ca.gov/employees/pages/state-holidays.aspx


dr-prodigy:
Great job.. Thank you very much! 👍

PR #367, opened 2020-10-06
Day after Christmas is St Stephen's Day in Ireland

chiuczek:
A small fix, just encountered your library and think highly of it and spotted something I could easily do.
The day after Christmas is not called "Boxing Day" in Ireland, it is called [St Stephen's Day](https://en.wikipedia.org/wiki/Saint_Stephen%27s_Day).


dr-prodigy:
Thank you! 👍

PR #368, opened 2020-10-07
Initial rough set of holidays for Kazakhstan

chiuczek:
This is an initial set of holidays for Kazakhstan, which I could make use of in a side-project for work.

There is no accounting for holidays that fall on weekends, which should be observed on the next available work day(s). Some of the public holidays cover two or three days, which makes the logic a little trickier than just a single day. I'll come back to that later but thought this could be useful pretty immediately as it is

### DIFF
--- a/holidays/countries/__init__.py
+++ b/holidays/countries/__init__.py
@@ -41,6 +41,7 @@ from .ireland import Ireland, IE, IRL
 from .italy import Italy, IT, ITA
 from .israel import Israel, IL, ISR
 from .japan import Japan, JP, JPN
+from .kazakhstan import Kazakhstan, KZ, KAZ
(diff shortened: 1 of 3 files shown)


chiuczek:
I'm not sure why the coverage has reduced, I only _added_ code and tests. It's strange that tests are now missing a file I didn't touch...

dr-prodigy:
Hey there @chiuczek 
thank you very much for your support.
You can test your code and check coverage with commands described here:
https://github.com/dr-prodigy/python-holidays#coverage

I let you complete the work, then I'll merge it.. thanks again, cheers!
dr-p

dr-prodigy:
Closed for obsolence

PR #372, opened 2020-10-08
ISO code implementation for France

### LINKED ISSUE #348: Revisit #211 to include FR for France
This is an appeal to reconsider the request in #211 

Started using this library and I really like it, but this one exception to the [iso3166 standard](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes) is most unexpected. I'll skip a lengthy opinion on the value of standards and go directly to my question: if there was a pull request to adhere to iso3166, would it be considered? Or would that be wasted work because of reasons (backwards compatibility or anything else)?

Thanks,
rfm.

mborsetti:
Closes #348 

As long as dateutil.relativedelta.FR is not imported within the FR class, which isn't, we're fine.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -121,7 +121,7 @@ England                       None
 Estonia             EE/EST    None
 EuropeanCentralBank ECB/TAR   Trans-European Automated Real-time Gross Settlement (TARGET2)
 Finland             FI/FIN    None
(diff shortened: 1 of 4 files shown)


dr-prodigy:
Okay, you persuaded me :-) 👍

PR #375, opened 2020-10-08
Pop named() exception

### LINKED ISSUE #349: pop_named() sometimes generates an exception
This simple code block generates an exception when I try to pop "Columbus Day" (last line):

```
from datetime import date
  
import holidays

us_holidays = holidays.CountryHoliday('US', prov=None, state='CA', years=2022)

print(us_holidays.get('2022-10-10').format("%s"))

us_holidays.pop_named("Columbus Day")
```

The exception is `dictionary changed size during iteration`, and it seems to be related to popping elements out of a dictionary while you're iterating through it: https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error.  The element to be popped isn't actually removed.

I'm far from a competent Python programmer, but iterating over `list(self)` seems to solve the problem for me - no exception, and the item is removed properly:

```
    def get_named(self, name):
        # find all dates matching provided name (accepting partial
        # strings too, case insensitive), returning them in a list
        return [key for key in list(self) if name.lower() in self[key].lower()]
```

mborsetti:
Fix #349

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -26,7 +26,7 @@ Released July 15, 2020
 - Switzerland fixes (cgrigis)
 - Germany fix (MikeTsenatek)
 - Added korean_cal attribute to Korea and Vietnam (seriousran, pelennor)
-- United States fixes (patrick-nicholson, raffg, dr-p)
+- United States fixes (patrick-nicholson, dr-p)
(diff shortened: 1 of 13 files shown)


mborsetti:
Wrong branch

PR #380, opened 2020-10-09
Fallback on convertdate if hijri-converter package is not installed

mborsetti:
If `hijri-converter` package is not installed, fall back to using less precise `converdate` now that it's part of the installation (used for hebrew calendar) instead of skipping estimating of Hara Rasa Puasa and Hara Rasa Haji Singapore holidays.

Coverage will decrease since not all lines can be tested.


dr-prodigy:
Hey great one.. Thank you!
dr-p

PR #389, opened 2020-11-11
Feature/add malawi holidays

pietervdw115:
Missing Eid al-Fitr public holiday as the date of Eid depends on the sighting of the moon, there may be variations in the exact date that is celebrated around the world. The announcement of the exact dates of Eid Al-Fitr may not happen until close to the start of Ramadan.


dr-prodigy:
Hi @pietervdw115
very good work, thank you very much!

Merging in beta, cheers!

Issue #390, opened 2020-11-11
Holiday on Observed Holiday


pietervdw115:
In South Africa, when a holiday falls on an observed holiday the observed holiday should be moved on to the next available day

PR #392, opened 2020-11-16
Fix/observed holiday on holiday za

pietervdw115:
This is a fix for Issue #390


dr-prodigy:
Everything fine now.. thanks! 👍

PR #398, opened 2020-12-03
allow 120 characters

mborsetti:
Useful for PRs to come (sorry, I can't work well with 80 characters)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,5 @@
 [flake8]
+max-line-length = 120
 ignore = F401,W504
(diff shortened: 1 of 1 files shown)


dr-prodigy:
As written on your other PR, I'm definitely not a fan of that limit, but it's still considered a standard / best practice in Python coding (pls refer to https://www.python.org/dev/peps/pep-0008/#maximum-line-length ) so we'd better keep it (although not enthusiastically :-) ).

mborsetti:
Sorry, forgot!

PR #400, opened 2020-12-03
removed dependency on convertdate (unreliable project)

mborsetti:
Removed dependency on convertdate, which has been unstable, by copying and adapting the the needed code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ install:
       $TRAVIS_PYTHON_VERSION == 'pypy3'
     ]]; then pip install hijri-converter; fi
   - pip install python-dateutil
-  - pip install convertdate
   - pip install korean_lunar_calendar
   - pip install six
   - pip install flake8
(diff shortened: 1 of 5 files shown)


dr-prodigy:
Hi Mike @mborsetti 
we can't unfortunately merge this, because convertdate, out of hebrew calendar, is also required for our function utils.get_gre_date: removing it would cause code inconsistency.
About the 79 -> 120 chars, I'm definitely not a fan of that limit, but it's still considered a standard / best practice in Python coding (pls refer to https://www.python.org/dev/peps/pep-0008/#maximum-line-length ) so we'd better keep it (although not enthusiastically :-) ).

dr-prodigy:
PS: in fact, we've already overcome the tests failure due to convertdate, adding a keep-back constraint in holidays dependencies, so we are ready to release.
We should then work for a fix with the convertdate team ( https://github.com/fitnr/convertdate/issues/38 ) who seems to support us, as soon as we provide a test snippet (I should work on it in the next days.. or would it be easy for @mborsetti perhaps?)
Thx!

mborsetti:
Sounds good ... but sorry I can't do it.

PR #403, opened 2020-12-03
Removed Python 2.7 and updated Travis CI

mborsetti:
Did this one as a separate PR in case it's controversial, but 2.7 has officially been abandoned for over 11 months!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,23 @@
 # https://travis-ci.org/dr-prodigy/python-holidays
-dist: bionic
+dist: focal
 language: python
 cache: pip
 
 before_install:
-  - pip install --upgrade pip
+  - sudo apt-get -y update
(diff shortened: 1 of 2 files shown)


dr-prodigy:
Hi Mike,
while I agree on the drop for Python 3.5 (now completely superseded, so not to be "encouraged"), I think that, although Python 2.7 is definitely sunset, we should still keep compatibility with it because of these reasons:
- it's not in fact limiting our development
- we instead invested a lot of time in the past to provide the best support / consistency on all Python versions (eg: convertdate replacing hijri-converter, made by you). This work is very well designed and we should definitely not waste it
- removing 2.7 would require removing all the aforementioned logic to avoid reduction on coverage % 
- it's still used a lot (I don't think that's a good thing, but still we have to consider it!)

I hope I convinced you, we will come back to this some day anyway!

PR #417, opened 2021-01-27
Use Alpha-2 code as 'country'

javicalle:
To follow the homogeneous use in the rest of the Countries, I propose to modify the value of 'country' by the Alpha-2 code of the ISO-3166 standard.

https://www.iso.org/glossary-for-iso-3166.html

### DIFF
--- a/holidays/countries/india.py
+++ b/holidays/countries/india.py
@@ -27,7 +27,7 @@ class India(HolidayBase):
                  'TN', 'AP', 'WB', 'KL', 'HR', 'MH', 'MP', 'UP', 'UK', 'TS']
 
     def __init__(self, **kwargs):
-        self.country = "IND"
+        self.country = "IN"
         HolidayBase.__init__(self, **kwargs)
 
     def _populate(self, year):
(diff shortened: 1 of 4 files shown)


javicalle:
It won't be straight forward...
I hadn't seen the handful of validations within UnitedKingdom. IMHO this requires a review, but for now I am going to modify it just for Ireland.

javicalle:
I have splitted Ireland holidays.
'Common holidays' are populated in `UnitedKindom._populate()` method. Specific holidays are populated in method `_country_especific()`.
Ireland class overwrite this method with irish holidays or specific implementations for 'common' holidays.

Not sure if this is the way we want to go.
I just wanted to be able to have uniform 2 character country codes.
It's probably easier to define the code for Ireland and just replace all the conditionals with the value of the code:
```
if self.country != 'IE':
...
if year == 2020 and self.country != 'IE':

```

dr-prodigy:
Although - I confirm - it could be more accurate to completely split Ireland from UK, your proposed changes are very reasonable and improving: merging in beta, we will consider some further review in the future.
Thank you very much 👍

javicalle:
Thank you very much for your words.

I might also suggest treating "Isle of Man" as a country (according to ISO-3166-1) and it is possible that Wales, England, Scotland and Northern Ireland could be treated as United Kingdom states/provinces.

Thanks to you for your work.

PR #419, opened 2021-01-29
Fix for multi-year query

### LINKED ISSUE #418: Collecting holidays for Singapore takes to long!
Hi folks!

I noticed that, in newest version of you package (10.4), simple code which I'm already using takes a lot of time and at the end it raise error. I noticed that it happens only for 'Singapore'. In version 10.3 and earlier (10.2 for sure) it works normally. Please take a look into that. The following code is used daily in the popular fbprophet python package.
Code below

Best regards,
Mateusz Serocki

Checked on Windows and Linux
Python version 3.7
holiday version 10.4
hijri-converter version 2.1.1 # after uninstalled that package the problem still occurs with additional warning messages to install that one
```

import holidays #(version 10.4)
import numpy as np

country='Singapore'
years = np.arange(1995,2045)
holiday_names = getattr(holidays, country)(years=years).values() # <- it raise error
```
I attached error logs
[holidays error.txt](https://github.com/dr-prodigy/python-holidays/files/5885868/holidays.error.txt)

mborsetti:
Fixes #418

### DIFF
--- a/holidays/countries/singapore.py
+++ b/holidays/countries/singapore.py
@@ -186,7 +186,7 @@ class Singapore(HolidayBase):
         # a Sunday, the day next following not being itself a public holiday
         # is declared a public holiday in Singapore."
         for (hol_date, hol_name) in list(self.items()):
-            if hol_date.weekday() == SUN:
+            if hol_date.year == year and hol_date.weekday() == SUN:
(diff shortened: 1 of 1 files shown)


dr-prodigy:
Merged in beta.. finally!
Thank you Mike, cheers 👍

PR #420, opened 2021-02-01
update convertdate

jaemk:
Convertdate had a dependency cap (now removed) on pytz,
preventing downstream users from updating.
Convertdate 2.3.0 also introduced a breaking change to the
hebrew.to_jd_gregorianyear conversion - both return types
are now handled to support 2.2 and 2.3 being used downstream.

### DIFF
--- a/holidays/countries/israel.py
+++ b/holidays/countries/israel.py
@@ -12,7 +12,7 @@
 #  License: MIT (see LICENSE file)
 
 
-from convertdate import hebrew
+from convertdate import hebrew, gregorian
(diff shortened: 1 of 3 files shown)


jaemk:
Looks like this was fixed by 904a24f82fe01d9198b945b86f6426be3f03f13f

dr-prodigy:
In fact, it required a bit more work to get everything sorted :) but yes, now it's in fact fixed by v0.10.5.2 released some hours ago.
Thank you very much for your support!

PR #429, opened 2021-02-09
Add Saudi Arabia

OsaydAbdu:
For holidays in Saudi Arabia reference:
https://laboreducation.hrsd.gov.sa/en/gallery/274


dr-prodigy:
Brilliant.. Thank you very much! 👍

PR #438, opened 2021-02-21
Convert tests to pytest

s-weigand:
The change to pytest from #404  but as a more concise PR.

I will implement the other changes from #404 in separate smaller PR's, so they will be easier to review.

PR #404, opened 2020-12-07
Use gh action ci cd

s-weigand:
Here is the [long ago promised](https://github.com/dr-prodigy/python-holidays/issues/338#issuecomment-662592391) PR to change the CI-CD from travis to github actions. 

As mentioned in #341 I also implemented some other tooling changes:
- Added pre-commit
- Added autoformatting with black via pre-commit
- Changed flake8 linting to run via pre-commit
- Added tools to check `*.rst` files
- Changed test runner to be pytest
- Changed structure of tests, now the tests structure mirrors the structure of the project (IMHO makes it much easier to find the tests you are looking for)
- Changed contributing documentation to reflect changed tooling

Looking at some changes black made (e.g. for holidays/countries/hongkong.py), maybe a different formatter which is more configurable (e.g. [yapf](https://github.com/google/yapf)) would be better suited for this project.

For the CD to work you need to get an API token from pypi and set it a secret called `pypi_password` ([see](https://github.com/pypa/gh-action-pypi-publish#usage))

After that, you can simply run 
```console
$ bump2version major|minor|patch
$ git push --follow-tags
```
And the release is on its way ^^

PR #439, opened 2021-02-21
Use precommit

s-weigand:
This PR adds the pre-commit from #404 but without the auto-formatting.

It is based on #438, so no needed to review it before #438 is merged and this branch is rebased.
I will keep it in draft mode until it is ready for review.


dr-prodigy:
Please see #404 !

PR #446, opened 2021-03-09
fixed easter&labour day for serbia 2021

kosugor:
When one of the holiday days falls on Sunday, it is observed on next working day. Second day of Labour Day (May 2nd) is Sunday this year, which is also Orthodox Easter, so this year first working day is 4th of May.


dr-prodigy:
hi @kosugor thank you very much!
Cheers 👍

Issue #351, opened 2020-08-25
MLK still causes confusion at the state level


raffg:
If multiple holidays occur on the same day, they are separated with a `,`. This issue was fixed for `Martin Luther King, Jr. Day` by removing the comma (https://github.com/dr-prodigy/python-holidays/issues/295), however for Idaho the issue still exists:
https://github.com/dr-prodigy/python-holidays/blob/02d5d510202a02471cd0a939a11525cf02ca115c/holidays/countries/united_states.py#L107
I suggest removing the comma here as well. Or, in case this issue arises in other countries, maybe change the delimiter to a semicolon?

_Originally posted by @raffg in https://github.com/dr-prodigy/python-holidays/issues/295#issuecomment-680171130_

dr-prodigy:
Fixed and released in beta, thanks
dr-p

Issue #295, opened 2020-03-16
MLK day causes confusion


snoopyjc:
If multiple holidays occur on the same day, they are separated with ", ".  However, Martin Luther King day also contains a ", ", so my program doesn't know if that's a sequence of holidays or what.  I recommend removing the "," from the name of MLK day - in fact I subclassed Holidays and did just that!

dr-prodigy:
Good point.. fixed and pushed on beta branch.
Cheers!
dr-p

raffg:
This issue still exists with the Idaho holiday:
https://github.com/dr-prodigy/python-holidays/blob/02d5d510202a02471cd0a939a11525cf02ca115c/holidays/countries/united_states.py#L107
I suggest removing the comma here as well. Or, in case this issue arises in other countries, maybe change the delimiter to a semicolon?

PR #448, opened 2021-03-22
Remove commas from UK holidays

bletham:
The `get_list` method assumes that commas are not used in the names of individual holidays, by splitting overlapping holidays according to commas:
https://github.com/dr-prodigy/python-holidays/blob/cb875fb87e91f8c555966659bca1636519c47d6f/holidays/holiday_base.py#L144-L145

Violations of this assumption have come up before in #295 and #351 and been fixed by removing the comma.

There are some commas in UK holiday names causing the same problem:
```Python
import datetime
import holidays

uk_holidays = holidays.UK(years=[2021])
x = uk_holidays.get_list(datetime.date(2021, 8, 30))
print(x)
# ['Late Summer Bank Holiday [England', 'Wales', 'Northern Ireland]']
```
where it is split into a 3-element list instead of being a single element.

This fixes it by removing the commas. After the fix, the code above prints `['Late Summer Bank Holiday [England/Wales/Northern Ireland]']` as expected.

### DIFF
--- a/holidays/countries/united_kingdom.py
+++ b/holidays/countries/united_kingdom.py
@@ -114,7 +114,7 @@ class UnitedKingdom(HolidayBase):
         if self.country != "Scotland":
             name = "Easter Monday"
             if self.country == "UK":
-                name += " [England, Wales, Northern Ireland]"
+                name += " [England/Wales/Northern Ireland]"
             self[easter(year) + rd(weekday=MO)] = name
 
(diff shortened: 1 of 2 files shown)


dr-prodigy:
Hey @bletham thank you for your work.. just merged in beta!
Cheers
dr-p

bletham:
Awesome, thanks!

PR #452, opened 2021-03-27
fix quality assurance failures

giladmaya:

PR #453, opened 2021-03-31
Use setup.cfg only pattern instead of bump2version

s-weigand:
[As mentioned](https://github.com/dr-prodigy/python-holidays/pull/404#issuecomment-811237868), using a `setup.cfg` only pattern has the benefit (as does `bump2version`) that you only have to manage the version string in one place but in contrast to `bump2version` you don't need to take care of all the version parsing, replacing and especially that bug with the trailing whitespace which together with `pre-commit` nullifies all the benefits of `bump2version`.

### How a typical release workflow would look like
1. Change the version  string of `__version__` in `holidays/.__init__.py` to the desired value (push or PR)
2. Create a new release in the [release tab](https://github.com/dr-prodigy/python-holidays/releases), this will also create a tag
3. Wait for the CI to publish it to PyPi 🎉 

Here is also [proof that the release works](https://test.pypi.org/project/holidays/) as desired, with a quick release in the [PyPi test server](https://packaging.python.org/guides/using-testpypi/)

closes #404 closes #452

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,8 @@ repos:
     hooks:
       - id: rstcheck
         additional_dependencies: [rstcheck]
+
+  - repo: https://github.com/asottile/setup-cfg-fmt
+    rev: v1.16.0
+    hooks:
+      - id: setup-cfg-fmt
--- a/holidays/countries/jamaica.py
+++ b/holidays/countries/jamaica.py
@@ -24,7 +24,7 @@ class Jamaica(HolidayBase):
     # https://en.wikipedia.org/wiki/Public_holidays_in_Jamaica
 
     def __init__(self, **kwargs):
-        self.country = 'JM'
+        self.country = "JM"
         HolidayBase.__init__(self, **kwargs)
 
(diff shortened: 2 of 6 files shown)


dr-prodigy:
Thank you really much for your support @s-weigand !
I've just released this last request in master, and published a new release as suggested.. 
On a second thought, in fact, I decided to name it 0.11.1, due to the major change in the whole project structure :-) but your work will surely come handy in the next releases (given that I will probably maintain even the sub-sub versions numbers for some cases).
Currently waiting for the first publishing on pypi. In case that won't happen (I set up the required "PYPI_PASSWORD" secret in the repo too, but this still has to run once), I will use the manual procedure.
Thx again, cheers! 👍

s-weigand:
Looks like all [wents as planned](https://github.com/dr-prodigy/python-holidays/runs/2253979126?check_suite_focus=true) :tada:

dr-prodigy:
***k yeah!! 😄🤘
Thx loads, cheers
dr-p

s-weigand:
Automate the boring stuff or at least as much as possile 😄

PR #456, opened 2021-04-07
Added Singapore official holidays for 2022

mborsetti:
Freshly released per https://www.mom.gov.sg/employment-practices/public-holidays


dr-prodigy:
Grazie come sempre Mike! 👍

PR #480, opened 2021-06-30
Strip leading whitespace Botswana holiday movement

pietervdw115:
If a public holiday falls on an observed holiday, that public holiday is observed the next day. 
To implement this get the public holidays after adding the observed holidays and then splitting strings where more than one holidays falls on a date.
Added .lstrip() to this in order to remove the leading space


pietervdw115:
Hi @dr-prodigy ,
I have fixed a little visual issue by adding an .lstrip() in the code. 
Not sure why the tests are failing as the coverage is still 100%

PR #486, opened 2021-07-13
Update/nigeria holidays

pietervdw115:
The list of holidays were incomplete and some were calculated incorrectly 

Also added Muslim holidays to the best of my knowledge. 

These can and will change, but we will only know this a month before the holiday and even then the public holiday may not be on the day of the actual celebration


dr-prodigy:
Thank you very much! 👍

PR #487, opened 2021-07-13
Correct 2022 Spring Bank Holiday

orrock:
The Spring Bank Holiday is normally the last Monday in May. In 2022 it is moving to 2nd June to allow for a four day weekend, as per https://www.gov.uk/bank-holidays


dr-prodigy:
Thx! 👍

PR #494, opened 2021-07-19
Removed dependency on six

mborsetti:
We no longer support Python 2 so importing `six` no longer has any utility.

I have replaced the only instance of the module usage with the Python 3 equivalent (see https://six.readthedocs.io/#six.string_types) and have removed `six` as a project dependency.  Fixing typos (space missing) is a side benefit ;)


dr-prodigy:
Ciao Mike!
Always on the trail.. ;) thank you! 👍

PR #497, opened 2021-07-19
Type hinting and docstrings

mborsetti:
It's time we added type hints to make this package a bit more developer-friendly, so here's my first contribution to type hinting and docstring inline documentation (which I developed out of frustration while reviewing the Malaysian submission over the weekend).

While adding type hints I have encountered a few things in the `HolidayBase` class I simply could not avoid having to refactor, and a few that needed cleaning up:
* In `__init__` method:
  * The attribute `PROVINCES` was unused and has been removed.
  * Th call to the `__init__` of the superclass was missing and has been added.
  * Generation of a set of years was cleaned up.
  * Switched year iteration over set().copy() from list(set()). I contributed the original code change, which avoided corner cases of `RuntimeError: Set changed size during iteration`, so I take full responsibility for the original suboptimal solution.
* The `__add__` method had to be completely refactored:
  * The class `HolidaySum` cannot be buried inside a function or typing and docstring won't work in IDEs, so I moved it to top level.
  * The determination of the joint list of `country` and `prov` logically belongs to the class initializer, not the function, so I moved it there.
  * Same for the logic combining the boolean `expand` and `observed`.
  * A joint list for `state` was never created so added the code to do so.
* Note that type hinting in `__keytransform__` is correct once the six removal PR is applied.
* Fixed the `get_named` function which although unused internally it should probably work as advertised (thank you type hinting for pointing out the problem)!

I have also added `doctest` testing so that the code used as examples in the docstrings is guaranteed to run.

Finally, in utils.py.get_gre_date() I removed the bit in the description talking about what (used to) happen when I originally wrote this function) if hijri-converter library is not installed -- it is now a project requirement.  Also updated the format of the docstring.

TODO: Inline (docstring) documentation can be improved much more (for example `expand` and `observed` aren't even mentioned, and I still have to understand what they do), but it's a start!

### DIFF
--- a/holidays/__init__.py
+++ b/holidays/__init__.py
@@ -26,7 +26,7 @@ from holidays.constants import (
     NOV,
     DEC,
 )
-from holidays.holiday_base import HolidayBase, createHolidaySum
+from holidays.holiday_base import HolidayBase
 from holidays.utils import list_supported_countries, CountryHoliday
 
 __version__ = "0.11.2"
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -12,31 +12,177 @@
 #  License: MIT (see LICENSE file)
 
 from datetime import timedelta, datetime, date
+from typing import Any, Iterable, List, Optional, Tuple, Union
 
 import six
 from dateutil.parser import parse
 
 
 class HolidayBase(dict):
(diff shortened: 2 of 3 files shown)


mborsetti:
Wrong branch

PR #481, opened 2021-07-04
Add Malaysian holidays

jusce17:
The following holidays still need to be added 

- Isra and Miraj in Malaysia
- NUZUL AL-QUARAN DAY
- The Sultan of Pahang Hol in Pahang
- Birthday of SPB Yang di Pertuan Agong
- Birthday of the Sultan of Kedah in Kedah
- Birthday of the Raja of Perlis
- Hijri New Year Awal Muharram or Maal Hijrah
- The Sultan of Johor Hol
- Muhammad's Birthday

The following holidays are hardcoded and must be changed 
- Thaipusam Holiday

PR #498, opened 2021-07-19
Type hinting and docstrings

mborsetti:
It's time we added type hints to make this package a bit more developer-friendly, so here's my first contribution to type hinting and docstring inline documentation (which I developed out of frustration while reviewing the Malaysian submission over the weekend).

While adding type hints I have encountered a few things in the `HolidayBase` class I simply could not avoid having to refactor, and a few that needed cleaning up:
* In `__init__` method:
  * The attribute `PROVINCES` was unused and has been removed.
  * The call to the `__init__` of the superclass was missing and has been added.
  * Code that turns years into a set has been cleaned up.
  * Switched to iterating years over set().copy() instead of list(set()). I contributed the original code change, which avoided corner cases of `RuntimeError: Set changed size during iteration`, so I take full responsibility for the original suboptimal solution.
* The `__add__` method had to be completely refactored:
  * The class `HolidaySum` cannot be buried inside a function, or typing and docstring won't work in IDEs, so I moved it to the top level.
  * The determination of the joint list of `country` and `prov` logically belongs to the class initializer, not the function that instantiates the class, so I moved it there. A joint list for `state` was never created so I also added it to the code.
  * Same for the code that combines the boolean `expand` and `observed` -- refactored.
* Note that type hinting in `__keytransform__` is correct once the six removal PR is applied (it's generating an error in my IDE).
* Fixed the `get_named` function, which although apparently unused internally, it should probably work as advertised (thank you type hinting for pointing out the problem)!

I have also added `doctest` testing to `holidays_base.py` to ensure that the code used as docstring examples is guaranteed to run and produce the same results.

Finally, in `utils.py.get_gre_date()` I removed the bit in the docstring talking about what used to happen (when I originally wrote this function) if hijri-converter library is not installed -- it is now a project requirement, so it's irrelevant. Also updated the format of the docstring.

TODO: Inline (docstring) documentation can be improved much more, but I feel that this is a decent start.

SUGGESTION: Move the user documentation to the [Sphinx](https://www.sphinx-doc.org/en/master/) tool and host it (for free) on [readthedocs](https://readthedocs.org/).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,7 @@ repos:
     rev: 3.8.4
     hooks:
       - id: flake8
+        args: [--max-line-length=79]
 
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.7.0
--- a/holidays/__init__.py
+++ b/holidays/__init__.py
@@ -26,7 +26,7 @@ from holidays.constants import (
     NOV,
     DEC,
 )
-from holidays.holiday_base import HolidayBase, createHolidaySum
+from holidays.holiday_base import HolidayBase
 from holidays.utils import list_supported_countries, CountryHoliday
 
 __version__ = "0.11.2"
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -419,13 +419,16 @@ class HolidaySum(HolidayBase):
 
         >>> from pprint import pprint
         >>> import holidays
-        >>> nafta_holidays = holidays.US(years=2020) + holidays.CA() + holidays.MX()
-        >>> pprint(sorted(nafta_holidays.items(), key=lambda x: x[0])[:10])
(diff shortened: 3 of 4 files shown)


mborsetti:
Ciao Maurizio,

Any progress on merging this?  I was going to fix the Malaysian holidays #481 as promised this weekend but found that this is still pending, Since it's an enhancement that reduces development errors, I don't want to proceed until it's (preferably) merged or (hopefully not) rejected. Grazie.

dr-prodigy:
Hi Mike @mborsetti !
Sorry for being late.. I did some merging days ago, but this your PR was too complex to run it "blindly" (or almost so!).
Hereiam anyway.. thank you very much for your work, again a great improvement!
Ciao, see you 👍

mborsetti:
Thanks/Grazie.  Hopefully will get Malaysia done next weekend.

PR #515, opened 2021-08-22
Adding China

tserekh:
Information about holidays is taken from here https://en.wikipedia.org/wiki/Public_holidays_in_China

The most complicated and confusing holiday is Tomb-Sweeping Day which can be on 4, 5 or 6 April. It happens  `on the15th day after the Spring Equinox`.  Spring Equinox is on March 20 but Tomb-Sweeping Day varies. Maybe Chinese define it in a different way.
Hardcode Tomb-Sweeping Day was not an option because different source conflicts and provide different dates
So i considered it happens on every April 5.


xuweiearth:
In my opinion, "Tomb-Sweeping Day" should be `self.get_solar_date(year, 2, 23)`. Moreover, you can get annual Chinese holiday and the day to balance off the break from a Python package `lunar_python` by calling function `getHolidays(year)`.

PR #522, opened 2021-09-19
Fix .gitattributes

mborsetti:
Fixed `.gitattributes` file (had incorrect syntax) so I can work on this in Windows without end-of-line conversion issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-*	text=lf
-*.bat	text eol=crlf
+*       text eol=lf
+*.bat   text eol=crlf


dr-prodigy:
Merged.. Thank you Mike! 👍

dr-prodigy:
Hey there @mborsetti 
I had to get back to this to review the line ending logic (especially regarding the *.bat rule, which was introducing mixed line ending issues on the newly-added make.bat file coming created by Sphinx).
In particular, I switched to auto mode: it should suit both Linux, Mac and Win workstations.

Would you please confirm everything is still working properly for you?

Thx!

mborsetti:
Ciao Maurizio,

I see your logic but unfortunately it doesn't work that way: specifically, under automatic mode all of my future PRs will have CRLF line endings (because they are developed on Windows), as will those submitted by others who are also on Windows. And I think those from Mac environments will be coming in with LF as a line ending.

This PR was going to standardize all submissions to having LF, except for Windows-only *.bat files who would follow the Windows convention of CRLF.

dr-prodigy:
Ciao Mike,
I see.. but theoretically, in my knowledge, auto mode should:
- uniform all the files to a specific "internal" format (it should be cr, although I'm not totally sure of this) while pushing on the github server
- apply the right conversion on single machines upon pulling, depending on the workstation type (ie: cr for linux, lf for mac, crlf for win)
Moreover, this behaviour should be enforced on every users' workstation by the presence of the .gitattribute file, which defeats any git local settings on the user's machine.
So it should fit any case (it is in fact the suggested value to use).

Instead, forcing crlf on server side too (previous status) for specific files only, breaks the pre-commit EOL mix tests, making the tests fail: build checks suddenly failed, as soon as make.bat file from Sphinx - the one and only .bat file currently in the proj - appeared after my first Sphinx commit.

Could you make a double-check on the new behaviour please?
And btw.. why would you need a .bat file specific setting, not being any other .bat file in the project apart from this last make.bat one?

..Thx!

mborsetti:
Oh, I see the issue with pre-commit the way it's done.  Yeah, I don't think think there's a practical need to differentiate *.bat; I'm pretty sure Windows will run them even with with LF without complaining (unlike Bash which chokes when it finds CRLFs)!

I will test the "auto" settings and let you know.

PR #523, opened 2021-09-19
Refactor lunisolar and islamic calendar

mborsetti:
Prep work for Malaysia holidays cleanup.

- Refactored Chinese lunisolar calendar calculation into the new `LuniSolar `class shared in utils.py. This lengthy code was duplicated in Hong Kong (where it originates from before I copied it), Singapore, and now Malaysia.
- Used the new `LuniSolar `class for China holidays (instead of Korean (!))
- Renamed the shared function `get_gre_date)` (the original name I gave it) to `islamic_to_gre`, which is less confusing as we are now dealing with multiple calendars in utils.py.
- Renamed `isLeapYear` function in HongKong to the Pythonic `is_leap_year`
- A few docstring updates here and there


mborsetti:
Sorry, this failed and I realize the Malaysian file I was working on creeped into this PR.  Apologies, I don't know what to do now but it's all rolled into #524

PR #525, opened 2021-09-21
Australia Vic Grand Final Day 2021 Special Case Added

jeremychrimes:
I have made a quick change to the Public Holiday generation algorithm for Australia which adds a special case for the 2021 grand final date being Friday 24/SEP/2021. Every year should be a special case for grand final day as it can change depending on the needs of the AFL and is not fixed. 
I have created a simple test case for this in the test_grand_final test function.


dr-prodigy:
Just merged in beta.. thank you very much! 👍
Cheers

PR #534, opened 2021-10-01
Add holidays for Kazakhstan

chiuczek:
I made an earlier PR on this but left it get stale, so here's another go at adding Kazakh holidays to the package. The tests run ok this time, and all pre-commit checks pass. I still haven't worked through each holiday to account for observance of days that fall on weekends, but getting the list of holidays in would already be helpful for me and mean I can move an internal tool from using my fork to running directly on upstream


dr-prodigy:
Hi @chiuczek, apologies for the belated answer.. just merged on beta!
Thank you very much, cheers 👍

PR #564, opened 2021-11-29
Update colombian holidays

jahirfiquitiva:
- Update discovery of América to Columbus Day
- Use accented í for día

I'm colombian and we don't call that day "Discovery of America" but "Columbus day" and "Día de la raza" in Spanish.

Thanks!


dr-prodigy:
Made some mess with branches :-S , so I had to re-apply your fix with a new commit.
Merged in beta, thanks for your help! 👍

PR #574, opened 2021-12-07
Add `NorthMacedonia` import to __init__

javicalle:
Fix missing import for `NorthMacedonia` class/country.

PR #575, opened 2021-12-08
New `get_supported_countries` function

javicalle:
This PR adds a new `get_supported_countries` function that has a different approach to the existing `list_supported_countries`.
This implementation search for the `HolidayBase` subclasses that are in the `holidays.countries` folder/module.
With this approach only first subclass is take in consideration.
The method returns a map the keys of which are the countries (large name) and as values a `CountryInfo` object created for this purpose.

This class exposes the following information about the countries supported by the library:
* `name`
* classes extending from class
* `states` (if you have)
* `provinces` (if you have)

Also a `subcountries` property has been added to return the list of `states` or `provinces` in a unified way.

### DIFF
--- a/holidays/utils.py
+++ b/holidays/utils.py
@@ -19,6 +19,7 @@ from typing import Iterable, List, Mapping, Optional, Type, Union
 import holidays
 from hijri_converter import convert
 
+
 def list_supported_countries() -> List[str]:
     """List all supported countries, including their ISO 3166-1 Alpha country
     codes (Alpha-2 and Alpha-3).
--- a/holidays/utils.py
+++ b/holidays/utils.py
@@ -19,9 +19,6 @@ from typing import Iterable, List, Mapping, Optional, Type, Union
 import holidays
 from hijri_converter import convert
 
-import dataclasses
-
-
 def list_supported_countries() -> List[str]:
     """List all supported countries, including their ISO 3166-1 Alpha country
     codes (Alpha-2 and Alpha-3).
--- a/holidays/utils.py
+++ b/holidays/utils.py
@@ -50,7 +50,7 @@ def get_supported_countries() -> Mapping[str, "CountryInfo"]:
     }
 
 
-@dataclasses.dataclass
+# @dataclasses.dataclass  # only for Python > 6. Can be added as a dependency
 class CountryInfo:
     """Container for country info."""
 
--- a/holidays/utils.py
+++ b/holidays/utils.py
@@ -46,7 +46,7 @@ def get_supported_countries() -> Mapping[str, "CountryInfo"]:
         for k in countries
         if k.__module__.startswith(
             "holidays.countries"
-        )  # only subclasses from countries
+        )  # only subclasses from countries module
     }
 
 
--- a/holidays/utils.py
+++ b/holidays/utils.py
@@ -12,12 +12,14 @@
 #  License: MIT (see LICENSE file)
 
 import inspect
+from datetime import date, timedelta
 from functools import lru_cache
-from typing import Iterable, List, Optional, Type, Union
+from typing import Iterable, List, Mapping, Optional, Type, Union
 
 import holidays
-from datetime import date, timedelta
 from hijri_converter import convert
+
+import dataclasses
 
 
 def list_supported_countries() -> List[str]:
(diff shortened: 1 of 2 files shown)


javicalle:
Upssss, not dataclasses in Python6.
Would it be appropriate to add it as a dependency?I will delete it for now.

javicalle:
The new approach have some differences with the actual `list_supported_countries`.
Fisrts of all, don't return all the Alpha-2 and Alpha-3 code classes, only the 'large' country Name.

But also there are some diferences in the list that returns (despite teh short codes).
I have run a little test to validate diferences in implementations:
```python
l_s_c = holidays.utils.list_supported_countries()
g_s_c = holidays.utils.get_supported_countries()
filtered_cntrs = {fc for fc in l_s_c if len(fc) > 3}  # only validate against 'large' names

for k in filtered_cntrs:
    if k not in g_s_c:
        print("-- %s" % k)

for l in g_s_c:
    if l not in filtered_cntrs:
        print("++ %s" % l)
```

and here is the result:
```
-- PRTE
-- Ireland
-- Wales
-- Scotland
-- NorthernIreland
-- PortugalExt
-- IsleOfMan
-- Slovak
-- Polish
-- England
```

So, there is no 'country' in the new method that does not return the first one (there was one, but PR #574 is being corrected).
But there are a few 'countries' in the old method that are not included in the new one. Let's see which ones are:
* deprecated classes:
* * Wales
* * Scotland
* * NorthernIreland
* * IsleOfMan
* * Slovak
* * Polish
* `UnitedKingdom` subclasses
* * Ireland
* not filtered
* * PRTE

IMHO the deprecated classes is fine that they do not come out (we don't want them to be usable right?)
`Ireland` is a country of its own, so the initialization could be modified to make it extend from HolidayBase and get all the `UnitedKingdom` holydays.
And `PortugalExt` is something odd. The same approach as in `Ireland` can be taken here and extend from` HolidayBase`. `PRTE` appears here because size > 3 (and not, can not put > 4 because `Peru`).

javicalle:
:warning: Tests will fail until PR #574 is merged.

javicalle:
Close/open to launch tests

dr-prodigy:
Hi @javicalle ,
thank you for your work.

Although in the meantime I/we have made some improving code fixes (eg: removed support for already deprecated names, some UK cleanups, etc..), I still don't like so much the constraints inducted by this change, and I also feel it lacks some of the required strictness (especially regarding some countries lost on the way).

I would possibly prefer a different, more controlled approach, such as adding a class property to HolidayBase, eg:
```
class HolidayBase(dict):
    synonym = False
    [...]
```
to be then overridden by synonym classes, ie: all ISO code-based classes should do
```
class PRTE(dict):
    synonym = True

```
so finally, when returning the list, only non-synonym classes would be considered.

Although requiring some extra work (small refactoring on every class), this would allow complete control on the listed classes, without enforcing rules on the class hierarchy.

Pls let me know your thoughts / if you would like to work on this change.

Thank you very much, cheers!

javicalle:
I understand (and share) the objections to my approach. In my defense I will say that I have looked for a middle ground between the current implementation of `list_supported_countries` and that it was not too disruptive. 😇 

The proposed approach does not seem bad, but I have some doubts, how are the `aliases` related to the `main` class?

I've been trying another approach, which maybe could fit:

```python
class Spain (HolidayBase, metaclass = AKA):
    "" "Spain holidays." ""

    country = "ES"  #or maybe alpha2 
    alpha3 = "ESP"
   .../...
```

The `AKA` metaclass will create the class aliases:
```python
class AKA (type):
    """
    'Also Known As' metaclass to create aliases for a class.
    This will create aliases for the Alpha-2 and Alpha-3 country codes.
    """

    def __new __ (cls, classname, bases, attrs):
        class_ = type (classname, bases, attrs)
        globals (). update ({attrs.get ("country"): class_})
        globals (). update ({attrs.get ("alpha3"): class_})
        return class_
```

This way any Holiday will be defined in a single class. The aliases are created automagically. All class info can be getted from the class.
Constraints must be defined (is `country` attribute required? and `alpha3`?), controls must be added and so on, but I believe that can be a clean approach.

What do you think?

Any other approach would be fine to me but my concerns are mainly 2:
* be able to obtain some type of list that allows exploiting the configurations
* some freedom to create class structures in each 'country' file

Cheers.

dr-prodigy:
Hi @javicalle
thank you and sorry for my belated answer.
My last code snippets were not in fact exact nor totally clear: I've reviewed them.

Your approach seems promising: it's very smart and advanced (although maybe a bit complex to be understood, to be honest! :-) ).
I need a bit more time to analyze it in depth.

Pls let me know if the last changes I applied to my previous message make myself a little clearer, and what do you think about my proposition.

Thank you very much!

dr-prodigy:
Hi, this has been superseded by #608 (different, more structured approach, similar results), therefore I'm closing it.
Thanks a lot, cheers! 👍

PR #577, opened 2021-12-11
Defining ethiopian_isleap function

tedtad:


tedtad:
@dr-prodigy , thanks a lot.

PR #603, opened 2022-01-17
[Feature] Spain: Adds 2022 holidays

gtrabanco:
## Description
Adds 2022 holidays in Spain and it provinces.

All dates over 2022 will be considered as 2022 for estimate holidays but every year around October the national holidays are published in official publication BOE and the festivities of each province in similar publication that is done by province authorities.

This adds all festivities in 2022.

## Test

Deleted the test for province days later than 2022 because it is an estimation and due the explanation in the description we can not test and know which will be the days of each province always, there are some that changes every year.

## Known issues
I am not able to know it and if it is happening I am not able to fix it but, I think this library is not moving Sunday festivities to Mondays as it should do so 2022 holidays that are on Mondays by this reason I have to move them manually.

Anyway it is not always true that Sunday holidays go to Monday because it is holiday anyway but not always in all provinces.


dr-prodigy:
Fixed some small issues and merged in beta.. thx a lot! 👍

Issue #569, opened 2021-12-02
US Holidays Error: dictionary changed during iteration when passing years to holidays.US constructor


extrospective:
Attempting to use code to get US holidays for specific years.  This code works for some countries but not US (nor Canada)

```
import holidays
import numpy as np
print(holidays.__version__)
years = np.array([2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021])
x = holidays.US(years=years)
```

This code works when I try different countries: FR, GB, RU, DE, ES, CZ, AU, JP, CN, IN

This syntax **fails** to return results for the countries: US, CA

This syntax for passing numpy arrays I found in a ticket with a CZ example.  (https://github.com/dr-prodigy/python-holidays/pull/528).

I am running on Databricks runtime 10.0.  The holidays version is 0.11.3.1 as you see below.

The output I got was for US (or CA) is:
```
0.11.3.1
Out[41]: 
RuntimeError: dictionary changed size during iteration
```

Example output when I use another country, such as AU is:
```
0.11.3.1
Out[42]: {datetime.date(2016, 1, 1): "New Year's Day",
 datetime.date(2016, 1, 26): 'Australia Day',
 datetime.date(2016, 3, 25): 'Good Friday',
 datetime.date(2016, 3, 28): 'Easter Monday',
 datetime.date(2016, 4, 25): 'Anzac Day',
 datetime.date(2016, 12, 25): 'Christmas Day',
 datetime.date(2016, 12, 27): 'Christmas Day (Observed)',
 datetime.date(2016, 12, 26): 'Boxing Day',
 datetime.date(2017, 1, 1): "New Year's Day",
 datetime.date(2017, 1, 2): "New Year's Day (Observed)",
...
```

dr-prodigy:
Hi @extrospective
thank you for your interest in python-holidays.

I've made a test on your code, but I can't seem to replicate your issue, neither with US, nor with CA.
Here's my result:

```
>>> import holidays
>>> import numpy as np
>>> print(holidays.__version__)
0.11.3.1
>>> years = np.array([2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021])
>>> x = holidays.US(years=years)
>>> x
[...]

PR #586, opened 2022-01-05
Do not expand into other years

mborsetti:
This fixes #569 by ensuring that only holidays for the year requested are returned, preventing any inadvertent expansion.  This is enforced through a new test (which identifies the offending holiday, if any, for easier debugging). 

As a result quite a few countries had to be fixed as they were not behaving correctly. Of note particular note is Israel as it required extensive modifications as the variable `year`(!) was being overwritten within the function itself(!), and Botswana, which returned a holiday in 2019 at all times.

PR #606, opened 2022-01-21
Do not expand into other years v2

mborsetti:
While working on an another PR I discovered that  my PR #586 (Do not expand into other years) was incomplete: I missed that holidays for Saudi Arabia for 2006 would also return holidays for 2007, violating the "do not expand into other years" rule and triggering the "dict expansion while iterating" error in people's code.

The issue, it turns out, is multi-day Islamic holidays; in this specific case it was the Eid al-Adha Holiday running from 30-Dec-2006 to 03-Jan-2007, across Gregorian years.

Here is a fix for this issue, covering Saudi Arabia and all other countries affected. It's not very elegant but I couldn't think of a better solution (open to ideas). The fix has been successfully tested on all years from 1950 to 2050 inclusive.

I also addressed:
* The Saudi Arabia tests had some assertions remarked out, and these were added back in.
* The Saudi Arabia tests were strengthened by adding assertions for 2022 holidays.
* The code for Saudi Arabia had the remark "Weekend used to be THU, FRI before June 28th, 2013", which was not implemented in code; I implemented it
* The `test_all_countries` test I wrote for PR #586 (Do not expand into other years) does not need to be run in all Python versions since it's a logic test, not a code compatibility one; I fixed this so we don't have needlessly long test times (they are already getting a bit too lengthy IMHO).

### DIFF
--- a/holidays/countries/burundi.py
+++ b/holidays/countries/burundi.py
@@ -39,6 +39,15 @@ class Burundi(HolidayBase):
         HolidayBase.__init__(self, **kwargs)
 
     def _populate(self, year):
+
+        def _add_holiday(dt: date, hol: str) -> None:
+            """Only add if in current year; prevents adding holidays across
+            years (handles multi-day Islamic holidays that straddle Gregorian
+            years).
+            """
+            if dt.year == year:
+                self[dt] = hol
+
         # New Year's Day
         self[date(year, JAN, 1)] = "New Year's Day"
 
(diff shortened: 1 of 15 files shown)


mborsetti:
@dr-prodigy  Maurizio -- #609 will have to be pulled in here.  Let me know if you need more info to merge this one.

dr-prodigy:
Merged now! Going to look #609 .. in case of doubts I'll get back to you.
Thank you as usual Mike! 👍

mborsetti:
BTW this PR also fixes #569 which should probably be closed at the next release.

Issue #597, opened 2022-01-12
If state doesn't exist, it should raise an error


asmaier:
This is confusing:

```
>>> import date
>>> import holidays
>>> date(2021,3,8) in holidays.CountryHoliday("DE", state="BE")    # should raise an error, because state="BE" does not exist
False
>>> date(2021,3,8) in holidays.CountryHoliday("DE", prov="BE")
True
```
The first line after the imports should better raise an error, because for Germany one has to use `prov` and not `state`. At the moment this leads to subtle bugs in computing wether a day is a holiday or not.

mborsetti:
@dr-prodigy I have a proposed solution.

According to the [README](https://github.com/dr-prodigy/python-holidays/blob/master/README.rst), we don't have a single country where we use **both** `state `**and** `prov`.  Since this is the case, I propose we simplify everything by conforming to [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2#Current_codes) and call these **subdivisions** (argument name `subdivision` or `subd`) and make everything simple.

We can maintain backwards compatibility by assigning the value of either `state` or `prov`, if specified, to `subdivision`.

If you agree I can do global changes this weekend and submit a PR.

dr-prodigy:
Hey @asmaier
thank you for pointing this out. This inconsistency in fact has already been raised and would in fact require a review.

@mborsetti
thank you, I like your solution a lot! If you wish to implement it, I'll absolutely bless it :-)
Then - I believe - we could/should possibly review the current subdivision names to adhere to the reported ISO codes, but that can come at a later time.

Thanks both! 👍

mborsetti:
@dr-prodigy 
> I like your solution a lot! If you wish to implement it, I'll absolutely bless it :-)

Cool.  

Will do the **subdivisions**, but first I need to make sure that #586 and #602 are merged/commented into beta branch because of the interdependencies with the documentation and I don't want to have to deal with the hell of trying to merge (or rebase?) commits back into my branch. I almost lost it today with #600, #601, and #602 (I think I need to take a class on `git`!)

What name do you prefer for the parameter: `subdivision=`,  `subdiv=` or `subd=`?

> Then - I believe - we could/should possibly review the current subdivision names to adhere to the reported ISO codes, but that can come at a later time.

Yup, later time.

dr-prodigy:
Hey there @mborsetti !
Regarding your question, I didn't have a chosen answer, to be honest :-)
[...]

Issue #548, opened 2021-11-08
Changing `province` to `state` for India


canute24:
Wassup doc, xD

In India `province` is called a `state`. I was wondering if you are open to making this change and what the deprecation process is like.

dr-prodigy:
Hey there @canute24 ! :)
thank you for your interest and suggestion! 👍 

Generally speaking, I don't feel this as a major improvement, although it definitely would make sense to apply it.
On the other hand, being a breaking change, if willing to implement it, it should be implemented in more than one steps to avoid issues on projects depending on python-holidays.

India is also one of the less test-covered countries, and therefore would require a quite deep tests review.

A good plan could be:

1) implement a first PR supporting both `prov` and `state`. In case `prov` is used, fire a deprecation warning and continue
2) merge to beta -> release a new lib version
3) remove `prov` support. In case `prov` is used, fire a "deprecated" error and ignore the province value
4) merge to beta -> release a new lib version

While implementing those changes, test coverage (currently scoring a poor 58%) should ideally be improved too.

Pls let me know if you're interested to work on this, I will do my best to support you.

Thx, cheers!
dr-p

dr-prodigy:
Hi @canute24 
I've recently pushed a similar update (class initialization review => upgrade with retro-compatibility + deprecation warning, relevant tests) to United Kingdom: it's currently released on beta branch.
Relevant commit here:
https://github.com/dr-prodigy/python-holidays/commit/fcdfda92bb059f92a8de56ddcf4c84fe89f0c80c

You could get ideas / code snippets from there, if you wish to implement your change on India respecting what suggested above.

KUTGW, thx!
dr-p

dr-prodigy:
This is now implemented / available on beta through the deprecation of `prov` and `state`, both replaced by `subdiv`.
Thx, cheers!

Issue #374, opened 2020-10-08
Support ISO 3166-2 codes for subdivisions of countries


mborsetti:
`python-holidays` only supports `prov` and `state` subdivisions, however states are subdivided into one or more from these categories:

- cantons
- commune
- councils
- counties
- departments
- dependencies
- districts
- divisions
- emirates
- entities
- governorates
- governorates
- islands, groups of islands
- Länder
- administrations
- municipalities
- parishes
- prefectures
- provinces
- rayons
- regions
- sectors
- states
- territories
- towns
- unitary authorities
- voivodships
- wards

Instead of expanding the types of subdivisions naming supported by this library, I propose adopting ISO 3166-2 (https://en.wikipedia.org/wiki/ISO_3166-2) which already standardizes all types of subdivisions into a non-overlapping namespace. 

I envision being able to initialize holidays for the state of California in the USA (ISO 3166-2 code US-CA) as follows:
`us_ca_holidays = holidays.US('CA')`,
for the Emirate of Dubai as follows:
`dubai_holidays = holidays.AE('DU')`,
for the Ağsu rayon in Azerbajan as follows:
`az-agu_holidays = holidays.AZ('AGU')`
and so on.

@dr-prodigy  I can work on a pull request [with backward compatibility] if I know it's going to be utilized, but won't work on it otherwise.

dr-prodigy:
Hey there @mborsetti 
sorry this proposal went unnoticed until now, my bad!

This is definitely a great idea, which would provide a definitely higher flexibility in the config/init phase.. especially if coupled to `.CountryHoliday('[CODE]')`. 

As long as we provide compatibility with previous versions, I totally agree with your analysis and would really appreciate your improvement.
thx loads 👍

dr-prodigy:
Released with #608 merge. Thx!

PR #608, opened 2022-01-22
implemented subdiv

mborsetti:
Implemented backward-compatible `subdiv` as per #597, #374 and addressing  #548.

A few notes:
* Code is documented and type-hinted.
* Documentation is updated.
* `prov` and `state` are deprecated but still work.
* If the HolidayBase class is instantiated with a `subdiv` argument that is not supported, a NotImplementedError is raised.
* If an instance is instantiated using the `holidays.country_holidays('XX')` method and the argument has a country code that is not supported, a NotImplementedError is raised (note: it used to raise KeyError, which is an incorrect Exception in this instance).
* The utility function `list_supported_countries` now returns a dict with all the ISO 3166-1 Alpha-2 country codes of the countries supported as keys and a list of that country's supported subdivision codes as values. This is backward-compatible with existing code that iterates over countries.
* Each country class has a new `subdivisions` variable listing all the subdivisions its supports. It replaces `PROVINCES`, which was initialized quite inconsistently (fixed).
* Each country class has a new `_deprecated_subdivisions` variable listing all deprecated and/or aliases of subdivisions that are supported. This variable is private and used only to avoid triggering NotImplementedError.
* Isle of Man has its own ISO country code of "IM" and was dutifully extracted from UnitedKingdom while maintaining backward compatibility.
* PortugalExt was collapsed into a subdivision of Portugal while maintaining backward compatibility.
* New Zealand was using the incorrect name and ISO subdivision code for "West Coast", which was fixed while maintaining backward compatibility.

Open issues discovered:
* United Kingdom of Great Britain and Northern Ireland (GB) currently has "UK" as a subdivision, but the country only has four subdivisions: England, Northern Ireland, Scotland, and Wales. This needs to be to be better documented and/or cleaned up; unclear as to what "UK" is in this context.
* The legacy code for adding HolidayBases together does some funky things. For example, adding Italy Catania (IT, CA) with USA Mississippi (US, MS) and USA Michigan (US, MI) yields country=["IT", "US"] and subdiv=["CA", "MS", "MI"], which is very ambiguous. Is it California and Messina and Milano, or Catania, Mississippi and Milano, or ... you get the picture.


dr-prodigy:
Merged now! 👍

PR #609, opened 2022-01-27
changed UAE weekends to SAT, SUN

marcomasulli:
Hi @dr-prodigy, UAE changed its weekend days to SAT and SUN starting on Jan 1st 2022. Updated the country file to reflect this. Let me know if any other change is needed. Thanks


dr-prodigy:
Hi @marcomasulli thank you for your work.
Unfortunately, it is not mergeable (all changes/fixes should start from beta branch, and therefore there are conflicts).
Also, UAE has just been reviewed by @mborsetti with #586 and #606.
Long story short, I'll be applying your changes again on the current code: soon available on beta branch.
Thx again, best

marcomasulli:
Thanks to you, apologies for not doing the PR into the beta branch in the
first place. Cheers!

mborsetti:
@dr-prodigy The code in this PR isn't going to work because it would apply the new definition retroactively. Look at the newly-merged saudi_arabia.py; there we now define `WEEKEND` within `_populate()` so we can change its value based on the year being populated.

dr-prodigy:
@mborsetti you are right!
In my implementation in fact I don't apply the original code (which was in fact wrong for the reason you say), but instead I added the new weekend logic to `_populate()` as you correctly suggest.
Thank you for pointing this out! 👍

PR #634, opened 2022-02-25
Fix Spain 'observed' date

javicalle:
The `_is_observed` funtion takes care about observed dates, so no need to add them as 'observed' dates.
The tests have also been fixed so that they run correctly. In this case, the validations have to be done on the `observed` dates.


dr-prodigy:
Thank you! 👍

PR #639, opened 2022-03-09
fix: Update Irish bank holidays

### LINKED ISSUE #636: Ireland considering UK as base class and hence not being a country itself
Issue also opened here:
home-assistant/core#67542

Looks like Ireland is being considered as being part of the UK which is wrong as not all the holidays in the UK exist, or necessarily exist in Ireland.

Take a reference on this comment: https://github.com/home-assistant/core/issues/67542#issuecomment-1058079650

TeoTN:
Fixes #636

### DIFF
--- a/holidays/countries/ireland.py
+++ b/holidays/countries/ireland.py
@@ -16,20 +16,37 @@ from datetime import date
 from dateutil.easter import easter
 from dateutil.relativedelta import relativedelta as rd, MO
 
-from holidays.constants import MAR, MAY, JUN, AUG, OCT, DEC
+from holidays.constants import FEB, MAR, MAY, JUN, AUG, OCT, DEC
 from holidays.constants import MON, TUE, WED, THU, FRI, SAT, SUN, WEEKEND
 from holidays.holiday_base import HolidayBase
-from .united_kingdom import UnitedKingdom
 
+class Ireland(HolidayBase):
+    """
+    Official holidays in Ireland, as declared in the Citizen's Information
+    bulletin:
+    https://www.citizensinformation.ie/en/employment/employment_rights_and_conditions/leave_and_holidays/public_holidays_in_ireland.html
+    """
 
-class Ireland(UnitedKingdom):
     country = "IE"
+    subdivisions = []
 
     def __init__(self, **kwargs):
         HolidayBase.__init__(self, **kwargs)
 
-    def _country_specific(self, year):
-        # Ireland exclusive holidays
+    def _populate(self, year):
+        self[date(year, JAN, 1)] = "New Year's Day"
+
+        # St. Brigid's Day
+        if year >= 2023:
+            dt = date(year, FEB, 1)
+            self[dt] = "St. Brigid's Day"
+
+            if self.observed and dt.weekday() != FRI:
+                self[date(year, FEB, 1) + rd(weekday=MO)] = "St. Brigid's Day (Observed)"
+
+        # One-off day of rememberance and recognition
+        if year == 2022:
+            self[date(year, MAR, 18)] = "Day of Rememberance and Recognition"
 
         # St. Patrick's Day
         name = "St. Patrick's Day"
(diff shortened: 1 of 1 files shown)


holidays/countries/ireland.py
```diff
@@ -66,11 +83,19 @@ def _country_specific(self, year):
...
         # October Bank Holiday (last Monday in October)
         self[date(year, OCT, 31) + rd(weekday=MO(-1))] = "October Bank Holiday"
 
+        # Christmas Day
+        name = "Christmas Day"
+        self[date(year, DEC, 25)] = name
+        if self.observed and date(year, DEC, 25).weekday() == SAT:
+            self[date(year, DEC, 27)] = name + " (Observed)"
+        elif self.observed and date(year, DEC, 25).weekday() == SUN:
+            self[date(year, DEC, 27)] = name + " (Observed)"
```
javicalle, line 97: Is this part correct?
* if 12/25 (SAT) --> 12/27 (MON)
* if 12/25 (SUN) --> 12/27 (TUE)

This can be simplified as:
```python
        if self.observed and date(year, DEC, 25).weekday() in [SAT,SUN]:
            self[date(year, DEC, 27)] = name + " (Observed)"
```

or maybe like:
```python
        if self.observed and date(year, DEC, 25).weekday() in [SAT, SUN]:
            self[date(year, DEC, 27) + rd(weekday=MO)] = name + " (Observed)"
```

javicalle (commented):
Some doubts about "Christmas Day"

holidays/countries/ireland.py
TeoTN: Thanks, updated in the latest commit

TeoTN:
> Some doubts about "Christmas Day"

@javicalle Thanks, updated the PR with new commits to fix the issue

TeoTN:
@javicalle Could you please review it?

holidays/countries/ireland.py
```diff
@@ -92,12 +92,14 @@ def _populate(self, year):
...
+            self[date(year, DEC, 25) + rd(weekday=MON)] = name + " (Observed)"
 
         # St. Stephen's Day
         name = "St. Stephen's Day"
         self[date(year, DEC, 26)] = name
-        if self.observed and date(year, DEC, 26).weekday() in WEEKEND:
+        if self.observed and date(year, DEC, 26).weekday() == SAT:
+            self[date(year, DEC, 26) + rd(weekday=MON)] = name + " (Observed)"
+        elif self.observed and date(year, DEC, 26).weekday() == SUN:
             self[date(year, DEC, 26) + rd(weekday=TUE)] = name + " (Observed)"
```
javicalle, line 103: Now I understand this part.
I think that could be correct as is, just that maybe I would use:
```python
        if self.observed and date(year, DEC, 26).weekday() in WEEKEND:
            self[date(year, DEC, 26) + rd(days=2)] = name + " (Observed)"
```
I'm not sure if it's more understandable or easier to maintain, so I'll leave it as a suggestion.

javicalle (approved):
The code part LGTM

TeoTN:
How can we get it through the finishing line?

holidays/countries/ireland.py
TeoTN: I like it being a little bit more verbose ;)

javicalle: It's fine to me 👍🏻

dr-prodigy:
Sorry, been not available for some weeks! On this now, will let you know 👍

dr-prodigy:
Okay some fixes (tests, annotations, formatting) required, but then all done.
Released on beta branch, thank you both! 👍

TeoTN:
Sweet, thanks a mil!

sotatech:
Any progress with this, I'm still getting the same error on 2022.5.3.

Thanks.

crowbarz:
There hasn't been a release of `holidays` since 15-Feb-2022, so the merged changes are still only present in the beta branch.

A workaround for Home Assistant (assuming that is what "2022.5.3" is referring to) involves copying the core `workday` integration into `custom_components` (see [here](https://developers.home-assistant.io/blog/2021/01/29/custom-integration-changes/#versions) for more info) and pinning the `holidays` dependency to 0.12, the last working version, in `manifest.json`. You'll need to also add a `version` key (any string will do) to the manifest as it is required by custom components since 2021.6.

sotatech:
Thanks, I thought I was posting in HA Core Github, but obviously not! I'll try reverting holidays back to 0.12 as you suggest.

PR #646, opened 2022-04-02
CI: If one job fails, let the others continue

hugovk:
For example, if one jobs fails to upload to Coveralls, it won't cancel all the others. 

Should help situations like this:

image

https://github.com/dr-prodigy/python-holidays/actions/runs/2081388552

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -21,6 +21,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     needs: [pre-commit]
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.7"]


dr-prodigy:
Hi @hugovk thank you for your suggestion and work, but I'm not really sure to understand the benefits: all the checks in the CD/CI should obtain a positive result to consider the new build as valid.
Could you explain a case in which this could not be required?
Thx!

hugovk:
Sure, here's a demo build where I've forced one job to fail ("3.8, ubuntu-20.04"), just for demonstration purposes.

It doesn't have a `fail-fast:` line so defaults to `true`.

image

https://github.com/hugovk/test/actions/runs/2085503760

What happened is, as soon as that job failed, all the other running jobs were cancelled. In this case, only one had managed to complete in time, so we have one pass, one fail, and a whole bunch of cancellations.

Would those cancellations have passed or failed?

---

Here's the same thing but with `fail-fast: false`.

image

This time, that one job failed, but the others were allowed to carry on running to completion. 

In both cases *the build as a whole* is a failure and the failure needs to be investigated.

But we now have a fuller picture. It could be we saw failures on all Windows jobs. Or failures on all Python 3.7 jobs. Or Python 3.7 and 3.8.

And that immediately gives us a better idea of what went wrong.

https://github.com/hugovk/test/actions/runs/2085500033

dr-prodigy:
Hi @hugovk
in fact, the aim of your change was already clear, but thanks for your explanation!

My point is:
a) if there's a code issue, all the jobs will likely fail, so in that case the first failure would be sufficient to locate and fix the error. At the moment this is reported via email by GH Actions in the shortest time.
b) if instead there's some library / CI issue, a single specific job could be the only one to fail.. but again, we will be looking there to sort it out, in the aim of producing a clean build (which is always our principle, when releasing a PR / code change, even in beta), so no big plus in letting the other jobs continue to run after the first issue.

This said, I really don't see big benefits.. on the other hand, this option is less respectful of the Github CD/CI computation costs (we would surely run more jobs than the required), so in the end this is not something I'd like to do..

If instead I've lost some other considerations on the way.. please tell me :-)

Thx again, KUTGW!
dr-p

hugovk:
Hi! Yeah, this can be useful for when a new Python feature is used (e.g. `str.startswith()`) and you'll just see failures in old versions, or some sort of path/encoding problem that makes Windows fail.

But sure, this isn't very important and it's a good point about computation resources, let's close this :) KUTGW as well!

PR #651, opened 2022-04-07
Feature/add nyse holidays

nadime:
Change to add NYSE holidays as an additional "financial markets" set of holidays.  These were still done within the "countries" infrastructure, but perhaps once a few more financial markets are added we should modify the structure slightly.

I've included only "full day" holidays for the NYSE historically and done my best to match with a few sources:
https://s3.amazonaws.com/armstrongeconomics-wp/2013/07/NYSE-Closings.pdf
and 
https://nyseholidays.blogspot.com/2012/11/nyse-holidays-from-<year1>-<year2>.html where year1/year2 are always a year ending in X1 and X0 (e.g. 1941-1950).

I've tried my best to conform to the standards I saw, for example adding " (Observed)" at the end of holidays when the observed holiday does not fall on the actual holiday, but please let me know if you see anything else you'd like changed.

Cheers!


mborsetti:
Very nice work!

Per https://github.com/dr-prodigy/python-holidays/issues/458, the [ISO 20022](https://www.iso20022.org/market-identifier-codes) code to be used here is `XNYS`.

@dr-prodigy not sure how you want to categorize these so they can grow (we already have something for European markets).

nadime:
Ah yes, of course - you're referring to the README, right (fixed there)?  Or did I miss somewhere else you want the right MIC?

nadime:
Are we waiting on anything here?  Can we merge this into beta?

dr-prodigy:
Hi @nadime thanks for your great work!
I in fact decided to apply some refactoring to allow for a better financial markets integration, adding a dedicated package (`financial`), adding a specific method for retrieving those classes only (`list_supported_financial()`) and reviewing tests accordingly.
I also added another synonym for the new class (`XNYS`) as suggested by @mborsetti.
Merged in beta, thx! 👍

PR #659, opened 2022-05-05
Add special cases for observing Labour Day in Greece

tudorvaran:
https://www.timeanddate.com/holidays/greece/labor-day
https://en.wikipedia.org/wiki/Public_holidays_in_Greece
From the wikipedia article:
> A public holiday that occurs on a Sunday is not transferred to another date, with the exception of 1 May, which is regarded by the locals more as a general strike than a public holiday.


dr-prodigy:
Brilliant! I've merged this into beta.
Thank you very much for your contribution @tudorvaran ! 👍

PR #661, opened 2022-05-22
Typechecking

dimbleby:
- Tighten up the typechecking on HolidayBase
- ditto holidays.utils
- then I did the UK, because it's nearby
- then I picked off a couple of the more egregious warnings where the same function was defined twice
- then I'd had enough, I certainly don't intend to go through every country...

### DIFF
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -14,10 +14,12 @@
 import warnings
 from datetime import timedelta, datetime, date
 from typing import (
+    cast,
     Any,
     Dict,
     Iterable,
     List,
+    Mapping,
     Optional,
     Set,
     TYPE_CHECKING,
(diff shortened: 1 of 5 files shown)


holidays/holiday_base.py
```diff
@@ -302,17 +306,16 @@ def __contains__(  # type: ignore[override]
           :func:`dateutil.parser.parse`,
         * or a :class:`float` or :class:`int` representing a POSIX timestamp.
         """
-        return dict.__contains__(self, self.__keytransform__(key))
+        if not isinstance(key, (date, datetime, str, float, int)):
+            raise TypeError("Cannot convert type '%s' to date." % type(key))
```
dimbleby, line 310: IMO `return False` might be better here, but you have unit tests checking for the current behaviour

dimbleby:
(need a more recent version of black to fix the workflow)

holidays/holiday_base.py
dr-prodigy: Hi @dimbleby I understand / agree on the check on the date format -> exception, but why did you remove the return value (True / False) from the __contains__ definition? :-/

dr-prodigy: Oh right, it was hidden by getitem code in fact, sorry!
Now I see your point: I definitely prefer the date format check as it is, ie: raise an exception in case of unknown format 👍

dimbleby:
rebased this - and #662 - on the beta branch so as to pick up updated black and in the expectation that the workflow will now pass

PR #620, opened 2022-02-10
mypy implementation

mborsetti:
* Implemented [mypy](http://mypy-lang.org/) static type checker in pre-commit
* Updated pre-commit to current versions ($ [pre-commit autoupdate](https://pre-commit.com/#pre-commit-autoupdate))
* Added `py.typing` file to implement [PEP 561 packaging type information](https://www.python.org/dev/peps/pep-0561/#packaging-type-information)
* Minor cleanup of errors in static typing / incompatibilities with mypy and addition of `# type: ignore`

PR #662, opened 2022-05-22
include py.typed in published package

dimbleby:
The py.typed that was added at #620 is not being published, so no users of this library can take advantage of the types that it exposes.


dimbleby:
>  Coverage increased (+0.0004%) to 98.939% 
got to love coveralls!

PR #667, opened 2022-06-04
Update README.rst (can be better grammar)

ajsharda17:


dr-prodigy:
I personally don't see a real value in this PR. Moreover, even if there is no written law on that, it makes the description depart from the "standard" format used in pypi. Due to this, I'm not merging it, sorry.

PR #671, opened 2022-06-05
Fix missing holidays.financial

synapticarbors:
This addresses https://github.com/dr-prodigy/python-holidays/issues/670 which is also breaking the conda-forge build.


dr-prodigy:
Thank you! 👍

PR #676, opened 2022-06-26
Make Colombian holidays preserve historical observation

bkthomps:
Colombia essentially has four types of holidays:
1. fixed non-easter based
2. flexible non-easter based
3. fixed easter based
4. flexible easter based

The easter based holidays depend on when easter occurs. The flexible holidays occur on the next Monday after the date the holiday falls on, unless it already falls on a Monday.

This change does the following:
1. fixes non-easter based holidays not being set (ie: new year's wasn't being set for 2022: https://github.com/dr-prodigy/python-holidays/issues/674).
2. I rigorously went through the Colombian laws and decrees about holidays and labour since 1905 (wasn't precisely codified until 1939) and went through examples of physical calendars to verify this. Proof is in the code as comments.
3. I improved the test cases to increase confidence in the code correctness.

This branch was rebased off master as instructed

### DIFF
--- a/holidays/countries/colombia.py
+++ b/holidays/countries/colombia.py
@@ -26,7 +26,15 @@ class Colombia(HolidayBase):
         HolidayBase.__init__(self, **kwargs)
 
     def _add_with_bridge(self, _date, name):
-        if self.observed and _date.weekday() != MON and _date.year >= 1983:
+        """
+        On the 6th of December 1983, the government of Colombia declared which
+        holidays are to take effect, and also clarified that a subset of them
+        are to take place the next Monday if they do not fall on a Monday.
+        This law is "Ley 51 de 1983" which translates to law 51 of 1983.
+        Link: https://www.funcionpublica.gov.co/eva/gestornormativo/norma.php?i=4954#:~:text=Determina%20los%20d%C3%ADas%20festivos%20en,remuneraci%C3%B3n%20del%20d%C3%ADa%20festivo%20art.
+        """
+
+        if self.observed and _date.weekday() != MON and _date.year > 1983:
             self[_date + rd(weekday=MO)] = name + " (Observed)"
         else:
             self[_date] = name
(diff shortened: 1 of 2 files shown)


jorgeegomez:
@bkthomps Could you please reconsider the weekends code for the six fixed Colombian holidays? I think the day should be considered a holiday even if it's Saturdy or Sunday. Lots of people work on Saturday and Sunday, and it makes a difference whether the day is a holiday or not. 

The current law can be found [here](https://www.funcionpublica.gov.co/eva/gestornormativo/norma.php?i=4954)

bkthomps:
> @bkthomps Could you please reconsider the weekends code for the six fixed Colombian holidays? I think the day should be considered a holiday even if it's Saturdy or Sunday. Lots of people work on Saturday and Sunday, and it makes a difference whether the day is a holiday or not.
> 
> The current law can be found [here](https://www.funcionpublica.gov.co/eva/gestornormativo/norma.php?i=4954)

Can you elaborate on what is incorrect please? Most holidays move the the next Monday, but some area always on the day which they fall on (namely those mentioned in section 1 but not section 2).

jorgeegomez:
I think Sundays (and even more so Saturdays) should be included in the list of holidays when they're a fixed-date holiday.

In other words, I think `len(holidays.CO(years=y).items()` should be `18` for any year (after 1986 at least).

It's factually incorrect to say that 2022-01-01 or 2024-12-08 are not a holiday in Colombia.

bkthomps:
> I think Sundays (and even more so Saturdays) should be included in the list of holidays when they're a fixed-date holiday.
> 
> In other words, I think `len(holidays.CO(years=y).items()` should be `18` for any year (after 1986 at least).
> 
> It's factually incorrect to say that 2022-01-01 or 2024-12-08 are not a holiday in Colombia.

Unless I'm missing something, this is what I've change the code to do. 2021-01-01 is considered a holiday in the test cases.

dr-prodigy:
Perfect work! Merged in beta, thx! 👍

PR #677, opened 2022-06-26
Make Venezuela holidays preserve historical observation

bkthomps:
I rigorously went back through the historical records by analyzing laws and decrees until 1909 to get an accurate representation of what holidays there were over the years. I added many more test cases to increase confidence in the code. I put the proof for each year I picked in the code.

For some holidays, such as declaration of independence, I put conservative dates such as when they declared independence because I could not find any laws or decrees in the historical record before 1909.

There is one holiday that was in a decree in 1909 and was overturned in 1918. I tried my best to find out what it was called, but could not unfortunately.

### DIFF
--- a/holidays/countries/venezuela.py
+++ b/holidays/countries/venezuela.py
@@ -32,12 +32,12 @@ class Venezuela(HolidayBase):
         """
         Overview: https://dias-festivos.eu/dias-festivos/venezuela/#
         Various decrees about holidays:
-          1909 (AUG 5): https://www.guao.org/sites/default/files/efemerides/69.Ley%20fiestas%20nacionales%201909.pdf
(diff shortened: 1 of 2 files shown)


holidays/countries/venezuela.py
```diff
@@ -41,47 +29,79 @@ def __init__(self, **kwargs):
...
-
-        # Christmas Day
-        self[date(year, DEC, 24)] = "Nochebuena"
-
-        self[date(year, DEC, 25)] = "Día de Navidad"
-
-        self[date(year, DEC, 31)] = "Fiesta de Fin de Año"
-
-        # Semana Santa y Carnaval
-        if date(year, APR, 19) == (easter(year) - rd(days=2)):
```
bkthomps, line 67: This logic is handled already.

dr-prodigy:
Thank you very much for this too! Merged 👍

PR #678, opened 2022-06-26
Add Cuba

bkthomps:
All primary and secondary sources in comments in the code. Could only find information for 1969 and onwards, so these holidays only apply to Socialist/Modern Cuba and not Batista's Cuba nor pre-Batista Cuba.


dr-prodigy:
Brilliant! Merged in beta, thank you very much 👍

PR #692, opened 2022-08-21
Add Pandas submodule

YusufBerki:
Added 2 functions to convert HolidayBase object into Pandas Series or Pandas DataFrame.

These functions were not added to the HolidayBase class as a method
in order not to require Pandas in library installation and use.
So only those who want to use these functions will have to install Pandas.

**Usage:**
```python
>>> import holidays
>>> us = holidays.US(years=[2022])

>>> from holidays.pandas import holidays_to_series
>>> holidays_to_series(us)
```
    2022-01-01                                     New Year's Day
    2022-01-17                         Martin Luther King Jr. Day
    2022-02-21                              Washington's Birthday
    2022-05-30                                       Memorial Day
    2022-06-19               Juneteenth National Independence Day
    2022-06-20    Juneteenth National Independence Day (Observed)
    2022-07-04                                   Independence Day
    2022-09-05                                          Labor Day
    2022-10-10                                       Columbus Day
    2022-11-11                                       Veterans Day
    2022-11-24                                       Thanksgiving
    2022-12-25                                      Christmas Day
    2022-12-26                           Christmas Day (Observed)
    Name: holiday, dtype: object

### DIFF
--- a/holidays/pandas.py
+++ b/holidays/pandas.py
@@ -1,0 +1,70 @@
+from holidays.holiday_base import HolidayBase
+import pandas as pd
+
+
+def holidays_to_series(cls: HolidayBase, name: str = 'holiday') -> pd.Series:
+    """Get Pandas Series from HolidayBase object
+
+    Example usage:
+        >>> from holidays.pandas import holidays_to_series
+        >>> holidays_to_series(holidays.US(years=[2022]))
+            2022-01-01                                     New Year's Day
+            2022-01-17                         Martin Luther King Jr. Day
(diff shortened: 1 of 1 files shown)


dr-prodigy:
Hi @YusufBerki , sorry, I have to agree with @arkid15r : although well documented, your PR doesn't add so much value, but instead enforces a dependency (even if not strict) to Pandas which, not being really required, at the moment we would rather avoid to add.
Therefore, we won't receive it as it is: we could reconsider it in case of more valuable additions.

Thanks all the same, cheers

PR #715, opened 2022-09-16
Fix observed holiday for christmas and boxing day canada

bkthomps:
For Canada, Christmas and Boxing day were now showing the non-observed day, unlike the other holidays for Canada.

### DIFF
--- a/holidays/countries/canada.py
+++ b/holidays/countries/canada.py
@@ -252,20 +252,16 @@ class Canada(HolidayBase):
         # Christmas Day
         if year >= 1867:
             name = "Christmas Day"
-            name_observed = name + " (Observed)"
+            self[date(year, DEC, 25)] = name
             if self.observed and date(year, DEC, 25).weekday() in WEEKEND:
-                self[date(year, DEC, 27)] = name_observed
-            else:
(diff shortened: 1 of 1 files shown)


bkthomps:
Could you change the coveralls failure threshold to 0.1% or something instead of 0% please?

image

dr-prodigy:
Hi @bkthomps ,
thank you for your contribution.
Regarding your last request (failure threshold) it is in fact meant to keep coverage level at the pre-existing state, so that's why we require a 0% loss to pass checks ( = all reviewed / added code should be covered).
Could you please review your PR to integrate unit tests for all your changes?
Please let me know if you are experiencing inconsistencies in the coverage calculation and check, thank you!

bkthomps:
@dr-prodigy the test coverage is still on those lines, but since I removed 2 lines it's trivially less.

The coverage decrease is -4.0e-05% so -0.00004% less. Can we make the minimum some small delta at least to prevent this case? Like 0.01% or something?

dr-prodigy:
Hi @bkthomps, thank you for your notes and suggestions.. I've updated the coverage threshold and merged your PR in beta.
KUTGW

PR #720, opened 2022-09-17
Fix honduras

bkthomps:
Honduras had weird behavior with self.observed and did not follow the Morazan Weekend correctly since it is the first week of October from wednesday to friday (technically noon wednesday to noon saturday).


bkthomps:
All code paths should be covered. I think the slight decrease is since I removed lines.

dr-prodigy:
Merged in beta, thank you very much!

PR #721, opened 2022-09-17
Rename Eswatini from Swaziland

bkthomps:
This country has a different name now (but keep it backwards-compatible).

### DIFF
--- a/holidays/countries/__init__.py
+++ b/holidays/countries/__init__.py
@@ -89,7 +89,7 @@ from .slovakia import Slovakia, SK, SVK
 from .slovenia import Slovenia, SI, SVN
 from .south_africa import SouthAfrica, ZA, ZAF
 from .spain import Spain, ES, ESP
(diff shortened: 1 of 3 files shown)


holidays/countries/__init__.py
```diff
@@ -89,7 +89,7 @@
 from .slovenia import Slovenia, SI, SVN
 from .south_africa import SouthAfrica, ZA, ZAF
 from .spain import Spain, ES, ESP
-from .swaziland import Swaziland, SZ, SZW
+from .eswatini import Eswatini, Swaziland, SZ, SZW
```
arkid15r, line 92: It seems that the imports need to be sorted alphabetically.

arkid15r (approved):
LGTM

holidays/countries/__init__.py
bkthomps: Done.

dr-prodigy:
Merged.. Thx!

PR #726, opened 2022-09-29
Singapore: Revised Date for Vesak Day in 2023

mborsetti:
Changed Vesak Day from 3 June 2023 to 2 June 2023 as per https://www.mom.gov.sg/newsroom/press-releases/2022/0929-revised-date-for-vesak-day-2023:
> Following the [Singapore Buddhist Federation’s announcement](https://www.buddhist.org.sg/) on the revised date for Vesak Day next year, the Ministry of Manpower (MOM) has gazetted 2 June 2023 as the new public holiday for Vesak Day.

Singapore Buddhist Federation announcement (🤦):
> The Singapore Buddhist Federation earlier advised the Ministry of Manpower that Vesak Day will fall on 3 June 2023. This was determined in reference to the Standard Chinese-Western Multi-Year Almanac.
> 
> Upon receiving feedback from members of the public, we have verified the date against other Chinese almanacs and have established that based on the Mahayana tradition, the date of Vesak Day 2023 shall fall on June 2, 2023 which coincides with the 15th day of the Fourth Lunar Month.
> 
> We are sorry for any inconvenience caused and will in future conduct additional checks and verifications to prevent similar occurrence.

Also fixed code to remove the assignment of objects of different type to the same variable name, which `mypy` in `pre-commit` picked up, by restricting the reuse of a variable name to objects of the same type.


holidays/countries/singapore.py
```diff
@@ -96,7 +96,7 @@ def _populate(self, year):
         # Hari Raya Puasa
         # aka Eid al-Fitr
         # date of observance is announced yearly
-        dates_obs = {
+        dates_fixed_multiple_obs: Dict[int, List[Tuple[int, int]]] = {
```
arkid15r, line 99: It seems that a tuple would be sufficient here (and below): Dict[int, Tuple[Tuple[int, int]]].

It also looks like the dict can be simplified to Dict[int, Tuple[int, int]] format.

arkid15r (commented):
Hi Mike,
Thanks for the contribution!

holidays/countries/singapore.py
mborsetti: Yeah, so I thought at first glance, forgetting in the moment why I had coded it this way in the first place.

I quickly remembered that the Islamic calendar year is shorter than the solar Gregorian calendar year by 10 to 12 days so there is a chance that these Islamic holidays can fall twice in the same Gregorian year, which I stumbled upon when doing extensive code testing.  So it cannot be simplified.

I will add a remark so we won't forget in the future!

Thanks.

mborsetti: Oh missed the part suggesting using a tuple instead of a list! Makes a ton of sense; will prepare a PR for it.

mborsetti:
@dr-prodigy anything more needed here?  It (started as) a simple date fix due to the announcement referenced in the first post.

dr-prodigy:
Hey Mike @mborsetti nope, not really anything more .. I'm just been very late at receiving updates in these weeks!
Now merged, thx 👍

PR #729, opened 2022-10-02
Split out extra packages

### LINKED ISSUE #362: Reduce mandatory packages bloat
I installed the package in a new machine and it installed the following seldom-used packages:

`pymeeus, convertdate, korean-lunar-calendar`

Any work to reduce (and/or standardize) the number of packaged installed would be helpful to combat package bloat.  For example, I have no need for Korean holidays yet I find myself having to install a package that, by its name, is required only for those type of holidays.

stianjensen:
Split out two dependencies that are only used for some specific countries so they're not installed by default. This is a breaking change for consumers using one or more of the affected countries. The advantage is that consumers who only use one or a few simple calendars don't need to install dependencies that will be unused.

This fixes #362.


stianjensen:
Open for comments on exactly which packages to split out (and what to call them). Also on the best way to put this in the documentation.

And I guess this should be made into a breaking change – since we're pre-1.0 I don't know what the best way to do that is.

dr-prodigy:
Hi @stianjensen 
this is long-waited task, so thank you very much for taking it over!
I could not look at it extensively (I've applied a couple of conflict fixes, basically), but unfortunately, apart from being a breaking change, I'm not sure it would pass all the required tests, nor allow for a 100% test coverage.
Then, I would kindly ask you to review it, ensuring to run
`tox`
and check its results when you're done with everything.
Thank you very much again, KUTGW

stianjensen:
All tests pass when running locally now, however I see that coverage has dropped from 100% on the files that have custom handling of missing packages. I'll look at adding test coverage to those somehow.
(Tests currently pass since they run with dependencies from requirements_dev.txt, which means mis-configurations of which package goes in which "extras" packages aren't currently picked up by the test suite). I'll look at other projects for prior-art on whether this can be added to the test suite somehow.

arkid15r:
Continued in https://github.com/dr-prodigy/python-holidays/pull/889

PR #733, opened 2022-10-07
Fixed incorrect subdivision code

MichaelThessel:
Looks like the docs have a typo for the Yukon Territory (YU should be YT). In the code this seems to be correct.

https://en.wikipedia.org/wiki/ISO_3166-2:CA


dr-prodigy:
Thank you!

PR #734, opened 2022-10-10
Keep context based import sorting for `holidays.constants`.

### LINKED ISSUE #732: Improve `holidays.constants` multi-line import sorting
Keep week days and months multi-line imports ordered according to the date/time context (which is more readable taking into account the nature of python-holidays package).

```I would have dreamt to see those holidays.constants multi-line import move to the isort format (definitely nicer and more readable than the black's version).```

_Originally posted by @dr-prodigy in https://github.com/dr-prodigy/python-holidays/issues/722#issuecomment-1268560210_

### PR DESCRIPTION
**Pull request by arkid15r**
Resolves #732 

As suggested in the issue above the week day and month constants should be kept in their context order (MON, TUE, WED, ...SUN / JAN, FEB, MAR, ... DEC)

This PR adds `no_inline_sort = true` to the setup.cfg in order to skip isort inline sorting. It means this kind of imports should be **enforced manually on code reviews**.

### CODE REVIEW

**Comment by dr-prodigy** [comment]
Hi Ark, good job, thank you!
A couple of notes:
- I would have hoped to see cases like this
```
from holidays.constants import JAN, APR, MAR, MAY, JUN, JUL, AUG, SEP, OCT, NOV, DEC
from holidays.constants import FRI, SAT, SUN, WEEKEND
```
not converted to this
```
from holidays.constants import (
    JAN,
    APR,
    MAR,
    MAY,
    JUN,
    JUL,
    AUG,
    SEP,
    OCT,
    NOV,
    DEC,
    FRI,
    SAT,
    SUN,
    WEEKEND,
)
```
but that's fine as a first solution.
- I had to accordingly review `pyproject.toml` as well: don't forget that's the one which is used when running local tests through
`tox`
command.

About to merge, thank you again! 👍

PR #735, opened 2022-10-11
Remove not available --update option when installing

NicoHood:
$pip install --update holidays

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --update

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,7 +47,7 @@ The latest stable version can always be installed or updated via pip:
 
 .. code-block:: bash
 
-    $ pip install --update holidays
+    $ pip install holidays
 
 
 Documentation


README.rst
```diff
@@ -47,7 +47,7 @@ The latest stable version can always be installed or updated via pip:
 
 .. code-block:: bash
 
-    $ pip install --update holidays
+    $ pip install holidays
```
arkid15r, line 50:
```suggestion
    $ pip install --upgrade holidays
```

arkid15r (approved):
I believe the `--upgrade` option was intended originally and it should just replace current `--update` (invalid) flag.

Good catch, thank you!

README.rst
NicoHood: That might be the case, but why would you want the user to force an upgrade if already installed?

dr-prodigy:
Just released a fix in beta for this .. thank you both! :+1:

PR #744, opened 2022-10-17
Fix Honduras Morazan weekend bug.

arkid15r:
Add test for 2025 to cover this case (when Oct 1st is the first Wed of the month)

### DIFF
--- a/holidays/countries/honduras.py
+++ b/holidays/countries/honduras.py
@@ -70,9 +70,7 @@ class Honduras(HolidayBase):
             # Morazan Weekend
             # (First Wednesday of October from 12 noon to Saturday 12 noon)
             holiday_name = "Semana Morazánica [Morazan Weekend]"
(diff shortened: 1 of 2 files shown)


holidays/countries/honduras.py
```diff
@@ -71,7 +71,7 @@ def _populate(self, year):
             # (First Wednesday of October from 12 noon to Saturday 12 noon)
             holiday_name = "Semana Morazánica [Morazan Weekend]"
             first_wednesday = date(year, OCT, 1)
```
KJhellico, line 73: Why isn't just
`first_wednesday = date(year, OCT, 1) + rd(weekday=WE(+1))`
here?

arkid15r: Looks legit, thanks for the suggestion!

dr-prodigy:
Thx @arkid15r + @KJhellico , great job 👍  Merged in beta.

PR #746, opened 2022-10-19
Added observed holidays for Ukraine

KJhellico:
Added observed holidays for Ukraine and tests (last years and full for observed dates)

### DIFF
--- a/holidays/countries/ukraine.py
+++ b/holidays/countries/ukraine.py
@@ -38,7 +38,7 @@ from holidays.holiday_base import HolidayBase
 class Ukraine(HolidayBase):
     """
     Current holidays list:
-    http://zakon1.rada.gov.ua/laws/show/322-08/paran454#n454
+    https://zakon1.rada.gov.ua/laws/show/322-08/paran454#n454
 
     27.01.1995: holiday on weekend move to next workday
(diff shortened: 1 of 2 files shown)


KJhellico:
Uncovered code is for case when Easter is Apr 30, and nearest such day will be in 2062. Should I add test for it?

holidays/countries/ukraine.py
```diff
@@ -12,15 +12,42 @@
...
+    SUN,
+    WEEKEND,
+)
 from holidays.holiday_base import HolidayBase
 
 
 class Ukraine(HolidayBase):
     """
+    Current holidays list:
     http://zakon1.rada.gov.ua/laws/show/322-08/paran454#n454
```
arkid15r, line 41:
```suggestion
    https://zakon1.rada.gov.ua/laws/show/322-08/paran454#n454
```

arkid15r (approved):
> Uncovered code is for case when Easter is Apr 30, and nearest such day will be in 2062. Should I add test for it?

The more coverage the better. Go for it!

This PR looks great, thanks for the effort!

KJhellico:
Seems tests is OK

dr-prodigy:
Thx :+1:

PR #748, opened 2022-10-20
Fixed Easter holidays for Bulgaria

KJhellico:


dr-prodigy:
Thank you very much @KJhellico! :+1:

PR #753, opened 2022-10-21
Portugal: add district holidays

Nalguedo:
Added holidays for the 18 portuguese districts
Added tests
Updated README.rst


thylux:
@Nalguedo Wouldn't it be better to use the names of the districts instead of numbers?

Nalguedo:
> @Nalguedo Wouldn't it be better to use the names of the districts instead of numbers?

I can add the names, but I've followed the same approach implemented for other countries.
https://en.wikipedia.org/wiki/ISO_3166-2:PT

I leave the decision to @dr-prodigy

dr-prodigy:
Hi @thylux
although I definitely understand your reasons (readability), as a rule of thumb, whenever a standard ISO code is available, I'm more in favour of it, as opposed to using a long description.
So I'll merge @Nalguedo 's version as is.
Thank you both, cheers!

PR #756, opened 2022-10-21
Optimizations for Ireland

KJhellico:


dr-prodigy:
Thank you @KJhellico ! :+1:

PR #769, opened 2022-10-26
Weekdays/weekends refactoring

arkid15r:
This PR streamlines weekday/weekend constants usage and addresses some isort/flake8 issues:
  - Deprecate MON-SUN constants in favor of dateutil MO-SA.
  - Deprecate and move `WEEKENDS` from utils to HolidayBase (on per country basis as some countries have not usual SAT, SUN weekends).
  - Add `_is_weekend` method into `HolidayBase`.
  - Change isort multi-line import behavior to bracket-less.
  - Update flake8 settings in order to make it stricter.
  - Remove unused imports.
  - Remove unused/commented out code.


holidays/countries/hongkong.py
```diff
@@ -74,19 +59,24 @@ def _populate(self, year):
         new_year_date = date(dt.year, dt.month, dt.day)
         if self.observed:
             self[new_year_date] = name
-            if new_year_date.weekday() in [MON, TUE, WED, THU]:
+            if new_year_date.weekday() in (
+                MO.weekday,
+                TU.weekday,
+                WE.weekday,
+                TH.weekday,
+            ):
```
KJhellico, line 67:
```suggestion
            if new_year_date.weekday() <= FR.weekday:
```

KJhellico (changes requested):

holidays/financial/ny_stock_exchange.py
```diff
@@ -181,7 +181,7 @@ def _populate(self, year):
                 begin + timedelta(days=n)
                 for n in range((end - begin).days + 1)
             ):
-                if self._is_weekend(d):
+                if d.weekday() in {SA.weekday, SU.weekday}:
```
KJhellico, line 184: Why not `_is_weekend(d)`? Here and two changes below

dr-prodigy:
While the _is_weekend(d) improvement is very valuable, I feel like moving to dateutil for day names isn't really beneficial (even if I understand reusing standard names has some advantages, "MO.weekday" is longer and probably even less clear than a quick "MON", IMHO!).
Would you mind keeping the first part of this PR only @arkid15r ?
Thx!

holidays/financial/ny_stock_exchange.py
arkid15r: Just a merge error, fixed.

arkid15r:
> Would you mind keeping the first part of this PR only @arkid15r ?

I was trying to move from redundant imports here (e.g. MO from dateutil, MON from constants). This problem may soon be resolved without this PR. So yeah, I can revert weekdays to the original version.

arkid15r:
PTAL.

dr-prodigy:
@arkid15r today I've sent an update on the related PR #841 (unfortunately, I'm not in favor of its merge).. do you need any additional comments on this too?

arkid15r:
> do you need any additional comments on this too?

Not sure what you mean, tbh. My impression was that you had requested to revert MO-SA refactoring (keeping MON-SUN instead) which was done. The `_is_weekend` method was "very valuable" according to your previous comment and is still there.

I you have additional comments please add them. I'll do my best to address those as well.

dr-prodigy:
Hi @arkid15r okay, now it's clearer thx! 
When I read "PTAL" I thought you were possibly waiting for further answers and/or notes from me, so I just asked a confirmation to be safe :-)

Being this quite a large refactoring (89 changed files) and not having seen any further note from you, it was not obvious for me at which point of the agreed reviews you were, but given you're done with everything we agreed on , I'll be fixing the arisen conflicts and merging this ASAP.  

Thx again, KUTGW :+1:

arkid15r:
I see, sorry for being ambiguous. Yes, I just wanted to request another review round from you.

Thank you!

Issue #768, opened 2022-10-25
Optimize Kst day lookup process


arkid15r:
Replace while loops with `dateutil` logic:

The list of code parts to optimize:
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/angola.py#L63
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/botswana.py#L57
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/mozambique.py#L38
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/south_africa.py#L138
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/south_africa.py#L158
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/south_africa.py#L169
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/south_africa.py#L176
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/zimbabwe.py#L56
  - https://github.com/dr-prodigy/python-holidays/blob/a772f79327a64639ec1fe89c79165d585c3a9b3d/holidays/countries/zimbabwe.py#L67

arkid15r:
Hey @KJhellico, would you like to look into this? 
For some reason I though may be interested :)

KJhellico:
Definitely! 😋 I will look into it

dr-prodigy:
#770 merged right now.. thank you both! 👍

PR #770, opened 2022-10-26
Replace while loops with dateutil logic (#768)

KJhellico:


arkid15r (changes requested):
This looks good in general and the direction is 100% correct.

However I don't think that current tests cover these cases in a sufficient manner. Could you add related tests for all affected countries and make sure they are green before and after the refactoring? 

It's important to have efficient code but compatibility/reliability takes priority here. I hope you understand my point.

Thank you @KJhellico!

KJhellico:
> It's important to have efficient code but compatibility/reliability takes priority here. I hope you understand my point.

Yes, I noticed that there are not enough tests (there was wrong calculation of Queen's Birthday in South Africa and Defense Forces Day in Zimbabwe). I will try to add more.

arkid15r:
> Yes, I noticed that there are not enough tests (there was wrong calculation of Queen's Birthday in South Africa and Defense Forces Day in Zimbabwe). I will try to add more.

Let me know if you need help with that.

KJhellico:
I found that South Africa observed calculation was incorrect (if holiday fall on a Sunday, only followed Monday should be holiday, not followed workday). Should I change it in this PR or make separate? @arkid15r

KJhellico:
I removed SA from this PR, and added tests for affected holidays in rest of countries

arkid15r (approved):
It looks like you've fixed Zimbabwe heroes day bug with this refactoring, well done!
Looks good to me!

PR #771, opened 2022-10-27
Rework Norway and Sweden logic.

arkid15r:
- Remove `assert`s
  - Rework all Sundays collection logic (use dateutil.rrule). 
  - Extract repetitive test_sundays code to a separate file.


dr-prodigy:
Thank you Ark! 👍

PR #772, opened 2022-10-28
After run tox

poshingchu:
Added special case for 2022 Chinese Mid-Autumn Festival


dr-prodigy:
This is completed in #779 (merged), so I'm closing it.. thx! 👍

Issue #727, opened 2022-09-30
South Africa rollover rules regarding holidays falling on Sundays don't appear to be correct


bryndivey:
#392 changed South African holiday calculations to add observed holidays when the existing holiday falls on a Sunday, per:

```The Public Holidays Act (Act No 36 of 1994) determines that whenever any public holiday falls on a Sunday, the Monday following on it shall be a public holiday.```

However it doesn't just check if it can add it to the following Monday but continues checking for an "open" day until it can place it, leading to it incorrectly marking 27 December 2022 as a public holiday:

```
In [80]: list(sorted(holidays.ZA(years=2022).items()))
Out[80]:
[(datetime.date(2022, 1, 1), "New Year's Day"),
 (datetime.date(2022, 3, 21), 'Human Rights Day'),
 (datetime.date(2022, 4, 15), 'Good Friday'),
 (datetime.date(2022, 4, 18), 'Family Day'),
 (datetime.date(2022, 4, 27), 'Freedom Day'),
 (datetime.date(2022, 5, 1), "Workers' Day"),
 (datetime.date(2022, 5, 2), "Workers' Day (Observed)"),
 (datetime.date(2022, 6, 16), 'Youth Day'),
 (datetime.date(2022, 8, 9), "National Women's Day"),
 (datetime.date(2022, 9, 24), 'Heritage Day'),
 (datetime.date(2022, 12, 16), 'Day of Reconciliation'),
 (datetime.date(2022, 12, 25), 'Christmas Day'),
 (datetime.date(2022, 12, 26), 'Day of Goodwill'),
 (datetime.date(2022, 12, 27), 'Christmas Day (Observed)')]```

bryndivey:
@pietervdw115 since you submitted the original change just want to check if there's something I'm missing here?

dr-prodigy:
Issue fixed and released. Closing, thx!

PR #773, opened 2022-10-28
South Africa: updates

KJhellico:
- replace while loops with dateutil logic (#768)
- Queen's Birthday day fix
- observed holidays calculation fix (#727)
- added tests


dr-prodigy:
Thank you very much @KJhellico , merged ! 👍

PR #777, opened 2022-11-04
Replace Python built-in asserts with unittest asserts.

arkid15r:
Migrate `assert` from holidays/ to test/


dr-prodigy:
Well done Ark! Some more good code standardization.. Merged, thx 👍

PR #779, opened 2022-11-05
Added special case on 2022 Hong Kong

poshingchu:
Added special case on 2022 Chinese Mid-Autumn Festival

Added test case on test_hongkong.py

Run tox and passed pre-commit


dr-prodigy:
Thank you very much @poshingchu ! 👍

magggzzz:
@poshingchu thank you very much for your update, but the day of mid autumn itself , it not a holiday at all , only the day after is a holiday , so the day of mid autumn , should'nt be on the holiday list , its not a public holiday, thank you

PR #798, opened 2022-11-16
Zambia: optimizations, added special holidays

KJhellico:


holidays/countries/zambia.py
```diff
@@ -27,70 +28,77 @@ class Zambia(HolidayBase):
...
+            (JUL, 7, "Funeral of Kenneth Kaunda"),
+            (AUG, 12, "General elections"),
+            (AUG, 13, "Counting in general elections"),
+            (AUG, 24, "Presidential inauguration"),
+        ),
+        2022: ((MAR, 18, "Funeral of Rupiah Banda"),),
+    }
 
     def _populate(self, year):
         super()._populate(year)
```
arkid15r, line 56: I'd move it below the line 60.

arkid15r (commented):
LGTM

holidays/countries/zambia.py
KJhellico: > I'd move it below the line 60.

`special_holidays` declaration?

arkid15r: Nope :).
I see the weird code selection here too, my bad. 

I meant `super()._populate(year)` line. It should be executed after the year condition check is satisfied.

KJhellico: Done.

dr-prodigy:
Thank you! KJ 👍

PR #799, opened 2022-11-16
Vietnam: optimizations, observed holidays calculation

KJhellico:


holidays/countries/vietnam.py
```diff
@@ -41,53 +42,40 @@ def _populate(self, year):
...
-            king_hung_date = date(dt.year, dt.month, dt.day)
-            self[king_hung_date] = name
-        else:
-            pass
+            king_hung_date = self.get_solar_date(year, 3, 10)
+            self[king_hung_date] = "Hung Kings Commemoration Day"
 
         # Liberation Day/Reunification Day
-        name = "Liberation Day/Reunification Day"
         libration_date = date(year, APR, 30)
```
arkid15r, line 69: As the date objects being used once only I'd  optimize them with `self[date(year, APR, 30)] = "Liberation Day/Reunification Day"` approach.

arkid15r (commented):
👍

holidays/countries/vietnam.py
KJhellico: Done. Thank you!

dr-prodigy:
👍

PR #801, opened 2022-11-18
Optimize `if var in iterable:` statements.

arkid15r:
Replace a tuple/list with a set in order to achieve O(n) -> ~O(1) complexity improvement.


holidays/countries/israel.py
```diff
@@ -61,7 +62,7 @@ def _populate(self, year):
         observed_delta = 0
         if self.observed:
             day_in_week = memorial_day_dt.weekday()
-            if day_in_week in (2, 3):
+            if day_in_week in {TU.weekday, WE.weekday}:
```
KJhellico, line 65:
```suggestion
            if day_in_week in {WE.weekday, TH.weekday}:
```

arkid15r: Included, thanks!

dr-prodigy:
Very worthy refactoring.. thank you ark! 👍

PR #804, opened 2022-11-23
A preview of `max-line-leng = 99`.

arkid15r:
Based on https://github.com/dr-prodigy/python-holidays/pull/802#discussion_r1029917750

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

See https://peps.python.org/pep-0008/#maximum-line-length


dr-prodigy:
@arkid15r + @mborsetti + everyone: while I'm not a huge fan of short code lines (in my personal projects I'm never so strict, and always prefer being comfortable with myself) I feel like python-holidays' case is quite different: not only 80 cols is one of the python's main / most accepted standards (PEP-8) which is why we adopted it, but also a large number of projects and contributors use it / work on it (which would make trickier to maintain a non-standard consistency overtime, eg: forcing people to change their favourite editor settings to avoid seeing 80 cols again) so we should always aim to usability / inclusivity for anyone.
For this reason, although I completely understand your point of view, I'm not in favour of incrementing the max column limit.

Some good food for thought here:
https://stackoverflow.com/questions/88942/why-does-pep-8-specify-a-maximum-line-length-of-79-characters

Anyway, I really appreciate you guys debating and always proposing ways to improve our library, so please keep on doing it! :)

PR #809, opened 2022-11-28
Uruguay: update holiday names

KJhellico:


holidays/countries/uruguay.py
```diff
@@ -36,60 +35,64 @@ def _populate(self, year):
...
-        self[easter(year)] = "Día de Pascuas [Easter Day]"
+        self[easter_date] = "Día de Pascuas [Easter Day]"
 
         holiday_pairs = (
             (
                 # Desembarco de los 33 Orientales en la playa de la Agraciada.
                 date(year, APR, 19),
-                "Desembarco de los 33 Orientales Landing of the 33 Orientals "
-                "Aterrissagem dos 33 Orientais Sbarco dei 33 orientali",
+                "Desembarco de los 33 Orientales "
```
arkid15r, line 89: Could this fit a single line?

arkid15r (commented):
Looks good!

holidays/countries/uruguay.py
```diff
@@ -36,60 +35,64 @@ def _populate(self, year):
...
 
         # Partially paid holidays:
 
-        # Día de Reyes - Feriado en el cual se conmemora la llegada de
-        # los reyes magos a Jesus.
-        self[date(year, JAN, 6)] = "Día de Reyes"
+        # Día de los Niños.
+        self[date(year, JAN, 6)] = "Día de los Niños [Children's Day]"
 
         # Natalicio de José Gervacio Artigas.
```
arkid15r, line 60: I believe it should be `Gervasio` instead of `Gervacio` here and below.

KJhellico: Yes, exactly. Thank you!

holidays/countries/uruguay.py
KJhellico: Yes, exactly fit :)

dr-prodigy:
👍

PR #719, opened 2022-09-17
Add i18n

bkthomps:
Add the ability to have i18n and add English and French translations for Canada. Once this base commit is merged, I'll add translations for other countries so that each country has translations for its official languages and English.

PR #810, opened 2022-11-29
Gettext translations

dragoon:
Hi,
I've re-done the i18n PR with the python's gettext module.
It's subjective, but I think that i18n module is easier to work with -- you have simple yml files which you can edit directly.
To work with gettext, you have to use command-line tools to generate binary .mo files.

I was also not able to rework 1 test that compares countries for equality since in my code I add translate method at runtime.
Anyway, let me know what you think.

Otherwise i18n functionality is desperately needed I think, currently some countries have their holidays names in local language, others in English (even when not local, e.g. some arabic countries).


arkid15r (commented):
Hi @dragoon,
this is really nice!

I was waiting for PR #719 comments to be addressed by the its author (or look into fixing it myself after the NY holidays). It seems this PR slightly changes the priority. I'll look into it after I'm done with current PH refactoring tasks.

The PR looks quite good at a glance. I may have some suggestions after a deeper look and playing with the new functionality.

Thank you!

dragoon:
Hi @arkid15r. Thanks for the quick look. Looking forward to further comments.

arkid15r:
Hi @dragoon, 
I've just filed #827 stacked on this one. 

Please take a look/review it when you have a chance.

Thank you!

PR #819, opened 2022-12-05
Paraguay holidays update

KJhellico:
- added special holidays
- fixed observance calculation

### DIFF
--- a/holidays/countries/paraguay.py
+++ b/holidays/countries/paraguay.py
@@ -13,9 +13,21 @@ from datetime import date
 
 from dateutil.easter import easter
 from dateutil.relativedelta import relativedelta as rd
-from dateutil.relativedelta import MO
 
-from holidays.constants import WED, WEEKEND, JAN, MAR, MAY, JUN, AUG, SEP, DEC
+from holidays.constants import (
+    WEEKEND,
+    JAN,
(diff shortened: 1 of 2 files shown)


holidays/countries/paraguay.py
```diff
@@ -27,82 +39,201 @@ class Paraguay(HolidayBase):
...
-            pass
-        elif date(year, MAR, 1).weekday() >= WED:
-            self[date(year, MAR, 1) + rd(weekday=MO(+1))] = name
+        dates_obs = {
+            2013: (MAR, 4),
+            2016: (FEB, 29),
+            2018: (FEB, 26),
+            2022: (FEB, 28),
+        }
+        if year in dates_obs:
```
arkid15r, line 169: Optional (here and below): 
```
self[date(year, *dates_obs.get(year, (MAR, 1)))] = "Día de los Héroes de la Patria [Patriots Day]"
```

arkid15r (commented):
LGTM

holidays/countries/paraguay.py
KJhellico: Thank you! That's why I like Python, for things like this :)

dr-prodigy:
👍 👍

PR #822, opened 2022-12-06
Angola: optimizations, observance calculation fix

KJhellico:


holidays/countries/angola.py
```diff
@@ -45,7 +45,7 @@ def _populate(self, year: int) -> None:
         # Observed since 1975
         # TODO do more research on history of Angolan holidays
 
-        if year < 1975:
+        if year <= 1974:
```
arkid15r, line 48: `super()._populate(year)` should go after 1974 check.

holidays/countries/angola.py
```diff
@@ -84,20 +82,17 @@ def _populate(self, year: int) -> None:
...
+            for k, v in list(self.items()):
+                if k.year == year:
+                    if year <= 2017:
+                        if k.weekday() == SUN:
+                            self[k + rd(days=+1)] = v + " (Observed)"
+                    else:
+                        if k.weekday() == TUE and k != date(year, JAN, 1):
+                            self[k + rd(days=-1)] = v + " (Day off)"
+                        elif k.weekday() == THU:
+                            self[k + rd(days=+1)] = v + " (Day off)"
```
arkid15r, line 95:
```suggestion
        if self.observed and year >= 1995:
            for k, v in list(self.items()):
                if k.year != year:
                  continue
                if year <= 2017:
                    if k.weekday() == SUN:
                        self[k + rd(days=+1)] = v + " (Observed)"
                else:
                    if k.weekday() == TUE and k != date(year, JAN, 1):
                        self[k + rd(days=-1)] = v + " (Day off)"
                    elif k.weekday() == THU:
                        self[k + rd(days=+1)] = v + " (Day off)"
```

arkid15r (commented):
🚀

PR #828, opened 2022-12-09
chile: Día Nacional de los Pueblos Indígenas

Nalguedo:
Fix the holiday day when year > 2021 according to the calculated solstice datetime adjusted to Chile timezone UTC-3 
Added necessary util Class _AstroMeeusAlgorithms

Issue: https://github.com/dr-prodigy/python-holidays/issues/666

### DIFF
--- a/holidays/countries/chile.py
+++ b/holidays/countries/chile.py
@@ -32,6 +32,7 @@ from holidays.constants import (
     DEC,
 )
 from holidays.holiday_base import HolidayBase
+from holidays.utils import _AstroMeeusAlgorithms
 
 
 class Chile(HolidayBase):
@@ -94,7 +95,13 @@ class Chile(HolidayBase):
         if year == 2021:
             self[date(year, JUN, 21)] = name
         if year > 2021:
-            self[date(year, JUN, 19)] = name
+            astro_alg = _AstroMeeusAlgorithms()
+            equinox = astro_alg.jd2date(astro_alg.summer(year))
+            # Received date for UTC timezone needs to be adjusted
+            # to match Chile's timezone
+            # https://www.feriadoschilenos.cl/#DiaNacionalDeLosPueblosIndigenasII
+            adjusted_date = equinox + rd(hours=-4)
+            self[date(year, JUN, adjusted_date.day)] = name
 
         # Saint Peter and Saint Paul (Law 18.432)
         name = "San Pedro y San Pablo [Saint Peter and Saint Paul]"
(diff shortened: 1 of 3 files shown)


holidays/countries/chile.py
```diff
@@ -94,7 +95,12 @@ def _populate(self, year):
         if year == 2021:
             self[date(year, JUN, 21)] = name
         if year > 2021:
-            self[date(year, JUN, 19)] = name
+            astro_alg = _AstroMeeusAlgorithms()
+            equinox = astro_alg.jd2date(astro_alg.summer(year))
+            # Received date for UTC timezone needs to be adjusted
+            # for Chile https://time.is/Chile UTC -3
+            adjusted_date = equinox + rd(hours=-3)
+            self[date(year, JUN, adjusted_date.day)] = name
```
KJhellico, line 103:
```
        if year >= 2021:
            dt = date(year, JUN, 21)
            if year >= 2022 and (year - 2022) % 4 > 1:
                dt = date(year, JUN, 20)
            self[dt] = name
```
Up to 2053 it will give the same result

Nalguedo: @KJhellico After reviewing your suggestion, I looked into other solstice/equinox-related holidays, such as Japanese Autumnal Equinox Day. I'm leaving it up to you and/or @dr-prodigy to decide. We can set it the way you suggested, or add the Meeus algorithms and refactor the existing holidays.
I'm fine with both ;)

KJhellico: [Here](https://www.timeanddate.com/holidays/chile/national-day-of-indigenous-peoples) and [here](https://www.feriadoschilenos.cl/#DiaNacionalDeLosPueblosIndigenasII) are another datelist, wich also refer to Meeus formulas and difference with yours in 2050 and 2079. And how do we know what algorithm government of Chile will use?

KJhellico: In process of Chile holidays research, I made my own PR #834 :)

Nalguedo: Great work with the research, @KJhellico xD ... I've realized that the solstice datetime for Chile is being calculated using UTC-4, I was using UTC-3 With this change, we get a full match with the list you've shared.

Nalguedo: If this PR is accepted I can update other holidays using the Meeus Algorithms

dr-prodigy:
I like this a lot, thank you @Nalguedo ! Just released in beta.. KUTGW! 👍

PR #829, opened 2022-12-09
Kazakhstan update

KJhellico:
- observed holidays
- holidays relevance period

### DIFF
--- a/holidays/countries/kazakhstan.py
+++ b/holidays/countries/kazakhstan.py
@@ -11,7 +11,9 @@
 
 from datetime import date
 
-from holidays.constants import JAN, MAR, MAY, JUL, AUG, DEC
+from dateutil.relativedelta import relativedelta as rd
+
+from holidays.constants import WEEKEND, JAN, MAR, MAY, JUL, AUG, DEC
 from holidays.holiday_base import HolidayBase
 from holidays.utils import _islamic_to_gre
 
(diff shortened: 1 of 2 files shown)


holidays/countries/kazakhstan.py
```diff
@@ -31,48 +33,64 @@ def _populate(self, year):
...
 
+        if self.observed:
+            for k, v in list(self.items()):
+                if k.weekday() in WEEKEND and k.year == year:
+                    next_workday = k + rd(days=+1)
+                    while next_workday.weekday() in WEEKEND or self.get(
+                        next_workday
+                    ):
+                        next_workday += rd(days=+1)
+                    self[next_workday] = v + " (Observed)"
```
KJhellico, line 87: Thess lines are not fully covered by test, although there is check with `self.holidays.observed = False`. @arkid15r, can you look into it? I tried new test functions :)

arkid15r: Try
```
self.assertHoliday(Kazakhstan(observed=True), *observed_holidays)
self.assertNoHoliday(Kazakhstan(observed=False), *observed_holidays)
```

KJhellico: It works. Thank you!

dr-prodigy:
I unfortunately had to revert this commit, due to a failing test on an observed holiday (2012-01-03).
Could you please review and provide a fix for that @KJhellico ?
Thx as usual! 👍

arkid15r:
The issue was caused by another [PR](https://github.com/dr-prodigy/python-holidays/blob/4008a7e3497d53fcb0dfa88a53998c1e3ff9dea9/test/common.py#L33). It was fixed in https://github.com/dr-prodigy/python-holidays/pull/844

Sorry for the inconvenience.

dr-prodigy:
Clear @arkid15r , thank you for explaining (honest, I could not dig so much on the issue yesterday)!
I've just merged #844 , will try to re-apply this one and see if all is right now!
Thx :+1:

PR #836, opened 2022-12-11
New Zealand: optimizations

KJhellico:


holidays/countries/new_zealand.py
```diff
@@ -78,9 +76,14 @@ class NewZealand(HolidayBase):
         "WTL",  # Correct code is WTC
     ]
 
-    def _populate(self, year):
-        super()._populate(year)
+    @staticmethod
```
arkid15r, line 79: It seems like a good candidate for DRY optimization (saw similar code in canada.py). No action required, just saying.

arkid15r (commented):
Looks good 👍

holidays/countries/new_zealand.py
KJhellico: Sorry, I didn't quite understand what it was about :(

arkid15r: We're going to have exactly the same code in 2 different places. Could be extracted and reused.

KJhellico: You mean `_get_nearest_monday`, I got it! :) Should I move it to utils.py?

PR #841, opened 2022-12-13
Introduce `HolidayBase::_is_monday` - `HolidayBase::_is_sunday` helper methods

arkid15r:
Migrate weekday checks to new logic.


dr-prodigy:
Hey @arkid15r first of all, thank you for this quite large work!

Although very well performed and accurate (as usual!), I don't feel like this refactoring provides so valuable improvements: if we had some centralized logic to add to those methods I could have seen the sense of incapsulating that code in new methods, but if I catch it right, your new methods `self._is_[day of week]` are providing some kind of alias for the usual syntax `k.weekday() == [day of week]` only, basically replicating code with no clear benefits on code simplification / readability.

Sorry for pointing this out after your complete work, but if there is no other reasons I can't currently see (explanation always welcome!), I would possibly avoid merging it.. :+1:

arkid15r:
This was my first step in order to introduce a level of abstraction and migrate from dealing with `date`, `relativedelta`, as well as MON-SUN, MO-SA where possible.

I hope this [PR](https://github.com/dr-prodigy/python-holidays/pull/865) completes the picture.

PR #846, opened 2022-12-14
Update gitignore.

arkid15r:
Exclude mypy/pytest cache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ MANIFEST
 *.pyc
 *.DS_Store
 coverage.xml
+.mypy_cache
+.pytest_cache
 .tox
 .venv
 docs/_build


dr-prodigy:
Hey there @arkid15r !
isn't this redundant?
Under `.mypy_cache` and `.pytest_cache` you should find 2 additional `.gitignore` files containing sth like
```
# Automatically created by mypy
*
```
so no cache file is in fact under VCS, apart from the 2 empty directories..

Does this make sense to you?

arkid15r:
You're right. Yesterday VSCode didn't recognize those directories as ignored for some reason. So I felt it needed to be fixed. It seems to be just fine without these new lines in the .gitignore. This PR can be discarded.

arkid15r:
Closing this out.

PR #839, opened 2022-12-13
Update south_africa.py

csloz:
Add 27th/Dec/2022 Presidential holiday for SA

PR #837, opened 2022-12-13
South Africa - 27 December 2022 declared public holiday

cjmeyer-polymorph:
https://www.gov.za/speeches/president-cyril-ramaphosa-declares-27-december-public-holiday-8-dec-2022-0000

PR #848, opened 2022-12-15
South Africa: added 27 Dec 2022 public holiday

KJhellico:
Alternative for #837 and #839, using `special_holidays`


dr-prodigy:
Thank you @KJhellico ! :+1:

PR #853, opened 2022-12-16
NYSE: fix Election Day holiday

KJhellico:
Election Day is Tuesday after first Monday in November, not first Tuesday


arkid15r (commented):
🎯

PR #857, opened 2022-12-17
United States update

KJhellico:
- Truman Day, Inauguration Day, Evacuation Day observation fixed
- optimizations

### DIFF
--- a/holidays/countries/united_states.py
+++ b/holidays/countries/united_states.py
@@ -88,19 +88,28 @@ class UnitedStates(HolidayBase):
         "WY",
     ]
 
+    def _add_with_observed(
+        self, dt: date, name: str, before: bool = True, after: bool = True
+    ) -> None:
+        self[dt] = name
+        if not self.observed:
+            return
+        if dt.weekday() == SAT and before:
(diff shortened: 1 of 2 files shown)


holidays/countries/united_states.py
```diff
@@ -88,19 +88,28 @@ class UnitedStates(HolidayBase):
...
         super()._populate(year)
 
         # New Year's Day
-        if year > 1870:
+        if year >= 1871:
             name = "New Year's Day"
-            self[date(year, JAN, 1)] = name
-            if self.observed and date(year, JAN, 1).weekday() == SUN:
-                self[date(year, JAN, 1) + rd(days=+1)] = name + " (Observed)"
+            self._add_with_observed(date(year, JAN, 1), name, False, True)
```
arkid15r, line 108: I believe it'd look more readable this way:
```suggestion
            self._add_with_observed(date(year, JAN, 1), name, before=False)
```

arkid15r (commented):
👍

holidays/countries/united_states.py
KJhellico: Done. Thank you!

PR #858, opened 2022-12-17
Malaysia multi-day Islamic holidays fixed

KJhellico:


holidays/countries/malaysia.py
```diff
@@ -325,24 +330,24 @@ def _populate(self, year):
...
-            hol_dates = [
-                (date(year, *date_obs), "") for date_obs in dates_obs[year]
-            ]
-        else:
-            hol_dates = [
-                (date_obs, estimated_suffix)
-                for date_obs in _islamic_to_gre(year, 12, 10)
-            ]
+        hol_dates = []
+        for yr in (year - 1, year):
```
arkid15r, line 334: Why would we need to look into the previous year here?

KJhellico: If Eid al-Adha falls on (year - 1, DEC, 31), second day holiday will be on (year, JAN, 1)

PR #861, opened 2022-12-20
Japan: Refactored japanese holidays

Nalguedo:
Refactored japanese holidays to make use of Meeus algorithms
Include pymeeus in requirements_dev.txt which was already a
transitive dependency associated with convertdate
Sort dependencies alphabetically in requirements_dev.txt


arkid15r:
Hey @Nalguedo, 
this PR https://github.com/dr-prodigy/python-holidays/pull/862 may help you with the failing tests.

Nalguedo:
> Hey @Nalguedo, this PR #862 may help you with the failing tests.

Shall I wait for this to be merged, or is there any way to force the CI to cherry-pick that commit?

Is there any plans to pin deps and transitive deps using pip-tools or poetry? ... I can help with that ;)

arkid15r:
> Shall I wait for this to be merged, or is there any way to force the CI to cherry-pick that commit?

I'd just update it locally as the merge into beta might take some time.

>Is there any plans to pin deps and transitive deps using pip-tools or poetry? ... I can help with that ;)

I'm not aware of such plans but the in general idea sounds appealing.

holidays/utils.py
```diff
@@ -949,7 +1000,7 @@ def jd2date(self, jd: float) -> datetime:
 
         h = int(f * 24)
         m = int((f - h / 24.0) * 1440)
-        s = round((f - h / 24.0 - m / 1440.0) * 86400.0)
+        s = round((f - h / 24.0 - m / 1440.0) * 86400.0, 2)
```
KJhellico, line 1003: Why round to 2 digits then make integer instead of one step? `round(n, 0)` return `int`

Nalguedo: I was getting values above `59sec` for some of the tests like some 1999 dates ... You can check with the following JD `2451258.574303147` 
Another thing is, we should keep this function as close as possible to the main implementation, it's easier to check other language implementations and compare it with our Python version.
Anyway, if you have any suggestion that fixes this without breaking the rest of the tests, I'll be happy to include it ;)

KJhellico: Yes, I checked and I agree :)
But I have more general question about using Meeus. We already have `pymeeus` in dependencies (from `convertdate`). Maybe use needed functions from it and don't add more code will be better? @dr-prodigy, @arkid15r, what do you think about it?

Nalguedo: I was checking the dependencies for version pinning proposal and I've found that too ... pymeeus will do the trick and it's already a dependency (from convertdate as mentioned by @KJhellico )
So maybe it's better to remove the code related to Meeus Algorithms :'( ... I'll close this PR in favour of a new PR that I will open using pymeeus ... if this is OK for everyone ;)

dr-prodigy:
> > Hey @Nalguedo, this PR #862 may help you with the failing tests.
> 
> Shall I wait for this to be merged, or is there any way to force the CI to cherry-pick that commit?
> 
> Is there any plans to pin deps and transitive deps using pip-tools or poetry? ... I can help with that ;)

Hi @Nalguedo thank you for your work! Given the blocking issue, I've merged @arkid15r 's PR already.

PS: while, philosophically speaking, we always try to avoid dep pinning, we could consider your interesting proposal .. we'll discuss it as soon as possible :+1:

Nalguedo:
If this change is OK, I'll remove Meeus algorithms from Utils and refactor Chile holiday using the same approach
@dr-prodigy @arkid15r @KJhellico

arkid15r:
> If this change is OK, I'll remove Meeus algorithms from Utils and refactor Chile holiday using the same approach @dr-prodigy @arkid15r @KJhellico

I'm okay with the general approach. Didn't look into the implementations though.

KJhellico:
Looks good 👍

arkid15r (commented):
🇯🇵

holidays/countries/japan.py
```diff
@@ -60,7 +62,17 @@ def _populate(self, year):
             self[date(year, FEB, 23)] = "天皇誕生日"
 
         # Vernal Equinox Day
-        self[self._vernal_equinox_day(year)] = "春分の日"
+        epoch = Sun.get_equinox_solstice(year, target="spring")
+        equinox = Epoch(epoch).get_full_date()
```
arkid15r, line 66: I recommend using "Asia/Tokyo" tz adjustment here mostly for consistency reasons (see Chile example) as DST is not observed in Japan.

Nalguedo: Updated as suggested ;) (much cleaner now ... Thanks)

dr-prodigy:
Thank you very much ! :+1:

PR #863, opened 2022-12-21
Chile: refactor equinox holiday to use pymeeus

Nalguedo:
pymeeus being a dependency already enables the equinox and solstice date calculation
Remove Meeus Algorithms code from utils and refactor Chile holidays to use pymeeus to get equinox dates


holidays/countries/chile.py
```diff
@@ -105,12 +106,19 @@ def _populate(self, year):
         if year == 2021:
             self[date(year, JUN, 21)] = name
         elif year >= 2022:
-            astro_alg = _AstroMeeusAlgorithms()
-            equinox = astro_alg.jd2date(astro_alg.summer(year))
+            epoch = Sun.get_equinox_solstice(year, target="summer")
+            equinox = Epoch(epoch).get_full_date()
```
arkid15r, line 110:
```suggestion
            equinox = map(int, Epoch(epoch).get_full_date())
```

holidays/countries/chile.py
```diff
@@ -105,12 +106,19 @@ def _populate(self, year):
...
             # https://www.feriadoschilenos.cl/#DiaNacionalDeLosPueblosIndigenasII
-            adjusted_date = equinox + rd(hours=-4)
+            adjusted_date = datetime(
+                equinox[0],
+                equinox[1],
+                equinox[2],
+                equinox[3],
+                equinox[4],
+                int(equinox[5]),
+            ) + rd(hours=-4)
```
arkid15r, line 121:
```suggestion
            adjusted_date = datetime(
                *equinox, tzinfo=tz.UTC
            ).astimezone(tz.gettz("America/Santiago"))
```

holidays/countries/chile.py
```diff
@@ -9,16 +9,17 @@
 #  Website: https://github.com/dr-prodigy/python-holidays
 #  License: MIT (see LICENSE file)
 
-from datetime import date
+from datetime import date, datetime
```
arkid15r, line 12:
```suggestion
from datetime import date, datetime

from dateutil import tz
```

arkid15r (commented):
A couple of suggestions for your consideration:

holidays/countries/chile.py
Nalguedo: Fixed as suggested ;)
Great improvement ... Thanks

holidays/countries/chile.py
Nalguedo: Fixed as suggested ;)

dr-prodigy:
This is a very valuable review, @Nalguedo , thx! I've just merged it.
Although it is an implicit dependency, in order to make this more robust, I will also explicitly include PyMeeus in the prerequisite libraries in a further commit.
Thank you :+1:

PR #866, opened 2022-12-22
Bahrain holidays

taljishi:
Adding support for Bahrain holidays bahrain.py.

This is my first contribution, and since I'm not a programmer or coder, I based some of the code with Saudi Arabia and UAE to get a head start

### DIFF
--- a/holidays/countries/bahrain.py
+++ b/holidays/countries/bahrain.py
@@ -137,7 +137,7 @@ class Bahrain(HolidayBase):
                 self[hol_date] = f"{new_hijri_year}* (*estimated)"
 
         # Ashoora
-        # Date is announced each year. Usually the 9th and 10th Day,
+        # Date is announced each year, for the 9th and 10th Day,
         # of the month of Muharam
         ashoora = "Ashoora"
         if year in dates_obs:
--- a/holidays/countries/bahrain.py
+++ b/holidays/countries/bahrain.py
@@ -1,0 +1,177 @@
+#  python-holidays
+#  ---------------
+#  A fast, efficient Python library for generating country, province and state
+#  specific sets of holidays on the fly. It aims to make determining whether a
+#  specific date is a holiday as fast and flexible as possible.
+#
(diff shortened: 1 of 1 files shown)


holidays/countries/bahrain.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Ashoora
+        # Date is announced each year, for the 9th and 10th Day,
+        # of the month of Muharam
+        ashoora = "Ashoora"
+        if year in dates_obs:
+            for date_obs in dates_obs[year]:
+                hol_date = date(year, *date_obs)
+                self[hol_date + rd(days=9)] = ashoora
+                self[hol_date + rd(days=10)] = f"{ashoora} Holiday"
```
KJhellico, line 147: If Ashoora calculated as 9 Muharram, it's no need to add 9
```suggestion
                self[hol_date] = ashoora
                self[hol_date + rd(days=+1)] = f"{ashoora} Holiday"
```

holidays/countries/bahrain.py
```diff
@@ -0,0 +1,177 @@
...
+        # National Day
+        self[date(year, DEC, 16)] = "National Day"
+        self[date(year, DEC, 17)] = "National Day"
+
+        # Eid Al Fitr
+        # Date is announced each year. Usually stretches along 3 or 4 days,
+        # in some instances prepending/appending a day or two
+        # before/after the official holiday.
+        fitr = "Eid Al Fitr"
+        if year in dates_obs:
```
KJhellico, line 81: In that case, you should define some known dates list as `dates_obs`

arkid15r:
@KJhellico, yeah, this PR's code is somewhat incomplete but that's fine. It's a draft so I believe no review was required. I'm going to take over @taljishi's work from here as we agreed in https://github.com/dr-prodigy/python-holidays/discussions/813

holidays/countries/bahrain.py
arkid15r: I'll try to find out the correct values for date objects soon.

PR #867, opened 2022-12-22
Kazakhstan: update holidays relevance period and names

KJhellico:


mihalken:
Hello. I'm already working on it. :) https://github.com/dr-prodigy/python-holidays/pull/864
I have suggestions...

holidays/countries/kazakhstan.py
```diff
@@ -13,81 +13,92 @@
...
-
-        # Unity Day
-        self[date(year, MAY, 1)] = "Unity Day"
-
-        # Defender's Day
+        name = "New Year"
+        self[date(year, JAN, 1)] = name
+        self[date(year, JAN, 2)] = name
+
+        # International Women’s Day
```
arkid15r, line 39: Not sure whether ' (vs ’) is more standard option.

arkid15r (commented):
LGTM

holidays/countries/kazakhstan.py
KJhellico: Thank you. Replaced with '

PR #879, opened 2022-12-29
Japan: substitute holidays and citizens' holidays calculation

KJhellico:


KJhellico:
I'm not sure, but maybe "substitute holidays" is `observed`? Or it's just holidays of special type?

holidays/countries/japan.py
```diff
@@ -143,166 +143,22 @@ def _populate(self, year):
...
-        )
-        for holiday in table:
-            month = holiday[0]
-            day = holiday[1]
-            years = holiday[2]
-            if year in years:
-                self[date(year, month, day)] = "振替休日"
+        # When a national holiday falls on Sunday, next working day
+        # shall become a public holiday (振替休日) - substitute holidays
+        for dt in sorted(list(self.keys())):
```
arkid15r, line 148: I'm not sure if you need `sorted()` here and below. But if you do, the `for dt in sorted(self.keys()):` would work just fine.

arkid15r (commented):
>I'm not sure, but maybe "substitute holidays" is observed? Or it's just holidays of special type?

My understanding is that "substitute holiday" is a direct translation of "振替休日". And the meaning of that seems to be equal to "observed holiday" term.

🇯🇵  LGTM!

holidays/countries/japan.py
KJhellico: Yes, `sorted()` isn't needed here. Thank you!

KJhellico:
> My understanding is that "substitute holiday" is a direct translation of "振替休日". And the meaning of that seems to be equal to "observed holiday" term.

I think so too. So make it `self.observed` depended now.

arkid15r:
Great refactoring 👍, thank you!

PR #880, opened 2022-12-29
Singapore holidays update

KJhellico:
- observed holidays calculation fixed
- some holidays dates fixed

### DIFF
--- a/holidays/countries/singapore.py
+++ b/holidays/countries/singapore.py
@@ -84,6 +84,10 @@ class Singapore(HolidayBase):
         super().__init__(years, expand, observed, subdiv, prov, state)
 
     def _populate(self, year) -> None:
+        def _add_holiday(dt: date, hol: str) -> None:
+            if dt.year == year:
+                self[dt] = hol
+
         super()._populate(year)
 
         # New Year's Day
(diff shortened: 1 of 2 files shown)


holidays/countries/singapore.py
```diff
@@ -84,6 +84,10 @@ def __init__(
         super().__init__(years, expand, observed, subdiv, prov, state)
 
     def _populate(self, year) -> None:
+        def _add_holiday(dt: date, hol: str) -> None:
```
arkid15r, line 87: The name could be just `add_holiday(dt: date, hol: str) -> None:` as that's an inner function with no name conflicts.

arkid15r (commented):
🇸🇬 LGTM!

holidays/countries/singapore.py
KJhellico: Many countries with islamic holidays have this function exactly as `_add_holiday`, so I think it's better to be uniformly for now.

dr-prodigy: Agree w @KJhellico , the underscore suggests it's a private member 👍

arkid15r:
Great work on code and test improvements!

PR #886, opened 2023-01-03
Introduce `HolidayBase::_is_leap_year()`

arkid15r:
Refactor ET, HK.


mborsetti:
@arkid15r You're adding amazing code to this library, and it's great, but can I ask a question: What is the reasoning behind this. At first look, all this extra code simply adds layers of complexity to calling the built-in Python method [calendar.isleap](https://docs.python.org/3/library/calendar.html#calendar.isleap) directly.  If so, I have a hard time reconciling it with many of the [Python principles](https://en.wikipedia.org/wiki/Zen_of_Python), such as "Simple is better than complex", "Flat is better than nested", "Explicit is better than implicit". I'm probably just inexperienced, so apologies in advance!

arkid15r:
>a question about this specific PR's motivation? 

Hi Mike, thanks for the question! This PR continues my efforts (https://github.com/dr-prodigy/python-holidays/pull/769/, https://github.com/dr-prodigy/python-holidays/pull/841/, https://github.com/dr-prodigy/python-holidays/pull/865) to introduce an abstraction layer in order to simplify country holidays implementation. 

Here it's more about general OOP principles rather than Python ones. In other words the layer of complexity you see here is a part of an abstraction layer which allows to migrate from the lower level operations (like creating/using date objects, importing days of week, WEEKEND constants, etc), make the code better structured, readable, and testable. 

In this particular case one wouldn't need to care about leap year logic implementation and importing necessary modules, but still be able to override the method for a special cases if needed (e.g. Ethiopia). In a general case you can think of these changes as of lego blocks for better construction experience. 

I hope this answers your question.

mborsetti:
@arkid15r Thanks for taking the time to explain this to me -- and for all the wonderful work!

PR #892, opened 2023-01-12
Clean up timedelta/relativedelta usage

arkid15r:
- Standardize `relativedelta` days offset usage
  - Replace `timedelta` with `relativedelta`

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Check out repo
         uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v4.4.0
+        uses: actions/setup-python@v4.5.0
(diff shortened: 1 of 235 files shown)


dr-prodigy:
Hi @arkid15r , I like the idea behind this PR a lot! although I have an important note to add: given this refactoring is very large, I would prefer to have it split in 2, ie:
1/ business logic code refactoring, no changes on tests
2/ no changes on BL, tests refactoring
so to guarantee no regressions without requiring extensive visual reviews.
Do you think it's still possible?
Thx a lot!

arkid15r:
That's a great point, @dr-prodigy!

I implemented and tested these changes in the 2 steps you've just described. However, the PR doesn't reflect that.
It makes perfect sense to split it into 2 as you suggested. I've extracted tests into https://github.com/dr-prodigy/python-holidays/pull/894

Thanks for the review!

PR #909, opened 2023-01-26
Bosnia and Herzegovina holidays update

KJhellico:


KJhellico:
Tests needs to be rewritten with `assertHolidaysName`, now I see it. It will be done ASAP.

test/countries/test_bosnia_and_herzegovina.py
```diff
@@ -10,285 +10,281 @@
...
-class TestBA(unittest.TestCase):
+class TestBA(TestCase):
     def setUp(self):
-        self.holidays = holidays.BA(observed=False)
-        self.holidays_observed = holidays.BA(observed=True)
-
-    def test_new_years(self):
-        for dt in (date(2010, 12, 31), date(2017, 1, 3)):
-            self.assertNotIn(dt, self.holidays)
+        self.holidays = BosniaAndHerzegovina(observed=False)
```
arkid15r, line 20: Not sure why we're testing non-default (observed=False) behaviour here.

arkid15r (commented):
LGTM

test/countries/test_bosnia_and_herzegovina.py
KJhellico: Before it was like that (`self.holidays = holidays.BA(observed=False) …`), so I left it. :)

arkid15r:
Merged! Thanks for reworking this, @KJhellico!

PR #914, opened 2023-01-29
Fix Islamic holidays calculation

KJhellico:
- fix for holidays that happen twice in a Gregorian year (primarily islamic New Year)
- years range checking more strictly

### DIFF
--- a/holidays/utils.py
+++ b/holidays/utils.py
@@ -25,7 +25,7 @@ from functools import lru_cache
 from typing import Dict, Iterable, List, Optional, Union
 
 from hijri_converter import convert
-from hijri_converter.ummalqura import GREGORIAN_RANGE
+from hijri_converter.ummalqura import GREGORIAN_RANGE, HIJRI_RANGE
 
 from holidays import countries, financial
 from holidays.holiday_base import HolidayBase
@@ -318,17 +318,18 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
     """
 
     # To avoid hijri_converter check range OverflowError.
-    dt = (g_year, h_month, h_day)
-    dt_min, dt_max = GREGORIAN_RANGE
-    if dt < dt_min or dt > dt_max:
+    g_year_min, g_year_max = (d[0] for d in GREGORIAN_RANGE)
+    h_year_min, h_year_max = (d[0] for d in HIJRI_RANGE)
+    if g_year <= g_year_min or g_year > g_year_max:
         return ()
 
     h_year = convert.Gregorian(g_year, 1, 1).to_hijri().year
-    gre_dates = (
-        convert.Hijri(year, h_month, h_day).to_gregorian()
-        for year in range(h_year - 1, h_year + 2)
+    h_years = (
+        y for y in range(h_year, h_year + 3) if h_year_min <= y <= h_year_max
     )
-
+    gre_dates = (
+        convert.Hijri(y, h_month, h_day).to_gregorian() for y in h_years
+    )
     return (gre_date for gre_date in gre_dates if gre_date.year == g_year)
 
 
--- a/holidays/utils.py
+++ b/holidays/utils.py
@@ -301,7 +301,7 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
     calendar is about 11 days shorter.
 
     Relies on package `hijri_converter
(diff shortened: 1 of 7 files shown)


holidays/utils.py
```diff
@@ -318,17 +318,18 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
...
-    dt_min, dt_max = GREGORIAN_RANGE
-    if dt < dt_min or dt > dt_max:
+    min_year, max_year = (d[0] for d in GREGORIAN_RANGE)
+    if g_year <= min_year or g_year > max_year:
         return ()
 
     h_year = convert.Gregorian(g_year, 1, 1).to_hijri().year
+    min_h_year, max_h_year = (d[0] for d in HIJRI_RANGE)
+    h_years = (
+        y for y in range(h_year, h_year + 3) if min_h_year <= y <= max_h_year
```
arkid15r, line 328: This one is interesting. I've had a goal to understand the lunar calendars better for a long time. And I still didn't get to a comfortable point with that.

My understanding is that we have a previous (h_year -1) and next (h_year + 2) hijri year based holidays covered at the moment. And you are replacing it with just two next year.

My question is how come it's safe to remove it (the passing tests confirm it's safe to do so or we just don't have enough cases covered)? Perhaps, a brief explanation of calculation logic would be great to add as a comment. 

I'd also like @mborsetti to look into this change as an original author (thank you!).

arkid15r (commented):
Hey @KJhellico, that's a great improvement for Islamic holidays!

Could you take a look into the comments below?
Thank you!

holidays/utils.py
mborsetti: > I'd also like @mborsetti to look into this change as an original author (thank you!).

@arkid15r and @KJhellico, will do over the weekend.

KJhellico: > My question is how come it's safe to remove it (the passing tests confirm it's safe to do so or we just don't have enough cases covered)? Perhaps, a brief explanation of calculation logic would be great to add as a comment.

**h_year** begin before or at same day with **g_year** → **h_year-1** end strictly before **g_year** begin → there can't be days from **h_year-1** in  **g_year** → we don't need to check **h_year-1**.

**h_year+2** we need to check for holidays which falls on Hijri year begin (eg. Hijri New Year).

So, I check `range(h_year, h_year + 3)`.

holidays/utils.py
```diff
@@ -318,17 +318,18 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
...
+    )
     gre_dates = (
-        convert.Hijri(year, h_month, h_day).to_gregorian()
-        for year in range(h_year - 1, h_year + 2)
+        convert.Hijri(y, h_month, h_day).to_gregorian() for y in h_years
```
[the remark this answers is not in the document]
KJhellico: I tried to replace `convert.Hijri(y, h_month, h_day).to_gregorian()` with `date(*convert.Hijri(y, h_month, h_day).to_gregorian().datetuple())` and got holiday_base.py:312 not covered by tests. How can I solve this?

arkid15r (commented):
This is a great Islamic holidays accuracy improvement.
LGTM, so I'll merge it later this week if there is no objections from other contributors.

Nicely done!

mborsetti:
Apologies for the delay.

@KJhellico awesome job!  

A fabulous 69% improvement in speed.

You also caught a logic bug in the function I wrote which incorrectly returned only one new Islamic new year date for years 1943, 1976, 2008, 2041, and 2073 instead of the two correct ones.

Masterful, thanks!

mborsetti (changes requested):
Typo (mine from original docs). Suggest

`<https://www.pypi.org/project/hijri_converter>`__.`

mborsetti (changes requested):
Suggest the following docstring change:

```
        :return:
            An Iterable of Gregorian dates within the Gregorian year specified
            that matches the Islamic (Lunar Hijrī) calendar day and month
            specified. An empty Iterable is returned if the Gregorian year
            is outside of the covered period, which as of hijri_converter 2.2.4
            (in January 2023) is Gregorian years 1925 to 2076 inclusive, or
            equal to the contents of hijri_converter.ummalqura.GREGORIAN_RANGE
            plus/minus 1 year.
```

arkid15r:
This is a great fix! Thank you @KJhellico!

@mborsetti thanks for the review and suggestions!

PR #924, opened 2023-02-02
Refactor days offset calculation for recently added countries

KJhellico:


arkid15r (commented):
Thanks for making it consistent again!

Could you include Albania and Dominican Republic too?

KJhellico:
> Could you include Albania and Dominican Republic too?
Done. I missed them, thanks!

arkid15r:
Great stuff!

PR #933, opened 2023-02-08
Optimize Burundi observed holidays

KJhellico:
Performance tests:
```
def calc_hol1():
    for year in range(1962, 2100):
        hols = Burundi(years=year)

def calc_hol2():
    hols = Burundi(years=range(1962, 2100))

t1 = timeit.timeit(calc_hol1, number=100)
t2 = timeit.timeit(calc_hol2, number=100)
```
| Test | Current beta |   PR  |
|:----:|:------------:|:-----:|
|  t1  |        3.114 | 3.117 |
|  t2  |        6.797 | 2.920 |

### DIFF
--- a/holidays/countries/burundi.py
+++ b/holidays/countries/burundi.py
@@ -35,67 +35,66 @@ class Burundi(HolidayBase):
     country = "BI"
 
     def _populate(self, year):
+        def _add_with_observed(hol_date: date, hol_name: str) -> None:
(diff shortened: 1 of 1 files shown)


arkid15r (commented):
LGTM

// You can skip the performance tests part as we both [agreed on these changes earlier](https://github.com/dr-prodigy/python-holidays/pull/925#discussion_r1098379759)

PR #939, opened 2023-02-12
Add Northern Mariana Islands (ISO-3666-2 code US-MP)

mborsetti:
Add Northern Mariana Islands (ISO-3666-2 code US-MP)


mborsetti:
merged into #941 (IDE issues)

PR #961, opened 2023-02-17
Optimize Azerbaijan observed holidays

KJhellico:
Maybe it's not the best implementation, but I tried. :) @arkid15r, look, please.

### DIFF
--- a/holidays/countries/azerbaijan.py
+++ b/holidays/countries/azerbaijan.py
@@ -11,6 +11,7 @@
 
 from datetime import date
 from datetime import timedelta as td
+from typing import Dict, List
 
 from holidays.constants import JAN, MAR, APR, MAY, JUN, JUL, AUG, SEP, OCT
 from holidays.constants import NOV, DEC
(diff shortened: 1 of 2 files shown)


holidays/countries/azerbaijan.py
```diff
@@ -26,54 +27,52 @@ class Azerbaijan(HolidayBase):
...
-            while next_workday.year == year and (
-                self._is_weekend(next_workday) or self.get(next_workday)
-            ):
-                next_workday += td(days=+1)
-            _add_holiday(next_workday, f"{hol_name} (Observed)")
-
-        def _add_holiday(hol_date: date, hol_name: str) -> None:
-            if hol_date.year == year:
-                self[hol_date] = hol_name
+        def add_to_list(hol_date: date, hol_name: str) -> None:
```
arkid15r, line 30: I believe this will be refactored to a class level method later (as a general shift to a cleaner OOP approach) combined with `_add_holiday()`.  As for now just:
```suggestion
        def collect_holiday(hol_date: date, hol_name: str) -> None:
```

holidays/countries/azerbaijan.py
```diff
@@ -26,54 +27,52 @@ class Azerbaijan(HolidayBase):
...
+                hol_list[hol_date].append(hol_name)
+            else:
+                hol_list[hol_date] = [hol_name]
 
         if year <= 1989:
             return None
 
         super()._populate(year)
 
+        hol_list: Dict[date, List[str]] = {}
```
arkid15r, line 41: A bit misleading naming (list vs dict). I prefer using something like `holiday_date_names_mapping`.

arkid15r (commented):
>Maybe it's not the best implementation, but I tried. :) @arkid15r, look, please.

We all do our best here. Let's use what you've come up with for now! 👍

KJhellico:
Naming is always been my weak point :)

holidays/countries/azerbaijan.py
KJhellico: Done.

holidays/countries/azerbaijan.py
KJhellico: Done. Thank you!

arkid15r (commented):
👍  LGTM

PR #969, opened 2023-02-20
Refactor calendars: extract to a separate module

arkid15r:


KJhellico:
Maybe we should rename `_*Calendar` classes and `_islamic_to_gre` function without underscores? They are actually external, not local.

KJhellico:
> Not sure why you think they're external. They actually aren't as we don't want to expose them as an "officially supported" part of the package according to current convention.

Yes, really, they are only used inside the package. I didn't think about it.

PR #977, opened 2023-02-22
Update Mexico holidays

jahirfiquitiva:
According to this official document from the Mexico government (https://www.gob.mx/profedet/articulos/sabes-cuales-son-los-dias-de-descanso-obligatorios-163134?idiom=es) (in Spanish) New Year's and Independence Day are not re-calculated even if they happen on a weekend day

From the list in the doc, holidays should be:
1. January 1st
2. February 5th, or the first Monday in February
3. March 21st, or the third Monday in March
4. May 1st
5. September 16th
6. November 20th, or the third Monday in November
7. Every 6 years, December 1st, corresponding to `Change of Federal Government`
8. December 25th

### DIFF
--- a/holidays/countries/mexico.py
+++ b/holidays/countries/mexico.py
@@ -34,14 +34,7 @@ class Mexico(HolidayBase):
 
         # New Year's Day
         name = "Año Nuevo [New Year's Day]"
-        dt = date(year, JAN, 1)
-        self[dt] = name
-        if self.observed and self._is_sunday(dt):
-            self[dt + td(days=+1)] = name + " (Observed)"
-        # The next year's observed New Year's Day can be in this year
-        # when it falls on a Friday (Jan 1st is a Saturday)
-        if self.observed and self._is_friday(year, DEC, 31):
-            self[date(year, DEC, 31)] = name + " (Observed)"
+        self[date(year, JAN, 1)] = name
 
         # Constitution Day
         name = "Día de la Constitución [Constitution Day]"
@@ -71,8 +64,7 @@ class Mexico(HolidayBase):
 
         # Independence Day
         name = "Día de la Independencia [Independence Day]"
-        dt = date(year, SEP, 16)
-        self._add_with_observed(dt, name)
+        self[date(year, SEP, 16)] = name
 
         # Revolution Day
         name = "Día de la Revolución [Revolution Day]"


KJhellico:
It's always been like that? Or New Year's Eve and Independence Day were movable also until a certain year?

KJhellico:
OK, we'll assume that it's always been that way. But I want to better understand the law. :) Please explain, if you can.

1. **Fixed holidays (Jan 1, May 1, Sep 16, Dec 25)**: if this day fall on Sunday, there is no additional holidays - just this one day, and that's all? No non-working day in Monday?
2. **Movable holidays (Feb 5, Mar 21, Nov 20, Dec 1)**: if this day not fall on Monday, is it not a holiday at all? So if it's Wednesday, for example, is it just a working day? And the holiday will be next Monday?

jahirfiquitiva:
@KJhellico 

1. That's correct. There's no other holidays, just that day.
2. They will move to the right day as mentioned in the PR description, but the exact day would be a normal day. Except Dec 1, which would be a fixed holiday too (just happening every 6 years only)

KJhellico:
> 1. That's correct. There's no other holidays, just that day.

[Wiki](https://en.wikipedia.org/wiki/Public_holidays_in_Mexico) says: "When a statutory holiday falls on a Sunday, Monday is considered a statutory holiday; if a statutory holiday falls on Saturday, Friday will be considered a statutory holiday."
So it's completely wrong? Then we need to change the entire code for Mexico...

> 2. They will move to the right day as mentioned in the PR description, but the exact day would be a normal day. Except Dec 1, which would be a fixed holiday too (just happening every 6 years only)

Yes, right, Dec 1 is fixed too, I've made mistake. Thanks!

jahirfiquitiva:
Hmm, I would say Wiki is incorrect, given the information is taken from a government page. Anyway, that might be true for those holidays that are movable, I guess

KJhellico:
> that might be true for those holidays that are movable, I guess

No, movable holidays always move to Monday

jahirfiquitiva:
@KJhellico right, well, given the official doc by the mx government, still I would say wiki is wrong.

KJhellico:
After studying Law of Labor and its revisons, I came to the conclusion that Mexico does not have "observed" holidays and I made PR #980. @jahirfiquitiva, please look at it.

jahirfiquitiva:
it looks good, thanks @KJhellico

arkid15r:
@jahirfiquitiva, Can I close this PR in favor of https://github.com/dr-prodigy/python-holidays/pull/980 then?

PR #979, opened 2023-02-22
Refactor common.TestCase

arkid15r:
Refactor common.TestCase:
  - Add context specific assertions for non-observed holidays.
  - Migrate existing tests.


KJhellico (approved):
LGTM

PR #989, opened 2023-02-28
Add functions for relative weekdays calculation

KJhellico:
## Proposed change

Optimize relative weekdays calculation. Most common cases of `dateutil.relativedelta` use (n-th weekday of month, n-th weekday from date) replaced with own simple functions. To start, I chose 2 countries that have many such holidays - US and Canada.

Perhaps functions names need to be corrected. :)

## Type of change

## Checklist

### DIFF
--- a/holidays/calendars.py
+++ b/holidays/calendars.py
@@ -63,6 +63,41 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
     return (gre_date for gre_date in gre_dates if gre_date.year == g_year)
 
 
+def _get_nth_weekday_from(n: int, weekday: int, from_dt: date) -> date:
+    """
+    Return date of a n-th weekday after (n is positive)
+    or before (n is negative) a specific date
+    (e.g. 1st Monday, 2nd Saturday, etc).
+    """
+
+    if n > 0:
+        delta = (n - 1) * 7 + (weekday - from_dt.weekday()) % 7
+    else:
+        delta = (n + 1) * 7 - (from_dt.weekday() - weekday) % 7
+    return from_dt + td(days=delta)
+
+
+def _get_nth_weekday_of_month(
+    n: int, weekday: int, month: int, year: int
+) -> date:
+    """
+    Return date of n-th weekday of month for a specific year
+    (e.g. 1st Monday of Apr, 2nd Friday of June, etc).
+    If n is negative the countdown starts at the end of month
+    (i.e. -1 is last).
+    """
+
+    if n < 0:
+        month += 1
+        if month > 12:
+            month = 1
+            year += 1
+        start_date = date(year, month, 1) + td(days=-1)
+    else:
+        start_date = date(year, month, 1)
+    return _get_nth_weekday_from(n, weekday, start_date)
+
+
 class _ChineseLuniSolar:
     def __init__(self) -> None:
         """
--- a/tests/test_calendars.py
+++ b/tests/test_calendars.py
@@ -13,6 +13,7 @@ import unittest
 from datetime import date
 
 from holidays import calendars
+from holidays.calendars import _get_nth_weekday_of_month
 from holidays.constants import FEB, MAR, MAY, JUN, JUL, AUG, OCT
 
 
@@ -110,3 +111,18 @@ class TestThaiLuniSolarCalendar(unittest.TestCase):
                 visakha_bucha_year_date[year],
                 self.calendar.visakha_bucha_date(year),
             )
+
+
+class TestRelativeWeekdays(unittest.TestCase):
+    def test_weekday_of_month(self):
+        # 1st Monday of 2023 months
+        for month, day in enumerate((3, 7, 7, 4, 2, 6, 4, 1, 5, 3, 7, 5), 1):
+            first_monday = _get_nth_weekday_of_month(1, 1, month, 2023)
+            self.assertEqual(first_monday.day, day)
+
+        # Last Saturday of 2023 months
+        for month, day in enumerate(
+            (28, 25, 25, 29, 27, 24, 29, 26, 30, 28, 25, 30), 1
+        ):
+            last_friday = _get_nth_weekday_of_month(-1, 5, month, 2023)
+            self.assertEqual(last_friday.day, day)
--- a/holidays/countries/canada.py
+++ b/holidays/countries/canada.py
@@ -13,11 +13,10 @@ from datetime import date
 from datetime import timedelta as td
 
 from dateutil.easter import easter
-from dateutil.relativedelta import MO, SU
-from dateutil.relativedelta import relativedelta as rd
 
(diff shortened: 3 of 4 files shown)


holidays/calendars.py
```diff
@@ -63,6 +63,37 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
     return (gre_date for gre_date in gre_dates if gre_date.year == g_year)
 
 
+def _n_weekday_from_date(d: date, weekday: int, n: int = 1) -> date:
```
arkid15r, line 66: I changed the args order too to better reflect the name:
```suggestion
def _get_nth_weekday_since(n: int,  weekday: int, since: date) -> date:
```

holidays/calendars.py
```diff
@@ -63,6 +63,37 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
...
+    """
+    Return n-th weekday from date (e.g. 1st Monday, 2nd Saturday).
```
arkid15r, line 68:
```suggestion
    Return date of a n-th weekday after a specific date (e.g. 1st Monday, 2nd Saturday, etc).
```

holidays/calendars.py
```diff
@@ -63,6 +63,37 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
...
+    """
+
+    if n > 0:
+        jumpdays = (n - 1) * 7 + (weekday - d.weekday()) % 7
```
arkid15r, line 72: I'd use `delta_days` or just `days`

holidays/calendars.py
```diff
@@ -63,6 +63,37 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
...
+    else:
+        jumpdays = (n + 1) * 7 - (d.weekday() - weekday) % 7
+    return d + td(days=jumpdays)
+
+
+def _n_weekday_of_month(
+    year: int, month: int, weekday: int, n: int = 1
+) -> date:
```
arkid15r, line 80:
```suggestion
def _get_nth_weekday_of_month(
    n: int, weekday: int, month: int, year: int
) -> date:
```

holidays/calendars.py
```diff
@@ -63,6 +63,37 @@ def _islamic_to_gre(g_year: int, h_month: int, h_day: int) -> Iterable[date]:
...
+    """
+    Return n-th weekday of month (e.g. 1st Monday of Apr, 2nd Friday of June).
+    If n < 0, the countdown begins at the end of month (i.e. -1 is last etc.)
```
arkid15r, line 83:
```suggestion
    Return date of n-th weekday of month for a specific year (e.g. 1st Monday of Apr, 2nd Friday of June, etc).
    If n is negative the countdown starts at the end of month (i.e. -1 is last)
```

tests/test_calendars.py
```diff
@@ -110,3 +111,48 @@ def test_visakha_bucha_date(self):
...
+                7,
+                4,
+                2,
+                6,
+                4,
+                1,
+                5,
+                3,
+                7,
+                5,
```
arkid15r, line 132: This
```suggestion
                5
```

tests/test_calendars.py
```diff
@@ -110,3 +111,48 @@ def test_visakha_bucha_date(self):
...
+            ),
+            1,
```
arkid15r, line 134: and this should help with the "platforming"
```suggestion
            1
```

tests/test_calendars.py
```diff
@@ -110,3 +111,48 @@ def test_visakha_bucha_date(self):
...
+                27,
+                24,
+                29,
+                26,
+                30,
+                28,
+                25,
+                30,
+            ),
+            1,
```
arkid15r, line 155: Here too
```suggestion
                30
            ),
            1
```

arkid15r (commented):
👍 

I think it's totally fine make it a part of `calendars.py` for now. We'll move this helpers either to `HolidayBase` or a separate helpers.py later.

mborsetti:
I am amazed at the optimization you got with this: a spot check shows >7 faster calculation for both functions. These numbers are incredible.

I guess it also means that ``dateutil`` is not very optimized; makes me wonder if ``dateutil.easter`` can be made faster too (which incidentally would have the side benefit of removing the dependency on ``dateutil`` altogether... 🙂

Bravo 🥇

holidays/calendars.py
KJhellico: Is it correct to use "since" for dates **before**?

tests/test_calendars.py
KJhellico: Wow, that's great! I should do this trick in Malaysia test!

holidays/calendars.py
mborsetti: IMHO no. I suggest that the function be ``called _get_nth_weekday_from``, the argument called ``from``, and that the docstring address the option of a negative ``n``.

holidays/calendars.py
KJhellico: > I guess it also means that `dateutil` is not very optimized; makes me wonder if `dateutil.easter` can be made faster too (which incidentally would have the side benefit of removing the dependency on `dateutil` altogether... 🙂

Yes, `dateutil.relativedelta` is quite complex: `relativedelta` is a class, `weekday` is a class too. `easter` is a relatively simple function, we can only take it as it is. The most complex thing is `dateutil.parser`…

KJhellico:
> I guess it also means that `dateutil` is not very optimized; makes me wonder if `dateutil.easter` can be made faster too (which incidentally would have the side benefit of removing the dependency on `dateutil` altogether... 🙂

Yes, `dateutil.relativedelta` is quite complex: `relativedelta` is a class, `weekday` is a class too. `easter` is a relatively simple function, we can only take it as it is. The most complex thing is `dateutil.parser`…

holidays/calendars.py
mborsetti: Ooops...🤦

Sorry

holidays/calendars.py
KJhellico: Done

holidays/calendars.py
KJhellico: Done

holidays/calendars.py
KJhellico: Done

tests/test_calendars.py
KJhellico: Done

holidays/calendars.py
KJhellico: Changed args order, function renamed for now to `_get_nth_weekday_from`, date argument name still to be questionable.

arkid15r: You can use `from_dt` or just `dt` then.

arkid15r (commented):
LGTM

PR #990, opened 2023-02-28
Update Chile holidays. Add localization: en_US, es, uk

KJhellico:
## Proposed change

- Add Chile holidays localization: en_US, es (default), uk
- Some holidays original names correction

## Type of change

- [ ] New country holidays support (thank you!)
- [x] Supported country holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/tests/processes improvement (best practice, refactoring, optimization)
- [ ] Dependency upgrade (version update)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (adds functionality to python-holidays in general)

## Checklist

- [x] I've followed the [contributing guidelines][contributing-guidelines]
- [x] This PR is filed against `beta` branch of the repository
- [x] This PR doesn't contain any merge conflicts
- [x] The code style looks good (`make pre-commit`)
- [x] I've added tests to verify that the new code works and all tests pass locally (`make test`)
- [x] The related [documentation][docs] has been added/updated (check off the box for free if no updates is required)

PR #991, opened 2023-02-28
Refactor `HolidayBase::get_named`

arkid15r:
## Proposed change
The PR aims to impove `HolidayBase::get_named` flexibility in terms of available holiday name lookup types:
  - Add `lookup` parameter.
  - Add `contains/iexact/icontains/iexact` lookup types.
  - Keep the current behavior with `icontains` as default lookup type.

## Type of change

## Checklist

### DIFF
--- a/tests/countries/test_chile.py
+++ b/tests/countries/test_chile.py
@@ -270,12 +270,14 @@ class TestChile(TestCase):
         )
 
         self.assertHolidayName(
-            "Día del Respeto a la Diversidad",
+            "Día del Respeto a la Diversidad "
+            "[Day of the Meeting of Two Worlds]",
             Chile(years=range(2000, 2020)),
         )
 
         self.assertHolidayName(
-            "Día del Descubrimiento de dos Mundos",
+            "Día del Descubrimiento de dos Mundos "
+            "[Discovery of Two Worlds' Day]",
             Chile(years=range(2020, 2050)),
         )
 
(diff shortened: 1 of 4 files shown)


mborsetti:
What's the motivation behind this PR? 

- I am having a hard time imagining the use case requiring using case-sensitivity when searching for a holiday;
- I can sort-of-see a use for regex, e.g. to find an observed Founder's Day holiday, although even this could be achieved easily by using ``get_named("founder's day")`` and applying a list comprehension on the result to remove the returned holidays that don't have ``observed`` in their name.

In any case, I would consider **not** replacing the default logic from one using ``str in`` (which is highly performant) to one based on ``re`` one (which isn't).

KJhellico:
Somewhere in the tests I saw cases where the ability to check for **exact name** would be useful. But I agree that this is not a regular occurrence, so we need to consider performance.

arkid15r:
As was pointed out by @mborsetti there're no need in regex based logic for these cases.
I've changed the implementation based on the review. Also changed the common.TestCase holiday name related assertions to use exact lookup.

PTAL

tests/countries/test_chile.py
```diff
@@ -270,12 +270,14 @@ def test_columbus_day(self):
         )
 
         self.assertHolidayName(
```
KJhellico, line 272: No need to change this test, in #990 I've updated this holiday name, it was wrong :)

arkid15r: I believe I'll have to resolve this when I merge it :)

KJhellico (commented):
LGTM

PR #889, opened 2023-01-09
Split package dependencies

arkid15r:
Resolves #362 

Redesign extras_require logic.
  - Use `importlib` to determine specific package availability
  - Update README.rst
  - Use `mock` to recover 100% coverage
  - Add `pymeeus` option for Chile and Japan

Based on #729 
> Split out three dependencies that are only used for some specific
> countries so they're not installed by default anymore. This is a
> breaking change for consumers using one or more of the affected
> countries. The advantage is that consumers who only use one or a few
> simple calendars don't need to install dependencies that will be unused.

PR #993, opened 2023-03-01
Use the built-in library for time zone functions

mborsetti:
## Proposed change

Use the built-in library for timezone functions.

Benefits:
* Implemented best practice of using built-in libraries when possible;
* Improved speed of line 120 of ``chile.py`` by ~1.6x;
* Removed dependency on external package ``python-dateutil`` for timezone.

## Type of change

## Checklist

### DIFF
--- a/holidays/countries/chile.py
+++ b/holidays/countries/chile.py
@@ -25,7 +25,7 @@ from holidays.holiday_base import HolidayBase
 # use standard library for timezone
 try:
     from zoneinfo import ZoneInfo
-except ImportError:
+except ImportError:  # pragma: no cover
     from backports.zoneinfo import ZoneInfo  # type: ignore[no-redef]
 
 
--- a/holidays/countries/chile.py
+++ b/holidays/countries/chile.py
@@ -9,10 +9,9 @@
 #  Website: https://github.com/dr-prodigy/python-holidays
 #  License: MIT (see LICENSE file)
 
-from datetime import date, datetime
+from datetime import date, datetime, timezone
 from datetime import timedelta as td
 
-from dateutil import tz
 from dateutil.easter import easter
 from dateutil.relativedelta import MO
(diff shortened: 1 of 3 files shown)


PPsyrius:
As a minor note, Japan's "Vernal Equinox Day" (春分の日) got the same problem here so they could probably get the same treatment as the Chilean one here.

mborsetti:
> As a minor note, Japan's "Vernal Equinox Day" (春分の日) got the same problem here so they could probably get the same treatment as the Chilean one here.

Indeed, not sure how I missed it, thanks!  PR update forthcoming.

holidays/countries/chile.py
```diff
@@ -23,6 +22,12 @@
 from holidays.constants import TUE, WED, THU, FRI, SAT, SUN
 from holidays.holiday_base import HolidayBase
 
+# use standard library for timezone
+try:
+    from zoneinfo import ZoneInfo
+except ImportError:  # pragma: no cover
```
arkid15r, line 28: It's not a coverage improvement, it's a coverage decrease suppression :)

I had a similar problem in [this PR](https://github.com/dr-prodigy/python-holidays/pull/889/). It took me a while to figure out a better solution. I'll take care of the refactoring after #889 is merged.

arkid15r (commented):
Great optimization! LGTM

arkid15r:
Thanks for another great contribution Mike!

holidays/countries/chile.py
mborsetti: Yeah, I agree, it's definitely a suppression.

I can't think how to do it "correctly", so I look forward to seeing what you come up with! :)

arkid15r: You can find an example in https://github.com/dr-prodigy/python-holidays/pull/889, I'm going to use pretty much the same approach.

PR #1001, opened 2023-03-03
Optimize TestAllInSameYear::test_all_countries test

arkid15r:
## Proposed change

Avoid loading l10n files.

## Type of change

## Checklist


tests/test_utils.py
```diff
@@ -81,7 +82,8 @@ def test_exceptions(self):
 class TestAllInSameYear(unittest.TestCase):
     """Test that only holidays in the year(s) requested are returned."""
 
-    def test_all_countries(self):
+    @mock.patch("pathlib.Path.rglob", return_value=())
+    def test_all_countries(self, unused_rglob_mock):
```
KJhellico, line 86: About this test, I've been meaning to ask. In #606 @mborsetti proposed to change it so that it would only run in the one Python version (latest) and extend it to years 1950-2051. But for some reason, these changes were not merged in beta (perhaps it was accidentally lost in the process of complex merging). Maybe we should consider them?

KJhellico: This full test helped me to find several problems with Islamic calendar.

mborsetti: > This full test helped me to find several problems with Islamic calendar.

Did running in 3.10, 3.9, 3.8 and/or 3.7 surface issues that running in 3.11 didn't?  My suggestion IIRC was along the lines to do the heavy logic tests in 3.11 (faster) only, and only do the "typical" tests in the other Python versions.

KJhellico: Actually, a wider range of years helped me :) For example, 2006 is a special year in this regard, and it doesn't fall in existing range` (current_year - 10, current_year + 3)`.

mborsetti: Yeah, I think we're in agreement.  We should do a very heavy battery of logical tests (lots of years, etc.), but run those only on the latest Python version otherwise full testing is going to take forever and there's little insights to be gained from running, let's say, 1999 on all versions of Python.

arkid15r: @KJhellico, @mborsetti thanks for your input! I've addressed the mentioned issues.

The coverage decrease and `Set changed size during iteration` error have to be addressed in separate PRs.

Issue #998, opened 2023-03-01
Maximize load performance 


arkid15r:
We advertise this library as "fast, efficient", and is performance or memory footprint affected by loading every single class for every single country (that's 360+ classes) even when the user needs holidays only for a single country, which I assume this to be 90%+ of use cases, or just a few countries? What about on small footprints (e.g. Raspberry Pi)? I would assume so, but am not very familiar with Python innards.

If so, are there architectures that we can implement to maximize load performance especially when only one (or a few) countries are needed?

_Originally posted by @mborsetti in https://github.com/dr-prodigy/python-holidays/issues/948#issuecomment-1432382359_

mborsetti:
Some thoughts

- We're importing a ton of alpha-3 codes, and nowhere in the documentation we say we support them. I don't really see a reason why we should support them (they are superfluous to the alpha-2 codes that are very familiar due to their use as TLDs); maybe it's a good time to start **deprecating the existing alpha-3 codes** (and stop adding them when new countries are added)?
- Do we even want a long-form nonstandard for countries, which also opens ourselves up to discussions on political issues such as whether the class for HK should be HongKong vs. HongKongSAR (the latter being closer to the official name), CN be China vs. PeoplesRepublicOfChina and on and on?  
- I also found IRL that finding a class called "HongKong" ("HongKongSAR"?) or "Eswatini" or similar in the debugger is very confusing -- Eswatini what? _What is the Eswatini or China or whatever object supposed to be_? We should consider **calling the base class in the format of HolidaysHK**, HolidaysSZ, etc. to make it very clear when debugging what is the function of an object of this class. 

I can (slowly) work on all of this if there's interest.

adamchainz:
I have found this issue because I'm profiling a project's startup time (as per [this blog post](https://adamj.eu/tech/2023/03/02/django-profile-and-improve-import-time/)), and `holidays` is one of the slower-to-import libraries. It takes 56ms on my machine, most of which is indeed wasted on loading countries which the project doesn't need.

Here's a profile:

<details>
<summary><code>python -X importtime -c 'import holidays'</code></summary>
<pre><code>
import time: self [us] | cumulative | imported package
import time:       268 |        268 |   _io
import time:        40 |         40 |   marshal
import time:       539 |        539 |   posix
import time:       586 |       1430 | _frozen_importlib_external
import time:       198 |        198 |   time
import time:       142 |        340 | zipimport
import time:        94 |         94 |     _codecs
[...]

PR #1003, opened 2023-03-03
Unify country imports order

arkid15r:
## Proposed change

Make the country imports look consistent.

## Type of change

## Checklist


mborsetti:
Hmmm... Since we say that 

> The standard way to refer to a country is by using its [ISO 3166-1 alpha-2 code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes),  the same used for domain names,

and don't even list or document the (nonstandard) long form name, stylistically doesn't it make more sense to import the documented 2-letter code first?

Also, as an aside, 

- We're importing a ton of alpha-3 codes, and nowhere in the documentation we say we support them. I don't really see a reason why we should support them (they are superfluous to the alpha-2 codes that are very familiar due to their use as TLDs); maybe it's a good time to start **deprecating the existing alpha-3 codes** (and stop adding them when new countries are added)? That would start the work on #998.
- Do we even want a long-form nonstandard for countries, which also opens ourselves up to discussions on political issues such as whether the class for HK should be HongKong vs. HongKongSAR (the latter being closer to the official name), CN China vs. PeoplesRepublicOfChina and on and on?  
- I also found IRL that finding a class called "HongKong" ("HongKongSAR"?) or "Eswatini" or similar in the debugger is very confusing -- Eswatini what? What is the Eswatini object supposed to be? We should consider **calling the base class in the format of HolidaysHK**, HolidaysSZ, etc. to make it very clear when debugging what is the function of an object of this class.

KJhellico (commented):
LGTM

arkid15r:
> Hmmm... Since we say that
> 
> > The standard way to refer to a country is by using its [ISO 3166-1 alpha-2 code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes),  the same used for domain names,
> 
> and don't even list or document the (nonstandard) long form name, stylistically doesn't it make more sense to import the documented 2-letter code first?
> 
> Also, as an aside,
> 
> * We're importing a ton of alpha-3 codes, and nowhere in the documentation we say we support them. I don't really see a reason why we should support them (they are superfluous to the alpha-2 codes that are very familiar due to their use as TLDs); maybe it's a good time to start **deprecating the existing alpha-3 codes** (and stop adding them when new countries are added)? That would start the work on [Maximize load performance  #998](https://github.com/dr-prodigy/python-holidays/issues/998).
> * Do we even want a long-form nonstandard for countries, which also opens ourselves up to discussions on political issues such as whether the class for HK should be HongKong vs. HongKongSAR (the latter being closer to the official name), CN China vs. PeoplesRepublicOfChina and on and on?
> * I also found IRL that finding a class called "HongKong" ("HongKongSAR"?) or "Eswatini" or similar in the debugger is very confusing -- Eswatini what? What is the Eswatini object supposed to be? We should consider **calling the base class in the format of HolidaysHK**, HolidaysSZ, etc. to make it very clear when debugging what is the function of an object of this class.

Hi Mike,
these ideas are totally worth a discussion/implementation. I suggest creating a separate issue(s) (or discussion(s) -- up to you) as after this PR is submitted the content will be pretty much forsaken.

PR #1090, opened 2023-04-09
Introduce Islamic holidays calculation with pre-calculated data [WIP]

KJhellico:
## Proposed change

In terms of performance, it is more efficient to use pre-calculated date arrays than to calculate them every time. Now we have 9 Islamic holidays:
- Eid al-Fitr
- Eid al-Adha
- Mawlid (Birthday of Prophet)
- Islamic New Year
- Ashura
- Isra and Mi'raj
- Beginning of Ramadan
- Nuzul Al-Quran Day
- Hari Hol of Johor

So I made `_IslamicHolidays` class with the dates of these holidays in 1925-2077 (the entire range supported by the `hijri-converter` package). This method will allow us to get rid of dependencies on external packages and increase speed.

I look forward to your feedback and comments.

## Type of change

## Checklist


mborsetti:
Awesome idea!  Makes a ton of sense.

May I suggest that:

1. For years that fall outside of the precalculated range, the existing function `_islamic_to_gre` (based on `hijri_converte`) is used as opposed as to returning no dates;
2. `hijri_converter` be lazily loaded, and the appropriate instructive Exception message is raised from within `_islamic_to_gre` if it cannot be imported (not present in the system);
3. (to be considered/confirmed): `hijri_converter` be removed as a default dependency;
4. The documentation be updated with the range of precalculated years and the optional requirement of installing `_islamic_to_gre` for years outside the range.

Nice job!

KJhellico:
> Awesome idea! Makes a ton of sense.

Thank you, Mike!

> 1. For years that fall outside of the precalculated range, the existing function `_islamic_to_gre` (based on `hijri_converte`) is used as opposed as to returning no dates;

1925-2077 - it's complete range supported by `hijri_converter`. There's just need to add a few holidays from 1924 (exact range is 1924-08-01 … 2077-11-16).

mborsetti:
> > Awesome idea! Makes a ton of sense.
> 
> Thank you, Mike!
> 
> > 1. For years that fall outside of the precalculated range, the existing function `_islamic_to_gre` (based on `hijri_converte`) is used as opposed as to returning no dates;
> 
> 1925-2077 - it's complete range supported by `hijri_converter`. There's just need to add a few holidays from 1924 (exact range is 1924-08-01 … 2077-11-16).

I missed that...I assumed it was a far broader library! 😆

Then jettisoning `hijri_converter` as a dependency would make sense.

Thanks again for doing this!

mborsetti:
Nice!

I suggest that we wait to push it until documented (the documentation is falling behind the code IMHO).

PPsyrius:
> Speaking of which, we should probably start doing per-country documentation...

That would be a great idea, though a section on the standard expected support range for all countries (AFAIK, it's 1950-2050; judging from most test cases) would be a good idea —that and migrating certain sections of the readme to GitHub project's Wiki section instead.

Let's move this conversation to a new draft PR.

holidays/calendars.py
```diff
@@ -942,3 +945,1420 @@ def ok_phansa_date(self, year: int) -> Optional[date]:
         else:
             delta_days = 309
         return start_date + td(days=delta_days)
+
+
+class _IslamicHolidays:
```
arkid15r, line 950: We use *Holidays for holiday_groups normally. I guess these entities are more lie calendars. I'd use something similar, e.g., IslamicLunar

holidays/calendars.py
```diff
@@ -942,3 +945,1420 @@ def ok_phansa_date(self, year: int) -> Optional[date]:
...
+    DATES = {
+        # Eid al-Fitr
+        "eid-al-fitr": {
```
arkid15r, line 953: I liked non-nested version better. Also the names are quite self-explanatory.
If you migrated them just for dict attribute access then I suggest using getattr instead.

holidays/calendars.py
```diff
@@ -942,3 +945,1420 @@ def ok_phansa_date(self, year: int) -> Optional[date]:
...
+            2073: ((FEB, 13),),
+            2074: ((FEB, 2),),
+            2075: ((JAN, 23),),
+            2076: ((JAN, 12),),
+            2077: ((JAN, 1),),
+        },
+    }
+
+    @staticmethod
+    def islamic_holiday_date(year: int, hol_type: str) -> Set[date]:
```
arkid15r, line 2361: The other calendars use a method per holiday approach. I believe this class should copy that behaviour too.
It's totally fine to submit it as is for now.

arkid15r (commented):
This is a great PR! Thank you!

Please find my comments below:

holidays/calendars.py
KJhellico: Changed to separate holiday arrays.

holidays/calendars.py
KJhellico: I don't know how to implement `_add_islamic_calendar_holiday()` functionality in this case (for multi-day holidays). 🤷‍♂️

holidays/calendars.py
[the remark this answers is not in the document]
arkid15r: Yes, I think it'd be really useful. You can attach it here or file in a separate PR

holidays/calendars.py
arkid15r: The holidays dates would be used in the calendar methods directly (not via `_add_islamic_calendar_holiday`)

arkid15r (commented):
LGTM

Let's merge the PR and re-face it similar to existing calendars after. I guess this portion of code is a quite right amount to digest.

PR #1114, opened 2023-04-19
Refactor country subdivisions

arkid15r:
## Proposed change

This is an attempt to reincarnate PR #730 keeping external API the same (for now):
  - Change subdivisions type from `List[str]` to `Tuple[str, ...]`.
  - Make the same change for `_deprecated_subdivisions`.
  - Sort countries subdivisions alphabetically.
  - Move utils tests from test_holiday_base.py to test_utils.py

## Type of change

## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -353,7 +353,7 @@ subdivisions are available:
      - None
    * - New Zealand
      - NZ
-     - Regions: AUK, CAN, HKB, MBH, NSN, NTL, OTA, STL, TKI, WGN, WTC, CIT. Sub-regions: South Canterbury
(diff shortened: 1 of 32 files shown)


KJhellico (commented):
LGTM. I don't think anyone needs to add or remove provinces from an existing `Holidays` instance (which would require a list).

arkid15r:
> LGTM. I don't think anyone needs to add or remove provinces from an existing `Holidays` instance (which would require a list).

That's what I'd expect too. However I'm thinking of making it a set (right?) so keeping it a list for now for compatibility with the current behavior. If not, we can use it as is and change to tuple later.

PR #1148, opened 2023-05-05
Introduce Asian holidays pre-calculated data

KJhellico:
## Proposed change

Introduce pre-calculated data for Chinese, Korean, Hindu lunisolar holidays instead of using internal `_ChineseLuniSolar` class and `korean-lunar-calendar` package. `_AsianLunisolar` class contains dates for 1901-2099 years for holidays:
- Lunar New Year (CN, HK, ID, KR, MY, PH, SG, TW, VN)
- Hung Kings' Commemoration Day (VN)
- Buddha's Birthday (HK, KR)
- Vesak (ID, SG)
- Dragon Boat Festival (CN, HK, TW)
- Mid-Autumn Festival (CN, HK, KR, TW)
- Chung Yeung Festival (HK)
- Vesak (MY)
- Diwali (MY, SG)
- Thaipusam (MY)

## Type of change

## Checklist


holidays/calendars/oriental.py
arkid15r, line 1: Hey @mborsetti, we need your input here: is the new calendar name acceptable/polite enough in your opinion?
If not - any ideas regarding some alternative names?

Thank you!

arkid15r (commented):
That's a major step towards our zero-dependency goal (at least calendar-wise).
Please allow some time for reviewing this.

Thank you!

holidays/calendars/oriental.py
mborsetti: Hi @arkid15r & @KJhellico, some great work here!  

@arkid15r sorry I'm swamped, so thanks for reaching out to me.

I would advise using "Asian" instead of "Oriental" for the module name and the holiday_group's class name, as the name of the continent is the (current) preferred way to refer to this region.

I would also strongly advise that the name of the lunisolar calendar class remain "Chinese" since this is the calendar's official and appropriate name, and it distinguishes it from the other Asian lunisolar calendars such as the Buddhist, Burmese, Jain, Kurdish, traditional Hindu, Japanese, Korean, Mongolian, Tibetan, and Vietnamese ones.

mborsetti: > At first, I also thought about "Asian", but Asian is a more general term (it includes Persian, Jewish and Middle East as well).

Orient does as well: https://en.wikipedia.org/wiki/Orient

> Actually in our case it's more a collection of dates than "calendar"...

True! A collection of dates derived from the Chinese lunisolar calendar!

KJhellico: > Orient does as well: https://en.wikipedia.org/wiki/Orient

Hmm, I agree. :)

> True! A collection of dates derived from ... the Chinese lunisolar calendar!

But Diwali, Thaipusam and, in fact, Vesak - **not** Chinese.

mborsetti: > But Diwali, Thaipusam and, in fact, Vesak - **not** Chinese.

Agreed, missed this. Then ``_LunisolarAsian`` or ``_AsianLunisolar``?

arkid15r: > Agreed, missed this. Then `_LunisolarAsian` or `_AsianLunisolar` would be my _suggestion_.

👍 _AsianLuniSolar seems to be the most balanced option in my opinion.

mborsetti: For the pedants out there, it would be capitalized `_AsianLunisolar` since lunisolar is a single word (https://www.britannica.com/topic/lunisolar-calendar)

> > Agreed, missed this. Then `_LunisolarAsian` or `_AsianLunisolar` would be my _suggestion_.
> 
> 👍 _AsianLuniSolar seems to be the most balanced option in my opinion.

KJhellico: So, `_AsianLunisolar` and `AsianCalendarHolidays` ✌️

arkid15r: As an alternative we could split the Chinese and other holidays back to their respective calendars and use them separately keeping the `ChineseLunisolar` name. I believe this would be a better solution structure/name wise.

holidays/holiday_groups.py
```diff
@@ -469,124 +469,322 @@ def _add_whit_sunday(self, name) -> date:
...
+        if days_delta != 0:
+            dt += td(days=days_delta)
         return self._add_holiday(
+            f"{name}* (*{estimated_label})"
+            if is_estimated and show_estimated
+            else name,
+            dt,
+        )
+
+    def _add_chinese_buddha_birthday(self, name) -> Optional[date]:
```
arkid15r, line 576:
```suggestion
    def _add_chinese_birthday_of_buddha(self, name) -> Optional[date]:
```
sounds better to my ear, please consider renaming.

tests/countries/test_south_korea.py
```diff
@@ -18,20 +18,21 @@
 class TestSouthKorea(TestCase):
     @classmethod
     def setUpClass(cls):
-        years = [1944, 1945, 1947, 1948, 1949, 1960, 1974, 1975] + list(
-            range(1981, 2031)
-        )
+        years = [1948, 1949, 1960, 1974, 1975] + list(range(1981, 2050))
```
arkid15r, line 21:
```
super().setUpClass(
    SouthKorea,
    years=(1948, 1949, 1960, 1974, 1975) + tuple(range(1981, 2050)),
)
```

arkid15r (commented):
🚀 LGTM

holidays/holiday_groups.py
KJhellico: Done.

tests/countries/test_south_korea.py
KJhellico: Done.

PR #1181, opened 2023-05-13
Add Belize holidays

### LINKED ISSUE #1154: Add Belize holidays
Useful links:
  - https://en.wikipedia.org/wiki/Public_holidays_in_Belize
  - https://www.pressoffice.gov.bz/public-and-bank-holidays-2022-updated/
  - https://www.timeanddate.com/holidays/belize/

KJhellico:
## Proposed change

Add Belize holidays (close #1154).

## Type of change

## Checklist

### DIFF
--- a/holidays/countries/belize.py
+++ b/holidays/countries/belize.py
@@ -122,6 +122,9 @@ class Belize(HolidayBase, ChristianHolidays, InternationalHolidays):
         # Garifuna Settlement Day.
         self._add_holiday("Garifuna Settlement Day", NOV, 19)
 
+        # Populated before Christmas for right obvserved Christmas calculation
+        # (if Christmas falls on Sunday, there should be no observed Christmas
+        # on Monday)
         # Boxing Day.
         self._add_christmas_day_two("Boxing Day")
 
--- a/holidays/countries/belize.py
+++ b/holidays/countries/belize.py
@@ -1,0 +1,137 @@
+#  python-holidays
+#  ---------------
+#  A fast, efficient Python library for generating country, province and state
+#  specific sets of holidays on the fly. It aims to make determining whether a
+#  specific date is a holiday as fast and flexible as possible.
+#
+#  Authors: dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
+#           ryanss <ryanssdev@icloud.com> (c) 2014-2017
+#  Website: https://github.com/dr-prodigy/python-holidays
+#  License: MIT (see LICENSE file)
+
+from datetime import date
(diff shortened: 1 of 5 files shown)


holidays/countries/belize.py
```diff
@@ -0,0 +1,137 @@
...
+    def _add_movable_holiday(self, *args) -> None:
+        # Chapter 289 of the laws of Belize states that if the holiday falls
+        # on a Sunday or a Friday, the following Monday is observed as public
+        # holiday; further, if the holiday falls on a Tuesday, Wednesday or
+        # Thursday, the preceding Monday is observed as public holiday
+        name, dt = self._parse_holiday(*args)
+        if self.observed:
+            if self._is_friday(dt) or self._is_sunday(dt):
+                dt = _get_nth_weekday_from(1, MON, dt)
+                name = "%s (Observed)" % name
```
arkid15r, line 47: Could be f-string, this one is also fine.

holidays/countries/belize.py
```diff
@@ -0,0 +1,137 @@
...
+        # Good Friday.
+        self._add_good_friday("Good Friday")
+
+        # Holy Saturday.
+        self._add_holy_saturday("Holy Saturday")
+
+        # Easter Monday.
+        self._add_easter_monday("Easter Monday")
+
+        # Labour Day.
```
arkid15r, line 97: Labor?

holidays/countries/belize.py
```diff
@@ -0,0 +1,137 @@
...
+            if year >= 2021
+            else "Pan American Day",
+            date(year, OCT, 12),
+        )
+
+        # Garifuna Settlement Day.
+        self._add_holiday("Garifuna Settlement Day", NOV, 19)
+
+        # Boxing Day.
+        self._add_christmas_day_two("Boxing Day")
```
arkid15r, line 126: Should be swapped with Christmas day.

arkid15r (commented):
🇧🇿 LGTM

Please see the minor issues I've spotted during the review

holidays/countries/belize.py
KJhellico: I used "Labour" here because Belize is a Commonwealth country, and it's like British version. :)

holidays/countries/belize.py
KJhellico: Boxing Day placed first for right obvserved Christmas calculation (if Christmas falls on Sunday, there should be no observed Christmas on Monday; and in this way, that day is already not empty). I couldn't invent a better way to do it. Only to use `observed_dates` with cycle at the end method.

PR #1294, opened 2023-06-07
Add Cambodia holidays

PPsyrius:
## Proposed change

Add Cambodia holidays (`en_US`, `km`, `th` localization).

Closes #1160.

## Type of change

- [x] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency upgrade (version update)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new python-holidays functionality in general)

## Checklist

- [x] I've followed the [contributing guidelines][contributing-guidelines]
- [x] This PR is filed against `beta` branch of the repository
- [x] This PR doesn't contain any merge conflicts and has clean commit history
- [x] The code style looks good: `make pre-commit`
- [x] All tests pass locally: `make test`, `make tox` (we strongly encourage adding tests to your code)
- [x] The related [documentation][docs] has been added/updated (check off the box for free if no updates is required)

[contributing-guidelines]: https://github.com/dr-prodigy/python-holidays/blob/beta/CONTRIBUTING.rst
[docs]: https://github.com/dr-prodigy/python-holidays/tree/beta/docs/source

PR #1300, opened 2023-06-09
Add `_add_childrens_day(...)` to `InternationalHolidays`

PPsyrius:
## Proposed change

This PR adds `_add_childrens_day(...)` to `InternationalHolidays`, currently used by Romania, Moldova, and upcoming Cambodia (#1294) holiday support.

The new `InternationalHolidays` is defaulted to `June 1` date, though another popular variation, November 20th is also supported via `self._add_childrens_day(name, NOV)`

## Type of change

## Checklist


holidays/groups/international.py
```diff
@@ -11,14 +11,25 @@
...
-from holidays.constants import JAN, MAR, MAY, OCT, NOV, DEC
+from holidays.constants import JAN, MAR, MAY, JUN, OCT, NOV, DEC
 
 
 class InternationalHolidays:
     """
     International holidays.
     """
 
+    def _add_childrens_day_jun1(self, name):
```
arkid15r, line 22: I believe we can combine these 2 holidays within one method with an optional parameter, for example`def _add_childrens_day(self, name, month="JUN"):` or similar. Adding _jun1 suffix to the name seems inconsistent with existing methods. It may be even better to keep it simple and not add the November related stuff at all until we need it.

PPsyrius: Fine for me IMO

PPsyrius:
The suggestion is now implemented for wider support, though the coverall test seems dead for some reason, even if the local `tox` test works perfectly.

tests/test_holiday_groups.py
```diff
@@ -37,3 +38,27 @@ def _populate(self, year):
...
+class TestInternationalHolidays(TestCase):
+    def test_add_childrens_day(self):
+        class TestHolidays(HolidayBase, InternationalHolidays):
+            def __init__(self, *args, **kwargs):
+                InternationalHolidays.__init__(self)
+                super().__init__(*args, **kwargs)
+
+            def _populate(self, year):
+                super()._populate(year)
+
```
arkid15r, line 52:
```suggestion
```

tests/test_holiday_groups.py
```diff
@@ -37,3 +38,27 @@ def _populate(self, year):
...
+        test_holidays = TestHolidays()
+
+        test_holidays._populate(2022)
+        test_holidays._add_childrens_day("Children's Day (November 20)", NOV)
```
arkid15r, line 56: Just a note: using "NOV" instead of NOV would make NOV importing unnecessary with pretty much the same readability.

arkid15r (commented):
LGTM

tests/test_holiday_groups.py
PPsyrius: That works for me - I will switch to that version then 🙏

KJhellico:
It has been merged into beta. Now you can use it for Cambodia holidays ;)

PR #1309, opened 2023-06-14
Migrate Curaçao to holiday groups

PPsyrius:
## Proposed change

- Migrate Curaçao holidays to `holiday_groups` standard.
- Add l10n supports for `en_US`, `nl`, `pp`, `uk` localization; `uk`'s l10n need some double checking from @KJhellico .
 (`pp` currently acts as stand in for ABC islands' Portuguese-creole Papiamento's: `pap`)

## Type of change

- [ ] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [x] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency upgrade (version update)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new python-holidays functionality in general)

## Checklist

- [x] I've followed the [contributing guidelines][contributing-guidelines]
- [x] This PR is filed against `beta` branch of the repository
- [x] This PR doesn't contain any merge conflicts and has clean commit history
- [x] The code style looks good: `make pre-commit`
- [x] All tests pass locally: `make test`, `make tox` (we strongly encourage adding tests to your code)
- [x] The related [documentation][docs] has been added/updated (check off the box for free if no updates is required)

[contributing-guidelines]: https://github.com/dr-prodigy/python-holidays/blob/beta/CONTRIBUTING.rst
[docs]: https://github.com/dr-prodigy/python-holidays/tree/beta/docs/source

Issue #784, opened 2022-11-13
Zurich half-day holidays in Switzerland


dragoon:
Zurich has 2 half-day holidays every year: [Sechseläuten](https://en.wikipedia.org/wiki/Sechsel%C3%A4uten) and [Knabenschiessen](https://en.wikipedia.org/wiki/Knabenschiessen).
I wonder if we can somehow include this in the library?

But then not sure how to mark them. I don't know if any other country has something similar so it is worth doing.

dr-prodigy:
Hi @dragoon thank you!
I can't give a final answer to your question, but if there are other countries with similar cases in existence, as of now we surely neither currently manage, nor even know them :-D

This said, to avoid implementing an impacting change for those 2 half days only (which I currently don't see a point in), I would suggest (very trivially..) to manage them just adding a note to their names, such as "Sechseläuten (half day)" or similar.. as we already do with observed holidays, in other words.
Not very accurate, while still better than nothing! 👍

mborsetti:
I observe that a similar situation takes place for NYSE, which has so-called reduced trading hours days when it closes at 13:00 (see footnotes at https://www.nyse.com/markets/hours-calendars#holidays).

Having some sort of standardized mechanism to handle those days would be nice and potentially very useful. Maybe add a `partial` boolean argument to the method, which if set to True will cause partial holidays to be returned as well as the full day ones, and that the partial ones have a label which ends with a country-standardized annotation in brackets (e.g., for NYSE, "Day after Thanksgiving [13:00 close]"?

alexanderschulze:
(+1) We have it in Germany too for Christmas Eve (12-24) and New Years Eve (12-31). Right now, we manually add these edge cases in our application, but support for a flag would be preferred .

onasipim:
Knabenschiessen seems wrong in 2026, it should only be listed on 14.09.26 as a half-day, not also on 12.09.26 and 13.09.26. Additionally it's actually only a holiday for the city of Zurich, but that can't be modeled in the library.

https://www.stadt-zuerich.ch/de/politik-und-verwaltung/arbeiten-bei-der-stadt/gut-zu-wissen/ferien-urlaub/feiertage-betriebsferientage-bft.html

PPsyrius:
[...]

PR #1320, opened 2023-06-20
Introduce holiday categories support

KJhellico:
## Proposed change

Currently, Python Holidays supports only **main** country holidays ("state holidays", or "federal holidays", or "paid holidays" etc). But many countries have other categories of holidays: bank holidays, school holidays, additional (paid or non-paid) holidays, religious holidays (valid only for these religions followers). So this PR introduce initial holiday categories support.

Set of existing holidays matches "PUBLIC" category (that is populated by default, if `categories` value is not specified).

For example:
```python
>>> from holidays import country_holidays
>>> from holidays.categories import BANK, PUBLIC
>>> at = country_holidays("AT")  # "common" holidays as default
>>> "2023-12-31" in at
False
>>> at_full = country_holidays("AT", categories=(BANK, PUBLIC))
>>> "2023-12-31" in at_full
True
>>> at_full["2023-12-31"]
Silvester
```
## Type of change

## Checklist


PPsyrius:
Good work IMO, although I got some questions here:

Should half-day holidays i.e. #784 or Curacao's New Year's Eve holiday (see in-code comment in #1309), be included in this PR or is it considered an entirely-different feature altogether?

For religious holidays another good example might be Argentina where [Muslim and Jewish citizens got extra holidays on top of mainstream Christian ones](https://www.argentina.gob.ar/interior/feriados-nacionales-2023), as well as an additional holidays that apply to citizens of Armenian origin only.

KJhellico:
> Should half-day holidays i.e. #784 or Curacao's New Year's Eve holiday (see in-code comment in #1309), be included in this PR or is it considered an entirely-different feature altogether?

Yes, I think such holidays can be added as "half-day" category (or something like). Maybe with (or without) some additional label...

> For religious holidays another good example might be Argentina where [Muslim and Jewish citizens got extra holidays on top of mainstream Christian ones](https://www.argentina.gob.ar/interior/feriados-nacionales-2023), as well as an additional holidays that apply to citizens of Armenian origin only.

These are exactly the cases I meant!

holidays/countries/austria.py
```diff
@@ -19,6 +19,7 @@
 class Austria(HolidayBase, ChristianHolidays, InternationalHolidays):
     country = "AT"
     default_language = "de"
+    supported_categories = ("bank", "common")
```
arkid15r, line 22: Not sure if `public` instead of `common` would be more appropriate here.

holidays/holiday_base.py
```diff
@@ -633,6 +642,12 @@ def _populate(self, year: int) -> Set[Optional[date]]:
 
         return dates
 
+    def _populate_categories(self):
```
arkid15r, line 645: The method names for subdiv and category holidays need to be aligned, e.g., either `_add_subdiv_holidays` and `_add_categories_holidays` or `_populate_subdiv_holidays` and `_populate_categories_holidays`. 

I'd also put both of them after special holidays part if possible.

holidays/utils.py
```diff
@@ -72,6 +73,9 @@ def country_holidays(
         language translation is not supported the original holiday names
         will be used.
 
+    :param categories:
+        Requsted holiday categories.
```
arkid15r, line 77:
```suggestion
        Requested holiday categories.
```

holidays/holiday_base.py
```diff
@@ -211,6 +211,10 @@ def _populate(self, year):
     """Country weekend days."""
     default_language: Optional[str] = None
     """The entity language used by default."""
+    categories: Optional[Tuple[str]] = None
+    """Requested holiday categories"""
```
arkid15r, line 215:
```suggestion
    """Requested holiday categories."""
```

holidays/holiday_base.py
```diff
@@ -211,6 +211,10 @@ def _populate(self, year):
...
+    supported_categories: Tuple[str, ...] = ()
+    """All holiday categories supported by this entity"""
```
arkid15r, line 217:
```suggestion
    """All holiday categories supported by this entity."""
```

holidays/holiday_base.py
```diff
@@ -223,6 +227,7 @@ def __init__(
         prov: Optional[str] = None,  # Deprecated.
         state: Optional[str] = None,  # Deprecated.
         language: Optional[str] = None,
+        categories: Optional[Tuple[str]] = None,
```
arkid15r, line 230: This needs to be reflected as a :param below.

arkid15r (commented):
This PoC looks great!

Besides the naming suggestions below I'd also like you to address category support check, i.e. raise `NotImplementedError` (duh) for cases when a requested category isn't supported.

holidays/holiday_base.py
KJhellico: I considered "populate" to be something more _global_ than "add", so I named it that way. BTW, need to implement categories support for subdivisions too 🙄

holidays/countries/austria.py
KJhellico: I was thinking about it.

holidays/utils.py
KJhellico: Done.

holidays/holiday_base.py
KJhellico: Done.

holidays/holiday_base.py
arkid15r: In the end both of them do the same stuff -- update the internal dict with new items. Their only difference to me is the update context: category vs subdiv.
I'm sure it's okay to keep the names as is within this PR.

holidays/countries/austria.py
```diff
@@ -19,6 +19,7 @@
 class Austria(HolidayBase, ChristianHolidays, InternationalHolidays):
     country = "AT"
     default_language = "de"
+    supported_categories = ("bank", "public")
```
arkid15r, line 22: It'd be better to introduce constants for each category and use them instead of literals.

arkid15r (commented):
👍 

Just one more idea for improvement:

KJhellico:
> This PoC looks great!
> 
> Besides the naming suggestions below I'd also like you to address category support check, i.e. raise `NotImplementedError` (duh) for cases when a requested category isn't supported.

It's worth to think about it. 👍

KJhellico:
But then we need to switch all countries at once to support categories. In order to have `supported_categories = ("public",)` everywhere.

holidays/holiday_base.py
```diff
@@ -288,6 +298,10 @@ def __init__(
                     DeprecationWarning,
                 )
 
+            wrong = set(self.categories).difference(ALL_CATEGORIES)
+            if len(wrong) > 0:
+                raise NotImplementedError(f"Categories does not supported: '{', '.join(wrong)}'.")
```
arkid15r, line 303:
```suggestion
            unknown_categories = self.categories.difference(ALL_CATEGORIES)
            if len(unknown_categories) > 0:
                raise NotImplementedError(
                    f"Category is not supported: {', '.join(unknown_categories)}."
                )

```

tests/countries/test_austria.py
```diff
@@ -157,6 +157,14 @@ def test_2022(self):
             ("2022-12-26", "Stefanitag"),
         )
 
+    def test_bank_2022(self):
+        self.assertHolidays(
+            Austria(categories=("BANK",), years=2022),
```
arkid15r, line 162: Now you can use the constant.

arkid15r (commented):
Some suggestions:

holidays/categories.py
arkid15r, line 1: Not sure if we need another file for this. Can we use `constants.py` instead?

holidays/holiday_base.py
```diff
@@ -223,6 +238,7 @@ def __init__(
         prov: Optional[str] = None,  # Deprecated.
         state: Optional[str] = None,  # Deprecated.
         language: Optional[str] = None,
+        categories: Optional[Set[str]] = None,
```
arkid15r, line 241: However, this one is potentially a bad choice:
```python
>>> import holidays
>>> categories = {holidays.constants.PUBLIC}
>>> us = holidays.US(categories=categories)
>>> ua = holidays.UA(categories=categories)
>>> us.categories
{'public'}
>>> ua.categories
{'public'}
>>> us.categories.add(holidays.constants.BANK)
>>> us.categories
{'public', 'bank'}
>>> ua.categories
{'public', 'bank'}
```

I suggest keeping a Tuple here.

holidays/utils.py
```diff
@@ -36,6 +36,7 @@ def country_holidays(
     prov: Optional[str] = None,
     state: Optional[str] = None,
     language: Optional[str] = None,
+    categories: Optional[Set[str]] = None,
```
arkid15r, line 39: This has to be consistent w/ `HolidayBase`.

arkid15r (commented):
LGTM

PPsyrius:
@KJhellico Hi, I'm working on extending the testcases right now for Thailand Holidays categories and yet it keep returning non-Bank/Government Holidays for my test results:
```python
    def test_raeknakhwan(self):
        name = "วันพืชมงคล"
        self.assertHolidays(
            Thailand(categories={GOVERNMENT}, years=range(1997, 2024)),
            ("1997-05-13", name),
            ("1998-05-13", name),
            ("2000-05-15", name),
            ("2001-05-16", name),
            ("2002-05-09", name),
            ("2003-05-08", name),
            ("2004-05-07", name),
            ("2005-05-11", name),
            ("2006-05-11", name),
            ("2007-05-10", name),
            ("2008-05-09", name),
            ("2009-05-11", name),
            ("2010-05-10", name),
            ("2011-05-13", name),
            ("2012-05-09", name),
            ("2013-05-13", name),
            ("2014-05-09", name),
            ("2015-05-13", name),
            ("2016-05-09", name),
            ("2017-05-12", name),
            ("2018-05-14", name),
            ("2019-05-09", name),
            ("2020-05-11", name),
            ("2021-05-13", name),
            ("2022-05-17", name),
            ("2023-05-11", name),
        )

    def test_bank_holiday(self):
        a_name = "วันหยุดเพิ่มเติมสำหรับการปิดบัญชีประจำปีของธนาคารเพื่อการเกษตรและสหกรณ์การเกษตร"
        m_name = "วันหยุดภาคครึ่งปีของสถาบันการเงินและสถาบันการเงินเฉพาะกิจ"

        self.assertHolidays(
            Thailand(categories={BANK}, years=range(2012, 2023)),
            ("2012-04-01", a_name),
            ("2012-07-01", m_name),
            ("2013-04-01", a_name),
            ("2013-07-01", m_name),
            ("2014-04-01", a_name),
            ("2014-07-01", m_name),
            ("2015-04-01", a_name),
            ("2015-07-01", m_name),
            ("2016-04-01", a_name),
            ("2016-07-01", m_name),
            ("2017-04-01", a_name),
            ("2017-07-01", m_name),
            ("2018-04-01", a_name),
            ("2018-07-01", m_name),
            ("2019-04-01", a_name),
            ("2020-04-01", a_name),
            ("2021-04-01", a_name),
        )
```

KJhellico:
> @KJhellico Hi, I'm working on extending the testcases right now for Thailand Holidays categories and yet it keep returning non-Bank/Government Holidays for my test results:

Did you replace `_populate()` with `_populate_public_holidays()` for Thailand?

KJhellico:
I see, problem is special holidays, they are always populated :)

PPsyrius:
> I see, problem is special holidays, they are always populated :)

Oh, that explains it

PR #1328, opened 2023-06-22
Introduce custom exceptions [WIP]

arkid15r:
## Proposed change

This PR Introduces PH custom exceptions. The goal is to share it early in order to gather naming and use case suggestions.
This is a breaking change. The documentation updates are coming soon.

## Type of change

- [x] Breaking change (a code change causing existing functionality to break)

## Checklist


PPsyrius:
IMO seems good to me, though we might want to add actionable suggestions on top of existing errors [as the ones slowly rolled out by the Python Foundation since 3.10 release](https://realpython.com/python312-error-messages/).

For example, rather than:
```python
f"Country '{self.country}' does not have subdivision " f"'{subdiv}'"
```
We might want to change to:
```python
f"Country '{self.country}' does not have subdivision " f"'{subdiv}'. Consider consulting f"{github}" for available subdivision list."
```
Or even better, get this to read from `readme.rst` file somehow:
```python
f"Country '{self.country}' does not have subdivision " f"'{subdiv}'. Available subdivisions for this country: f"{list}""
```

holidays/holiday_base.py
```diff
@@ -275,10 +276,14 @@ def __init__(
         if not isinstance(self, HolidaySum):
             if subdiv and subdiv not in set(self.subdivisions + self._deprecated_subdivisions):
                 if hasattr(self, "market"):
-                    error = f"Market '{self.market}' does not have subdivision " f"'{subdiv}'"
+                    error_message = (
+                        f"Market '{self.market}' does not have subdivision " f"'{subdiv}'"
+                    )
```
KJhellico, line 281:
```suggestion
                    error_message = f"Market '{self.market}' does not have subdivision '{subdiv}'"
```

holidays/holiday_base.py
```diff
@@ -275,10 +276,14 @@ def __init__(
...
                 else:
-                    error = f"Country '{self.country}' does not have subdivision " f"'{subdiv}'"
-                raise NotImplementedError(error)
+                    error_message = (
+                        f"Country '{self.country}' does not have subdivision " f"'{subdiv}'"
```
KJhellico, line 284:
```suggestion
                        f"Country '{self.country}' does not have subdivision '{subdiv}'"
```

KJhellico (commented):
I continue to have some doubts about the need for exceptions for Israel and (especially) Japan, but I like the idea of custom exceptions. 👍

KJhellico:
> Or even better, get this to read from `readme.rst` file somehow:
> 
> > f"Country '{self.country}' does not have subdivision " f"'{subdiv}'. Available subdivisions for this country: f"{list}""

`self.subdivisions` 😛

PPsyrius:
> > Or even better, get this to read from `readme.rst` file somehow:
> > 
> > > f"Country '{self.country}' does not have subdivision " f"'{subdiv}'. Available subdivisions for this country: f"{list}""
> 
> `self.subdivisions` 😛

Yeah, I forgot that exists somehow 💀

arkid15r:
> We might want to change to:
> 
> > f"Country '{self.country}' does not have subdivision " f"'{subdiv}'. Consider consulting f"{github}" for available subdivision list."
> 

So the general idea is to make the error messages more descriptive, right?


> I continue to have some doubts about the need for exceptions for Israel and (especially) Japan, but I like the idea of custom exceptions. 👍

This can be addressed separately, however I'd like to hear more on why the year out of range exception make you being doubtful.

arkid15r:
> Start year - as for many other countries it's a year when current holidays list was established. We typically use `if year <= >N: return None` for that. End year - for Israel, it may be an exception (due to used calendar implementation, we really have >no holidays data for later years). For Japan, I see no reason to limit. I think it's left from the time when observed holidays >were hardcoded.

My plan is to introduce start/end year control later and raise or warn when user requests holidays out of range of supported years (instead of returning an empty dict). Both start/end year range would be optional and enforced if they were set only.

So I guess `YearOutOfRangeError` can be ignored/removed for now.


>Right, same goes for cases where user's country input doesn't exists due to typo i.e. Lybia instead of Libya - in ideal >scenario the code would try to point out what they might meant to type after the initial exception message:

That's a good user experience suggestion. Eventually we'd like to have this implemented 👍

arkid15r:
I removed `YearOutOfRangeError`. As I mentioned it's early shared changes which I'm going to target to v0.30.
Please take another look -- any suggestions are welcome.

PR #1342, opened 2023-06-28
Migrate Marshall Islands to `holiday_groups`

PPsyrius:
## Proposed change

This PR migrates Marshal Islands to `holiday_groups`, as well as fixing incorrect in-lieu observances and extending support to 2020 with data from Internet Archives.

## Type of change

## Checklist

### DIFF
--- a/holidays/countries/marshall_islands.py
+++ b/holidays/countries/marshall_islands.py
@@ -13,7 +13,6 @@ import warnings
 from datetime import date
 from datetime import timedelta as td
 
-from holidays.calendars import _get_nth_weekday_of_month
 from holidays.constants import MAR, MAY, JUL, SEP, NOV, DEC, FRI
 from holidays.holiday_base import HolidayBase
 from holidays.holiday_groups import ChristianHolidays, InternationalHolidays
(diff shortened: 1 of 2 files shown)


holidays/countries/marshall_islands.py
```diff
@@ -13,78 +13,90 @@
...
 
     country = "MH"
+
+    # Special Holidays
+
+    # General Election Day
+    election_day = "General Election Day"
+
     special_holidays = {
-        2019: (NOV, 18, "General Election Day"),  # Third Monday in November
```
KJhellico, line 31: If we know the rule for this holiday, we can make it common, not special, like:
```python
        # General Election Day
        if year >= 1995 and (year - 1995) % 4 == 0:
            self._add_holiday("General Election Day", _get_nth_weekday_of_month(3, MON, NOV, year))

```

PPsyrius: Fine by me, though these events are infrequent enough that special holidays work. On top of that, who knows if these elections will get postponed or held earlier via snap elections in the future.

PPsyrius: Turns out the 1999 ones took place on the 4th Monday (NOV 22) rather than the 3rd (NOV 15) so the actual code would look like this:
```python
        if year >= 1995 and (year - 1995) % 4 == 0:
            self._add_holiday(
                # General Election Day
                "General Election Day",
                date(year, NOV, 22) if year == 1999 
                else self._get_nth_weekday_of_month(3, MON, NOV)
            )
```
Lemme know if you prefer this approach or whether sticking to the special holidays works better @KJhellico

KJhellico: Let's leave it as a special holidays :)

arkid15r (commented):
🇲🇭 LGTM

KJhellico:
Merged into beta. Thank you!

PR #1347, opened 2023-06-29
Add special holidays categories support

KJhellico:
## Proposed change

- Add special holidays categories support (special_public_holidays, special_bank_holidays etc)
- Add "workday" category, for official state holidays that are not non-working days (October 28 in Slovakia as an example)

## Type of change

## Checklist

### DIFF
--- a/holidays/constants.py
+++ b/holidays/constants.py
@@ -23,6 +23,7 @@ GOVERNMENT = "government"
 HALF_DAY = "half_day"
 PUBLIC = "public"
 SCHOOL = "school"
+WORKDAY = "workday"
 
 CHINESE = "chinese"
 CHRISTIAN = "christian"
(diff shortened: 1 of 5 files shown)


PPsyrius:
> `workday` category

That would really come in handy for all those "officially observed but not a non-working day list" for Thailand's "National Day of Importance" 20+ entries here:
image

Also, can you clarify a bit more on `special_public_holidays` - these are functionally the same as `special_holidays`, just that these only apply to the `PUBLIC` category, right?

holidays/constants.py
```diff
@@ -42,4 +43,5 @@
     ISLAMIC,
     PUBLIC,
     SCHOOL,
+    WORKDAY,
```
arkid15r, line 46: The category list keeps growing. It has to be documented when/how each of the categories should be used. Ideally with some use case examples.

arkid15r (commented):
LGTM

holidays/constants.py
arkid15r: I'll work w/ you on that soon, filed #1348 for now.

PR #1365, opened 2023-07-09
Introduce substituted holidays [PoC]

KJhellico:
## Proposed change

Some countries have a well-established practice of moving the working day between public holiday and weekend to another (non-working) day to create a long continuous holidays. For example, in Ukraine in 2021, Independence Day, 24.08, fell on a Tuesday, so the working day from Monday 23.08 was moved to Saturday 28.08, and Monday became a day off.

To support such cases by Python Holidays, I propose the following structure:
```python
    substituted_holidays = {
        ...
        2021: (
            (JAN, 16, JAN, 8),
            (AUG, 28, AUG, 23),
            (OCT, 23, OCT, 15),
        ),
}
```
when 1st pair (AUG, 28) is source non-working date ("date from"), 2nd (AUG, 23) - is target date ("date to"), that becomes a day off.

Substitution from other year is also supported now:
```python
    substituted_holidays = {
        2001: (
            (APR, 28, APR, 30),
            (MAY, 5, MAY, 10),
            (MAY, 6, MAY, 11),
            (JUN, 23, JUN, 29),
            (2001, DEC, 29, DEC, 31),
        ),
    ...
```
All comments and suggestions on naming, format and implementation are highly welcome (starting with name "substituted holidays" itself) :)

## Type of change

## Checklist


holidays/locale/ar/LC_MESSAGES/UA.po
```diff
@@ -3,95 +3,101 @@
...
 "Plural-Forms: nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 && n%100<=99 ? 4 : 5);\n"
 "Generated-By: Lingua 4.15.0\n"
-"X-Generator: Poedit 3.3.1\n"
+"X-Generator: Poedit 3.2.2\n"
+
+#. Day off (substituted from {day}.{month}).
+#: ./holidays/countries/ukraine.py:74
+#, python-format
+msgid "Вихідний день (перенесено з {day:02}.{month:02})"
+msgstr "يوم عطلة (استبدل من {day:02}/{month:02})"
```
KJhellico, line 23: @abh31000 I am not sure if this translation and the chosen date format are correct, so I would be happy to hear your opinion.

tests/countries/test_ukraine.py
```diff
@@ -316,97 +416,83 @@ def test_2022(self):
...
+            ("2021-12-27", "Різдво Христове (за григоріанським календарем) (вихідний)"),
         )
 
     def test_l10n_en_us(self):
         self.assertLocalizedHolidays(
             "en_US",
             ("2021-01-01", "New Year's Day"),
             ("2021-01-07", "Christmas (Julian calendar)"),
+            ("2021-01-08", "Day off (substituted from 1/16)"),
             ("2021-03-08", "International Women's Day"),
```
arkid15r, line 455: Not sure if it's valid w/o a year.

arkid15r (commented):
The base idea implementation looks great! I suggested a couple of aspects to think through:

holidays/holiday_base.py
```diff
@@ -557,6 +560,18 @@ def _add_subdiv_holidays(self):
             if add_subdiv_holidays and callable(add_subdiv_holidays):
                 add_subdiv_holidays()
 
+    def _add_substituted_holidays(self):
+        """Populate substituted holidays."""
+        substituted_label = getattr(
```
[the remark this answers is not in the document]
KJhellico: Hmm, that makes sense. But which exception is more applicable here?

holidays/locale/ar/LC_MESSAGES/UA.po
abh31000:
```suggestion
msgstr "يوم عطلة (استبدل من {يوم:02}.{شهر:02})"
```
Yeah like this it's correct. I also think "day" and "month" should be translated

KJhellico: No, here "day" and "month" is variable substitutions for` str.format()`:
```python
self._add_holiday(
    self.tr(substituted_label).format(day=from_day, month=from_month), to_month, to_day
)
```
And it looks like I'm going to replace it. :)

abh31000: Ah my bad ! :sleeping: 
Both the translation and date format are correct.

KJhellico: Now date format and substituted holiday name is set separately.

holidays/holiday_base.py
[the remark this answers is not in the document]
arkid15r: It could be just a ValueError until we migrate to custom exceptions approach.

arkid15r (commented):
That's a PoC we could start using 👍

PR #1500, opened 2023-10-05
Remove Columbus Day as state holiday in California

hissingpanda:
## Proposed change

Columbus day is no longer a state holiday in CA: https://www.sos.ca.gov/state-holidays

## Type of change

## Checklist

### DIFF
--- a/holidays/countries/united_states.py
+++ b/holidays/countries/united_states.py
@@ -186,6 +186,7 @@ class UnitedStates(ObservedHolidayBase, ChristianHolidays, InternationalHolidays
         if self._year >= 1937 and self.subdiv not in {
             "AK",
             "AR",
+            "CA",
             "DE",
             "FL",
             "HI",
--- a/tests/countries/test_united_states.py
+++ b/tests/countries/test_united_states.py
@@ -447,7 +447,7 @@ class TestUS(TestCase):
         )
         self.assertHolidayName(name, dt)
 
-        subdiv_dont = {"AK", "AR", "DE", "FL", "HI", "NV", "SD", "VI"}
+        subdiv_dont = {"AK", "AR", "CA", "DE", "FL", "HI", "NV", "SD", "VI"}
         for subdiv in set(UnitedStates.subdivisions) - subdiv_dont:
             self.assertHolidayName(name, self.state_hols[subdiv], dt)
 
@@ -470,7 +470,7 @@ class TestUS(TestCase):
             "2023-10-09",
         )
 
-        for subdiv in ("AK", "AR", "DE", "FL", "HI", "NV"):
+        for subdiv in ("AK", "AR", "CA", "DE", "FL", "HI", "NV"):
             self.assertNoHoliday(self.state_hols[subdiv], dt)
             self.assertNoHolidayName(name, self.state_hols[subdiv])
 


PPsyrius:
From a quick check, looks like [Delaware (DE)](https://www.pewresearch.org/short-reads/2023/10/05/working-on-columbus-day-or-indigenous-peoples-day-it-depends-on-where-your-job-is/) shouldn't have this too:
> Since then, several states have moved away from Columbus Day. California and Delaware dropped the holiday entirely in 2009, the latter swapping in a floating holiday for state workers. Maine, New Mexico, Vermont and D.C. all renamed the day Indigenous Peoples’ Day in 2019, while retaining it as an official holiday.

hissingpanda:
Looks like Delaware is already excluded from it. just pushed some updates to exclude CA from the US columbus day tests

KJhellico (dismissed):
PR is LGTM.  However, according to given sources, Columbus Day is no longer observed in a significant number of states, not just California. I will try to implement this in a separate PR.

arkid15r:
>PR is LGTM.

I'm not sure if the implementation is correct. Shouldn't we add a branching for the point in time when the holiday was removed and preserve history for previous years?

holidays/countries/united_states.py
```diff
@@ -186,6 +186,7 @@ def _add_subdiv_holidays(self):
         if self._year >= 1937 and self.subdiv not in {
             "AK",
             "AR",
+            "CA",
```
arkid15r, line 189: I read it as like the Columbus day holiday never existed. As far as I remember it was just renamed somewhere around 2019th. Please confirm with refs to some sources if possible.

arkid15r (commented):
Hi @hissingpanda, thanks for the PR!

The change you're proposing needs some clarification. Please share the information you used for making your decision to remove the holiday.

Thank you!

holidays/countries/united_states.py
PPsyrius: They were removed altogether in 2009 for `CA` and `DE`.

`DC`, `ME`, `NM`, and `VT` on the other hand kept it with a name change to "Indigenous Peoples' Day" in 2019.

Though if we're at it, may as well go deep down and check other cases as well like Hawaii's names change to "Discoverers Day" sometime prior to 2000 and other instances just to be sure. 👀

arkid15r: > They were removed altogether in 2009 for `CA` and `DE`.

Any sources?

Also https://www.gov.ca.gov/2019/10/14/governor-newsom-issues-proclamation-declaring-indigenous-peoples-day/

PPsyrius: https://www.pewresearch.org/short-reads/2023/10/05/working-on-columbus-day-or-indigenous-peoples-day-it-depends-on-where-your-job-is/

AFAIK California's Indigenous Peoples' Day declared in 2019 is an observed-only (no public holidays are given) https://en.m.wikipedia.org/wiki/Indigenous_Peoples%27_Day_(United_States)

hissingpanda: you're right it should be only removed since 2009 not before so should update the PR for that. Here are some news article sources at least and the state holiday calendar. Native American Day on the fourth friday of september was added in lieu of it but also seems to be an optional day off that only some state institutions observe.

https://www.latimes.com/nation/nationnow/la-na-nn-columbus-day-why-do-we-still-celebrate-20151012-htmlstory.html

> (Although California Gov. Arnold Schwarzenegger eliminated the Columbus Day state holiday as part of a budget-cutting measure in 2009, Los Angeles city and county offices still observe it. The Los Angeles Unified School District does not.)

https://www.pewresearch.org/short-reads/2023/10/05/working-on-columbus-day-or-indigenous-peoples-day-it-depends-on-where-your-job-is/

> California, for example, is among the 26 states that don’t have an official public holiday on the second Monday in October. But in the past, it has designated the day as [Indigenous Peoples’ Day by gubernatorial proclamation](https://www.gov.ca.gov/2022/10/10/governor-newsom-proclaims-indigenous-peoples-day-2022/). California law also designates the fourth Friday in September as Native American Day, which state employees may take in lieu of their annual personal holiday.

And here for the current year and previous you can see it's not on the state calendar: https://www.calhr.ca.gov/employees/pages/state-holidays.aspx

KJhellico:
@hissingpanda, thank you for raising the issue of this nuanced holiday and providing valuable sources. I've made [complex PR](https://github.com/vacanza/python-holidays/pull/1503) that addresses this more completely, not just for CA, so I'm going to close this PR. Please feel free to submit any questions or comments in #1503.

PR #1521, opened 2023-10-18
Use setdefault instead of hard-coding observed_rule and observed_since

BossColo:
## Proposed change

Instead of calling super().init() from countries with hard-coded values for override_rule and override_since, use kwargs.setdefaults() to add a value if one isn't passed in. This allows for users to set their own override_rule and override_since if desired.

## Type of change

## Checklist

### DIFF
--- a/holidays/countries/albania.py
+++ b/holidays/countries/albania.py
@@ -36,7 +36,8 @@ class Albania(
         InternationalHolidays.__init__(self)
         IslamicHolidays.__init__(self)
         StaticHolidays.__init__(self, AlbaniaStaticHolidays)
-        super().__init__(observed_rule=SAT_SUN_TO_NEXT_WORKDAY, *args, **kwargs)
+        kwargs.setdefault("observed_rule", SAT_SUN_TO_NEXT_WORKDAY)
+        super().__init__(*args, **kwargs)
 
     def _populate(self, year):
         super()._populate(year)
--- a/holidays/countries/angola.py
+++ b/holidays/countries/angola.py
@@ -50,7 +50,8 @@ class Angola(ObservedHolidayBase, ChristianHolidays, InternationalHolidays, Stat
         ChristianHolidays.__init__(self)
         InternationalHolidays.__init__(self)
         StaticHolidays.__init__(self, cls=AngolaStaticHolidays)
-        super().__init__(observed_rule=TUE_TO_PREV_MON + THU_TO_NEXT_FRI, *args, **kwargs)
+        kwargs.setdefault("observed_rule", TUE_TO_PREV_MON + THU_TO_NEXT_FRI)
+        super().__init__(*args, **kwargs)
 
     def _is_observed(self, dt: date) -> bool:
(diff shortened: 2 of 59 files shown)


BossColo:
Sorry about the force pushes. Didn't realize what I was doing when I clicked the Update branch button. Been a while since I used GitHub.

BossColo:
Rebased to resolve merge conflicts.

arkid15r:
Hi @BossColo, 
thank you for the PR!

The implementation is clean and pretty straightforward, I have a question about the motivation though: do you have an example when a user would need to override those values? 

I agree, it may be great to have the proposed `kwargs.setdefault` approach implemented from the user perspective. 
The point was in keeping PH a source of truth for holiday dates (including observances). This type of flexibility might discourage people submitting fixes for incorrect observance rules -- why bother when you can just override it, right?

Anyway, I'd like to hear more opinions here as I'm not convinced this is beneficial in a long term run. 

@KJhellico, @PPsyrius could you also look into this?

Thank you!

BossColo:
> Hi @BossColo, thank you for the PR!
> 
> The implementation is clean and pretty straightforward, I have a question about the motivation though: do you have an example when a user would need to override those values?
> 
> I agree, it may be great to have the proposed `kwargs.setdefault` approach implemented from the user perspective. The point was in keeping PH a source of truth for holiday dates (including observances). This type of flexibility might discourage people submitting fixes for incorrect observance rules -- why bother when you can just override it, right?
> 
> Anyway, I'd like to hear more opinions here as I'm not convinced this is beneficial in a long term run.
> 
> @KJhellico, @PPsyrius could you also look into this?
> 
> Thank you!

My personal use case stems from the fact that I work for a Canadian company with a worldwide footprint. The company has decided that for all jurisdictions, we will use the SAT_SUN_TO_NEXT_MON rule, regardless of the country in question. So if I want to use this library for US holidays, I have to override the __init__ method for the US class.

arkid15r:
> My personal use case stems from the fact that I work for a Canadian company with a worldwide footprint. The company has decided that for all jurisdictions, we will use the SAT_SUN_TO_NEXT_MON rule, regardless of the country in question. So if I want to use this library for US holidays, I have to override the __init__ method for the US class.

Thanks for providing more context, @BossColo. It may be weird but totally legit use case. Apparently, sometimes the library doesn't have to be the source of truth for holidays data.

Thank you for the clean PR and your first contribution!

arkid15r (approved):

BossColo:
> > My personal use case stems from the fact that I work for a Canadian company with a worldwide footprint. The company has decided that for all jurisdictions, we will use the SAT_SUN_TO_NEXT_MON rule, regardless of the country in question. So if I want to use this library for US holidays, I have to override the **init** method for the US class.
> 
> Thanks for providing more context, @BossColo. It may be weird but totally legit use case. Apparently, sometimes the library doesn't have to be the source of truth for holidays data.
> 
> Thank you for the clean PR and your first contribution!

Thank you for the kind words, and for the fantastic library!

PR #1548, opened 2023-11-10
Migrate NYSE holidays to ObservedHolidayBase

KJhellico:
## Proposed change

Migrate NYSE holidays to `ObservedHolidayBase`.

## Type of change

## Checklist

### DIFF
--- a/holidays/financial/ny_stock_exchange.py
+++ b/holidays/financial/ny_stock_exchange.py
@@ -39,18 +39,12 @@ class NewYorkStockExchange(
         kwargs.setdefault("observed_rule", SAT_TO_PREV_FRI + SUN_TO_NEXT_MON)
         super().__init__(*args, **kwargs)
 
-    def _add_observed_holiday(self, name, *args) -> None:
-        dt = date(self._year, *args)
-        is_obs, _ = self._add_observed(dt, name)
-        if not is_obs:
-            self._add_holiday(name, dt)
-
     def _populate(self, year):
         super()._populate(year)
 
-        # New Year Day.
+        # New Year's Day.
         name = "New Year's Day"
-        self._add_observed_holiday("New Year's Day", JAN, 1)
+        self._move_holiday(self._add_new_years_day(name))
         self._add_observed(self._next_year_new_years_day, name)
 
         # MLK, 3rd Monday of January.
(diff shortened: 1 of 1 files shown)


holidays/financial/ny_stock_exchange.py
```diff
@@ -43,53 +30,43 @@ class NewYorkStockExchange(HolidayBase, ChristianHolidays, InternationalHolidays
...
-        elif self._is_sunday(dt):
-            dt += td(days=+1)
-            name = f"{name} (Observed)"
-
-        self._add_holiday(name, dt)
+    def _add_observed_holiday(self, name, *args) -> None:
+        dt = date(self._year, *args)
+        is_obs, _ = self._add_observed(dt, name)
+        if not is_obs:
+            self._add_holiday(name, dt)
```
arkid15r, line 46: It didn't seem right to have two similar `_add_observed...` method. Could you check if this suggestion covers your needs:
```suggestion
    def _add_observed(self, name, *dt) -> None:  # type: ignore[override]
        dt = dt if len(dt) > 1 else dt[0]
        is_observed, _ = super()._add_observed(dt, name)  # type: ignore[arg-type, assignment]
        if not is_observed:
            self._add_holiday(name, dt)  # type: ignore[assignment]

```

You'll need to replace all `_add_observed_holiday` cases to `_add_observed`. Please note the args order: name, then date args

holidays/financial/ny_stock_exchange.py
```diff
@@ -43,53 +30,43 @@ class NewYorkStockExchange(HolidayBase, ChristianHolidays, InternationalHolidays
...
-
-        # https://www.nyse.com/publicdocs/nyse/regulation/nyse/NYSE_Rules.pdf
-        # As per Rule 7.2.: check if next year's NYD falls on Saturday and
-        # needs to be observed on Friday (Dec 31 of previous year).
-        dec_31 = date(year, DEC, 31)
-        if self._is_friday(dec_31):
-            self._add_observed_holiday("New Year's Day", dec_31 + td(days=+1))
+        name = "New Year's Day"
+        self._add_observed_holiday("New Year's Day", JAN, 1)
+        self._add_observed(self._next_year_new_years_day, name)
```
arkid15r, line 54:
```suggestion
        self._add_observed(name, self._next_year_new_years_day)
```

arkid15r (commented):
It seems the `ObservedHolidayBase::_add_observed` needs to be overridden here:

holidays/financial/ny_stock_exchange.py
KJhellico: We have `_move_holiday()` for such cases, I forgot about it. :)

arkid15r: That's a lot better!

arkid15r (approved):

PR #1560, opened 2023-11-17
Migrate South Korea holidays to ObservedHolidayBase

KJhellico:
## Proposed change

Migrate South Korea holidays to `ObservedHolidayBase`.

## Type of change

## Checklist


PPsyrius:
Nice to see this finally getting done ❤️

holidays/countries/south_korea.py
```diff
@@ -87,66 +77,46 @@ def __init__(self, *args, **kwargs):
...
-            obs_date = dt + td(days=+1)
-            while obs_date.weekday() in target_weekday or obs_date in self:
-                obs_date += td(days=+1)
-            for name in (name,) if name else self.get_list(dt):
-                if "Alternative holiday" not in name:
+    def _populate_observed(self, dts: Set[date], spec_holidays: Dict[date, str]) -> None:
+        for dt in sorted(dts.union(spec_holidays.keys())):
+            if not self._is_observed(dt):
+                continue
+            obs_date = self._get_observed_date(
```
arkid15r, line 88:
```suggestion
            dt_observed = self._get_observed_date(
```

holidays/countries/south_korea.py
```diff
@@ -87,66 +77,46 @@ def __init__(self, *args, **kwargs):
...
-        target_weekday = {SUN}
-        if include_sat:
-            target_weekday.add(SAT)
-        if (dt.weekday() in target_weekday or len(self.get_list(dt)) > 1) and dt.year >= since:
-            obs_date = dt + td(days=+1)
-            while obs_date.weekday() in target_weekday or obs_date in self:
-                obs_date += td(days=+1)
-            for name in (name,) if name else self.get_list(dt):
-                if "Alternative holiday" not in name:
+    def _populate_observed(self, dts: Set[date], spec_holidays: Dict[date, str]) -> None:
```
arkid15r, line 84: Let's get rid of the ambiguity:
```suggestion
    def _populate_observed(self, dts: Set[date], three_day_holidays: Dict[date, str]) -> None:
```

holidays/countries/south_korea.py
```diff
@@ -87,66 +77,46 @@ def __init__(self, *args, **kwargs):
...
-            # The day preceding %s.
-            self._add_holiday(self.tr("%s 전날") % self.tr(name), dt + td(days=-1)),
-            dt,
-            # The second day of %s.
-            self._add_holiday(self.tr("%s 다음날") % self.tr(name), dt + td(days=+1)),
-        ):
-            self._add_alt_holiday(dt_alt, name=name, include_sat=False)  # type: ignore[arg-type]
-
     def _populate_public_holidays(self):
+        def _append_observed(dt: date, since: int):
```
arkid15r, line 99: As this is a nested function:
```suggestion
        def append_observed(dt: date, since: int):
```

holidays/countries/south_korea.py
```diff
@@ -87,66 +77,46 @@ def __init__(self, *args, **kwargs):
...
+            if self._year >= since:
+                dts_observed.add(dt)
+
+        def _add_three_day_holiday(dt: date, name: str):
```
arkid15r, line 103: The leading underscore naming would make sense if it was `SouthKorea::_add_three_day_holiday` static method. I guess it's okay to move it there. 

Maybe it'd be a good idea not to use nested functions for such cases (or use nested functions only, the idea is about consistency, you know)

arkid15r (dismissed):
🇰🇷 LGTM

Some optional naming comments:

holidays/countries/south_korea.py
KJhellico: I made them nested functions for easy access to local `dts_observed` and `three_days_holidays`.

arkid15r: Yeah, I understand your reasons. My point was about general approach -- keep using nested functions with local variables  or use or class methods + attributes as an alternative.

KJhellico: I think class methods is too "heavy" for such small things. :)

arkid15r: It may be, I don't have a strong opinion on this yet. We should discuss it some time this year.

arkid15r (approved):

PR #1622, opened 2024-01-02
Update Uzbekistan holidays, add l10n support

KJhellico:
## Proposed change

Update Uzbekistan holidays, add l10n support (en_US, uk, **uz**).

## Type of change

## Checklist


holidays/countries/uzbekistan.py
```diff
@@ -9,51 +9,87 @@
...
+        - `Labor Code 21.12.1995 <https://lex.uz/docs/-142859>`_
+        - `Labor Code 28.10.2022 <https://lex.uz/docs/-6257288>`_
     """
 
     country = "UZ"
+    default_language = "uz"
+    # Estimated label.
+    estimated_label = tr("%s* (*taxminiy)")
+    # %s (Observed).
+    observed_label = tr("%s (ko‘chirilgan)")
```
arkid15r, line 34: Shouldn't we use ' instead of ‘ here?

holidays/countries/uzbekistan.py
```diff
@@ -9,51 +9,87 @@
...
+    References:
+        - https://en.wikipedia.org/wiki/Public_holidays_in_Uzbekistan
+        - `Labor Code 21.12.1995 <https://lex.uz/docs/-142859>`_
+        - `Labor Code 28.10.2022 <https://lex.uz/docs/-6257288>`_
     """
 
     country = "UZ"
+    default_language = "uz"
+    # Estimated label.
+    estimated_label = tr("%s* (*taxminiy)")
```
arkid15r, line 32: * need to be removed to align with the recently submitted changes

arkid15r (commented):
Looks good!

holidays/countries/uzbekistan.py
KJhellico: According to [Wiki](https://en.wikipedia.org/wiki/Uzbek_alphabet#Distinct_characters), the most correct symbol to use is `ʻ` (U+02BB), but official sites (including https://lex.uz) use `‘` (U+2018). That's why I used U+2018 as well.

arkid15r: You're right, this is the correct one.

Issue #284, opened 2020-02-02
Use recognised subdivisions for Austria


jbroudou:
As per #283, is is possible to use standard ISO 3166-2 codes for Austria?

Per wikipedia:
AT-1  Burgenland Burgenland
AT-2  Kärnten Carinthia
AT-3  Niederösterreich Lower Austria
AT-4  Oberösterreich Upper Austria
AT-5  Salzburg Salzburg
AT-6  Steiermark Styria
AT-7  Tirol Tyrol
AT-8  Vorarlberg Vorarlberg
AT-9  Wien Vienna

Per python-holidays - the codes for AT subdivisions are: 'B', 'K', 'N', 'O', 'S', 'ST', 'T', 'V', 'W' (W is the default) which appears to be the first letter of the subdivision name (see wikipedia list).

dr-prodigy:
Hi, released in beta, thank you!

dr-prodigy:
Hi,
in order to maintain consistency with other countries behaviour, these codes have been chosen instead:
1, 2, 3, 4, 5, 6, 7, 8, 9 (default)

PR #1656, opened 2024-01-23
Use common abbreviations for Austrian states.

sphh:
## Proposed change

Commit
https://github.com/vacanza/python-holidays/commit/4342d28ff0f6d6a63aff433c72b80d33589991f7 uses the numbers 1-9 for the nine states in Austria, but actually no one in Austria uses these numbers. This commit uses the common character codes (B, K, N, O, S, St, T, V, W) with the same default (9 and W) and adds the numbers as deprecated subdivisions. (See also #284)

Comment https://github.com/vacanza/python-holidays/issues/284#issuecomment-584852192 reasons, that the numbers were used to keep it consistent with the other countries. Looking at https://python-holidays.readthedocs.io/en/latest/index.html#available-countries numbers are nowadays very rarely used for subdivisions.

It is expected for the impact to be negligible, because none of these subdivisions actually does have an effect on the list of holidays.

## Type of change

## Checklist

- [ ] I've run `make pre-commit`, it didn't generate any changes
- [ ] I've run `make test`, all tests passed locally

### DIFF
--- a/holidays/countries/austria.py
+++ b/holidays/countries/austria.py
@@ -4,7 +4,8 @@
 #  specific sets of holidays on the fly. It aims to make determining whether a
 #  specific date is a holiday as fast and flexible as possible.
 #
-#  Authors: dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
+#  Authors: Stephan Helma <s.p.helma@gmx.net> (c) 2024
+#           dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
 #           ryanss <ryanssdev@icloud.com> (c) 2014-2017
 #  Website: https://github.com/dr-prodigy/python-holidays
 #  License: MIT (see LICENSE file)
@@ -21,7 +22,8 @@ class Austria(HolidayBase, ChristianHolidays, InternationalHolidays):
     default_language = "de"
     supported_categories = (BANK, PUBLIC)
     supported_languages = ("de", "en_US", "uk")
-    subdivisions = ("1", "2", "3", "4", "5", "6", "7", "8", "9")
+    subdivisions = ("B", "K", "N", "O", "S", "St", "T", "V", "W")
+    _deprecated_subdivisions = ("1", "2", "3", "4", "5", "6", "7", "8", "9")
 
     def __init__(self, *args, **kwargs) -> None:
         ChristianHolidays.__init__(self)
@@ -29,7 +31,7 @@ class Austria(HolidayBase, ChristianHolidays, InternationalHolidays):
 
         # Set the default subdivision.
         if not kwargs.get("subdiv", kwargs.get("state")):
-            kwargs["subdiv"] = "9"
+            kwargs["subdiv"] = "W"
 
         super().__init__(*args, **kwargs)
 


KJhellico:
These (digital) codes follow [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2:AT) standard.

arkid15r:
> These (digital) codes follow [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2:AT) standard.

👆 this would be the main reason to avoid merging this PR. As our goal is to use a standardized approach for all entities the ISO3166-2 codes fit better here even though nobody is used to it in AT.

@sphh just a note for possible future contributions: you should have run the tests as changing the subdivisions isn't enough. The [README](https://github.com/vacanza/python-holidays/blob/1cbd6c2f4fec00c41c8856aeb3831386d86e0419/README.rst?plain=1#L223) file contains per country documentation and needs to be updated as well.


@KJhellico while reviewing this PR I realized these subdivisions don't make much sense here as `AT` atm doesn't have any subdivision specific holidays implemented, right? We don't necessarily need to remove them now, just an improvement idea for v1.

sphh:
Absolutely nobody would use 9 for Vienna. And even worse, nobody would know, that 9 stands for Vienna. The number can be found absolutely nowhere: Not on license plates (at least the old ones used the character specified in my PR), not in area codes for phone number, not in zip codes, not anywhere.

No idea who came up with these codes. Probably no one cared in Austria and whoever wrote the ISO codes simply numbered them alphabetically because they had no idea what is used in Austria. But that is just a wild guess.

So it is ISO codes vs. practicability! What should be preferred?

An elegant way out of this would be to have aliases, which are not deprecated. Is that something you would accept? This could also be used to specify the subregion with the full name. This would fit well into the `holidays` approach, where the holidays for Austria can be get with ‘AT’, ‘AUT’ and ‘Austria’.

Re the other points: Well, before doing all the other work, I wanted to know, if it were to be accepted at all.

sphh:
And yes, there are different holidays for different states in Austria: The day of the local patron saints are school holidays, but these were not implemented in the `holidays` package.

arkid15r:
> So it is ISO codes vs. practicability! What should be preferred?
> 
> An elegant way out of this would be to have aliases, which are not deprecated. Is that something you would accept? This could also be used to specify the subregion with the full name. This would fit well into the `holidays` approach, where the holidays for Austria can be get with ‘AT’, ‘AUT’ and ‘Austria’.

I'd pick both if possible. If only one option is available I'd prefer the standardization (codes).

I like the idea of aliases, assuming that AT might not be the only case. This kind of logic needs to be implemented on the `HolidayBase` level to support any entity. I'll create a ticket

sphh:
> I'll create a ticket

Great! So I close this PR, because it did its part.

arkid15r:
@sphh I created https://github.com/vacanza/python-holidays/issues/1660. Is it something you would be comfortable to work on or rather leave it to PH core team to implement?

Once the subdivision alias support is there AT's mapping shouldn't be hard to add. Another point is that currently PH has not support of per subdivision holidays for AT 🤷‍♂️

sphh:
I can have a go, if you like. Three different approaches come do mind (all without in-depth investigation): Using a dictionary, where the key is the official ISO code and the value a tuple of all aliases. Secondly, use a tuple with a tuple for each subregion, where the first element is the official ISO code and all other elements aliases. And finally keep the subregion tuple as it is and have a separate dictionary with the aliases. Any preferences?

arkid15r:
All right, at this is no longer the PR related let's move to https://github.com/vacanza/python-holidays/issues/1660
Could you copy/paste your comment there?

Thank you!

PR #1664, opened 2024-01-24
Add Jersey holidays

PPsyrius:
## Proposed change

Add Jersey holidays.

Closes #1647.

## Type of change

- [x] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new python-holidays functionality in general)

## Checklist

- [x] I've followed the [contributing guidelines][contributing-guidelines]
- [x] I've run `make pre-commit`, it didn't generate any changes
- [x] I've run `make test`, all tests passed locally

[contributing-guidelines]: https://github.com/vacanza/python-holidays/blob/beta/CONTRIBUTING.rst
[docs]: https://github.com/vacanza/python-holidays/tree/beta/docs/source

PR #1668, opened 2024-01-26
Add United Kingdom's Whit Monday holidays for pre-1971 years

PPsyrius:
## Proposed change

As spotted in #1664, this implements the following changes for United Kingdom holidays:
- Implements __Whit Monday__ holidays from 1871 onwards until its eventual replacement by __Spring bank holidays__ in 1971 for __England__ (`ENG`), __Wales__ (`WLS`), and __North Ireland__ (`NIR`).
- Add a cut-off date; thus, bank holidays are no longer available before the implementation of the __Bank Holidays Act 1871__.
- In-Lieu observance only took effect from 1875 onwards as decreed in the __Bank Holidays Extension Act 1875__.

## Type of change

## Checklist


holidays/countries/united_kingdom.py
```diff
@@ -53,7 +60,15 @@ def __init__(self, *args, **kwargs):
         kwargs.setdefault("observed_rule", SAT_SUN_TO_NEXT_MON)
         super().__init__(*args, **kwargs)
 
+    def _is_observed(self, dt: date) -> bool:
+        # Bank Holidays Extension Act 1875
+        return self._year >= 1875
+
```
KJhellico, line 66:
```suggestion
        # Bank Holidays Extension Act 1875
        kwargs.setdefault("observed_since", 1875)
        super().__init__(*args, **kwargs)

```

holidays/countries/united_kingdom.py
```diff
@@ -109,6 +128,10 @@ def _populate_subdiv_eng_public_holidays(self):
         # Easter Monday
         self._add_easter_monday("Easter Monday")
 
+        # Whit Monday.
```
KJhellico, line 131: Better to move it to `_populate_subdiv_holidays()`, since there is already a group of holidays that are not exist in Scotland only. Perhaps even together with Easter Monday and Late Summer Bank Holiday, as they have identical rules in ENG, NIR and WLS.

KJhellico (commented):
An excellent point about Whit Monday absence! 👍

holidays/countries/united_kingdom.py
PPsyrius: That will affect Isle of Man too, in which I haven't found any evidence yet if they ever have this in their holiday list pre-1971 os that's something we need to keep in mind.

PPsyrius: The only evidence Isle of Man ever celebrates Whit Monday at all is this [British Pathe video](https://www.youtube.com/watch?app=desktop&v=JzdJ7v93A6M) - not exactly perfect but this will do for now. :+1:

KJhellico: Sorry, my fault - I forgot that this code is also working for whole UK now (without `subdiv`), so we can't place Whit Monday here (as there should be only holidays that exist in all provinces).

PPsyrius: After spending the whole morning looking for Manx pre-1971 calendar, I can now confirm that Isle of Man does indeed have Whit Monday as public holidays at one point (or at least for 1968): https://namanx.org/holy-days-a-hanging-buried-treasure-all-in-a-manx-calendar

image
image
image

KJhellico: You're really digging to deep, it's just fantastic! 👍 But I think we need to move Whit Monday from `_populate_subdiv_holidays()` to specific subdivisions (and the Isle of Man), as you did at first (because now we have Whit Monday in whole UK as well, which is incorrect). I'm sorry for misleading you with [that comment](https://github.com/vacanza/python-holidays/pull/1668#discussion_r1467545251).

PPsyrius: Fixed.

KJhellico (approved):
LGTM!

arkid15r (approved):
🇬🇧 🇮🇲  LGTM

PR #1673, opened 2024-01-27
Unify common holidays name translation

KJhellico:
## Proposed change

Unify common holidays name translation (step to switch to per-locale .po files):
- unify en_US translation of Labor Day, Christmas Day, Assumption Day, Easter Sunday etc.
- update comments

## Type of change

## Checklist


holidays/countries/egypt.py
```diff
@@ -58,15 +54,17 @@ def _populate_public_holidays(self):
             # National Police Day
             self._add_holiday_jan_25(tr("عيد الشرطة"))
 
-        # Coptic Easter - Orthodox Easter
+        # Coptic Easter
```
arkid15r, line 57: Could you add trailing dots to this file comments for consistency?

arkid15r (commented):
LGTM, great work!

Some thoughts after reviewing this:
  - for Islamic holidays we use both translated and "real" English names, e.g., `Eid al-Adha` and `Prophet's Birthday` (instead of Mawlid)
  - another example - for Mawlid we use `Prophet's Birthday` comment and next to it `Isra and Miraj` instead of `The Prophet's Ascension` (ID holidays)

It looks like something that can be unified too (now or later).

KJhellico:
Yes, for some Islamic holidays we use here "traditional" version of the name rather than translated one. Where Eid al-Fitr and Eid al-Adha can still be translated (as "Feast of Breaking the Fast" and "Feast of Sacrifice"), I can't find a suitable short name for Isra' and Mi'raj and Ashura, for example :)

arkid15r (approved):
LGTM

PR #1688, opened 2024-02-07
Update Taiwan holidays

KJhellico:
## Proposed change

Update Taiwan holidays (add substituted holidays according to official sources from 2010).

## Type of change

## Checklist


snapshots/countries/TW_COMMON.json
```diff
@@ -543,6 +543,9 @@
     "2010-02-14": "Chinese New Year",
     "2010-02-15": "Chinese New Year",
     "2010-02-16": "Chinese New Year",
+`    "2010-02-17": "Chinese New Year's Eve (observed)",
```
PPsyrius, line 546:
```suggestion
    "2010-02-17": "Chinese New Year's Eve (observed)",
```
rogue ` here

PPsyrius (changes requested):
Nice work finding all those exact holiday lists from the 2010s👍

PPsyrius:
LGTM 🇹🇼

PPsyrius (approved):
LGTM 🇹🇼

holidays/countries/taiwan.py
```diff
@@ -103,3 +124,72 @@ class TW(Taiwan):
 
 class TWN(Taiwan):
     pass
+
+
+class TaiwanStaticHolidays:
+    # Date format (see strftime() Format Codes).
+    substituted_date_format = tr("%Y-%m-%d")
+    # Day off (substituted from %s).
+    substituted_label = tr("休息日（%s日起取代）")
```
arkid15r, line 133: Is the trailing space intended here?

arkid15r (commented):
LGTM

holidays/countries/taiwan.py
PPsyrius: I don't think that's a trailing space but [Fullwidth Right Parenthesis](https://www.compart.com/en/unicode/U+FF09) instead of the usual [Right Parenthesis](https://www.compart.com/en/unicode/U+0029) - something we should look into later for project-wide standardisation whether to keep or get rid of them.

PPsyrius: Update: I did look up the proper standard, and w3c sticks with fullwidth parenthesises for CJK languages

https://github.com/w3c/clreq/issues/74
https://japanese.stackexchange.com/questions/18855/when-mixing-japanese-and-english-should-i-use-japanese-parentheses

KJhellico: Updated, now Fullwidth Parentheses are used everywhere for TW. Thank you for suggestion!

arkid15r: All right, this seems to be resolved!

arkid15r (approved):

PR #1762, opened 2024-04-13
Extend HolidayBase::_add_holiday family methods with Easter support

KJhellico:
## Proposed change

Add `_add_holiday_N_days_past_easter()` and `_add_holiday_N_days_prior_easter()` patterns support.

## Type of change

## Checklist


holidays/holiday_base.py
```diff
@@ -483,6 +483,24 @@ def __getattr__(self, name):
                     + timedelta(days=+int(days) if delta_direction == "past" else -int(days)),
                 )
 
+            # Handle <n> days <past/prior> easter patterns (e.g.,
+            # _add_holiday_8_days_past_easter() or
+            # _add_holiday_5_days_prior_easter()).
+            nth_weekday_of_month_with_delta = re.match(
+                r"_add_holiday_(\d{1,2})_days_(past|prior)_easter",
```
arkid15r, line 490: Let's keep it compatible with the rest:
```suggestion
                r"_add_holiday_(\d{1,2})_days?_(past|prior)_easter",
```

holidays/holiday_base.py
```diff
@@ -483,6 +483,24 @@ def __getattr__(self, name):
...
+                name,
+            )
+            if nth_weekday_of_month_with_delta:
+                (
+                    days,
+                    delta_direction,
+                ) = nth_weekday_of_month_with_delta.groups()
+                return lambda name: self._add_holiday(
+                    name,
+                    self._easter_sunday
```
arkid15r, line 500: Do we want to explicitly handle cases when the method is invoked for an entity with no `ChristianHolidays` inheritance?

holidays/holiday_base.py
arkid15r, line 1: Could you update [the tests](https://github.com/vacanza/python-holidays/blob/dd6f99240ba9a4b9bc218040b33096c3d4f64c85/tests/test_holiday_base.py#L917)?

arkid15r (commented):
I had started reviewing the code and after getting to L500 of `holiday_base.py` I realized this case isn't library wide and works for `ChristmasHolidays` based entities only. Do we want to put another `__getattr__` check for a limited number of countries? 

I like the more readable code, however in this case it's not an obvious benefit due to the performance tradeoff and limited use scenario.

holidays/holiday_base.py
KJhellico: Yes, it is used in not many countries (currently, 14). But pattern `"<past/prior> <last/<nth> <weekday> of <month>"` is used by 11. :)

KJhellico: Let's place it as the last pattern.

holidays/holiday_base.py
KJhellico: Since this is a feature for internal use, let's assume a thoughtful use. Likewise, it's possible to invoke something like `_add_holiday_apr_31()` or `_add_holiday_5_mon_of_feb()` (in non-leap year), or just make a typo and write `_add_holiday_yan_25()`, but we don't check it.

arkid15r (approved):
LGTM 👍

holidays/holiday_base.py
KJhellico: Done.

PR #1764, opened 2024-04-17
Refactor `HolidayBase::__getattr__`: speed up _add_holiday_* methods

KJhellico:
## Proposed change

Refactor `HolidayBase::__getattr__`: speed up \_add_holiday_* methods (remove regex use).

## Type of change

## Checklist

### DIFF
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -13,7 +13,6 @@
 __all__ = ("DateLike", "HolidayBase", "HolidaySum")
 
 import copy
-import re
 import warnings
 from calendar import isleap
 from datetime import date, datetime, timedelta, timezone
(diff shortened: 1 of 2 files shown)


holidays/holiday_base.py
```diff
@@ -427,95 +426,110 @@ def __getattr__(self, name):
...
-            month_day = re.match(r"_add_holiday_(\w{3})_(\d{1,2})", name)
-            if month_day:
-                month, day = month_day.groups()
-                return lambda name: self._add_holiday(
-                    name,
-                    date(self._year, getattr(gregorian, month.upper()), int(day)),
-                )
+            if len(tokens) == 5:
+                month, day = tokens[3:5]
+                if len(month) == 3 and month.isalpha() and len(day) < 3 and day.isdigit():
```
arkid15r, line 438:
```suggestion
                *_, month, day = tokens
                if month in MONTHS and day in DAYS:
```


And `calendars::gregorian` should contain 

```
DAYS = set(str(d) for d in range(1, 32))
MONTHS = set(
    m.lower()
    for m in ("JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT", "NOV", "DEC")
)
WEEKDAYS = set(wd.lower() for wd in ("MON", "TUE", "WED", "THU", "FRI", "SAT", "SUN"))

```

arkid15r (commented):
The performance improvement is great, let's tighten up the validation part as well. I suggested an idea on how we could narrow down the token parts values. Let me know If you have a better option in mind.

holidays/holiday_base.py
KJhellico: I tried to implement this. I made `MONTHS` and `WEEKDAYS` as mappings, it should be faster to get numbers than `getattr(gregorian, ...)`.

arkid15r (commented):
LGTM

arkid15r (approved):
🚀 LGTM 👍

PR #1782, opened 2024-05-01
Address numpy int argument conversion issue

### LINKED ISSUE #1779: The calculation of the lunar start date throws a TypeError exception for Thailand and Cambodia
version: 0.47
stack trace:
"python39\lib\site-packages\holidays\calendars\thai.py", line 233, in _get_start_date
    return _ThaiLunisolar.START_DATE + td(days=delta_days)
TypeError: unsupported type for timedelta days component: numpy.int32

arkid15r:
## Proposed change

Resolves #1779 

This change fixes `years` numpy int incompatibility issue. In order to test it properly we'll need `numpy` package installed in the test environment. I'm still not sure if that's a great idea as it's not really 💯 PH issue. On the other hand Prophet is a huge part of our ecosystem and it'd be very nice of us to take care of the issue.

FYI: there is a problem when running the new test on Windows pypy3.10 so I skipped it with `skipif`. It looks like `import numpy` causes that. If anyone knows how to fix that please let me know.

I also added pypy3.9 pypy3.10 to tox.ini env list.

@KJhellico @PPsyrius I'd like to know your opinion on this PR -- I'm open to removing the test or discarding the PR entirely based on your suggestions.

Thank you!

## Type of change

## Checklist


PPsyrius:
As long as `numpy` are only used for testing I have no qualm against that.

tests/third_party/test_numpy.py
```diff
@@ -0,0 +1,42 @@
...
+from holidays.countries.ukraine import Ukraine
+
+
+class TestNumpy(TestCase):
+    @pytest.mark.skipif(
+        platform().startswith("Windows")
+        and python_implementation().startswith("PyPy")
+        and python_version().startswith("3.10"),
+        reason='Avoid "make: *** [Makefile:63: test] Error 5" for pypy3.10 on Windows',
+    )
```
KJhellico, line 29: Without this skip, test also [ran without errors](https://github.com/KJhellico/python-holidays/actions/runs/8928197300/job/24523239200) for me.

arkid15r: It's not there yet :)

https://github.com/KJhellico/python-holidays/actions/runs/8928197300/job/24523240416

KJhellico: Yes, seems I was looking in the wrong place. :(

KJhellico (approved):
LGTM.

PR #1785, opened 2024-05-04
Replace `datetime::timedelta` with custom function

KJhellico:
## Proposed change

Replace `datetime::timedelta` with simple custom function for adding and subtracting days.

Performance test script:

```python
from datetime import date, timedelta
import timeit

def _delta_days(dt: date, n: int) -> date:
    return date.fromordinal(dt.toordinal() + n)

def with_timedelta():
    dt = date(2024, 1, 1)
    d2 = dt + timedelta(days=10)

def with_delta_days():
    dt = date(2024, 1, 1)
    d2 = _delta_days(dt, 10)

NUM = 10_000_000

t1 = timeit.timeit(with_timedelta, number=NUM)
t2 = timeit.timeit(with_add_days, number=NUM)

print(f"{NUM} times:")
print(f"timedelta = {t1:.3f}")
print(f"_delta_days = {t2:.3f} ({(t2/t1 - 1)*100:.2f}%)")
```

Result:
```
timedelta = 5.492
_add_days = 3.810 (-30.63%)
```

## Type of change

## Checklist


arkid15r (commented):
This would be another win for us on the performance grounds.

After glancing this I wonder if we could use some sort of a lightweight `timedelta` class version keeping all the syntax (add/radd methods with `days` support only).
Of course this would make sense if the performance gain doesn't degrade.

@KJhellico do you think it's worth a shot?

KJhellico:
I tried this, but there is no performance gain.

```python
from datetime import date, timedelta
import timeit

class td:
    def __init__(self, days: int = 0):
        self._days = days

    def __radd__(self, other):
        if isinstance(other, date):
            return date.fromordinal(other.toordinal() + self._days)
        return NotImplemented

    def __rsub__(self, other):
        if isinstance(other, date):
            return date.fromordinal(other.toordinal() - self._days)
        return NotImplemented


def _start_days(dt: date, n: int) -> date:
    return date.fromordinal(dt.toordinal() + n)

def with_timedelta():
    dt = date(2024, 1, 1)
    d2 = dt + timedelta(days=10)

def with_start_days():
    dt = date(2024, 1, 1)
    d2 = _start_days(dt, 10)

def with_class_td():
    dt = date(2024, 1, 1)
    d2 = dt + td(10)

NUM = 10_000_000

t1 = timeit.timeit(with_timedelta, number=NUM)
t2 = timeit.timeit(with_start_days, number=NUM)
t3 = timeit.timeit(with_class_td, number=NUM)

print(f"{NUM} times:")
print(f"timedelta = {t1:.3f}")
print(f"_start_days = {t2:.3f} ({(t2/t1 - 1)*100:.2f}%)")
print(f"class td = {t3:.3f} ({(t3/t1 - 1)*100:.2f}%)")
```

Result:
```
10000000 times:
timedelta = 5.534
_start_days = 3.749 (-32.25%)
class td = 5.505 (-0.52%)
```

holidays/calendars/gregorian.py
```diff
@@ -30,6 +29,14 @@
 WEEKDAYS = {w: i for i, w in enumerate(("mon", "tue", "wed", "thu", "fri", "sat", "sun"))}
 
 
+def _delta_days(dt: date, n: int) -> date:
+    """
+    Return date that is n days after (n > 0) or before (n < 0) specified date.
+    """
+
+    return date.fromordinal(dt.toordinal() + n)
```
arkid15r, line 37: Let's add some flexibility to this in case we ever need something more than `days`:

```suggestion
def _timedelta(dt: date, days: int = 0) -> date:
    """
    Return date that is `days` days after (days > 0) or before (days < 0) specified date.
    """

    return date.fromordinal(dt.toordinal() + days)
```

arkid15r (commented):
>I tried this, but there is no performance gain.

Thanks for sharing that! I'd definitely prefer performance over readability here. So our best next step would be to use the function based approach. I have just one naming/flexibility suggestion on this. 

The rest LGTM, great PR!

holidays/calendars/gregorian.py
KJhellico: Done.

arkid15r (approved):
LGTM

PR #1791, opened 2024-05-14
Update Malaysia holidays

KJhellico:
## Proposed change

Update Malaysia holidays:
- migrate to subdivision holidays methods
- rename subdivisions according to [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2:MY)
- l10n (en_US, **ms_MY**)
- update tests

## Type of change

## Checklist


holidays/countries/malaysia.py
```diff
@@ -136,259 +150,428 @@ def __init__(self, *args, **kwargs):
...
+            return None
 
-        # Chinese New Year (one day in the States of Kelantan and Terengganu,
-        # two days in the other States).
-        dts_observed.add(self._add_chinese_new_years_day("Chinese New Year"))
-        # The second day of Chinese New Year is not a federal holiday in
-        # Kelantan and Terengganu. However, it is gazetted as a state holiday
-        # in both states, effectively making it a nationwide holiday.
-        dts_observed.add(self._add_chinese_new_years_day_two("Chinese New Year Holiday"))
+        self.dts_observed = set()
```
arkid15r, line 159: There is an extra empty line which needs to be removed too
```suggestion
```

arkid15r (commented):
LGTM

holidays/countries/malaysia.py
KJhellico: `self.dts_observed` should be reset at each `_populate_public_holidays()` pass, for correct placement of "Eid al-Adha (observed)" in 2007 (early January, observed from 2006-12-31). Now it's made as `special_public_holidays_observed` in relevant subdivisions. If we remove `self.dts_observed = set()`, this holiday will be created by `self._populate_observed()` when `self.years` contains **both 2006 and 2007**, and will not be created when `self.years` **doesn't contain 2006**.

arkid15r: Gotcha! Let's add a comment then, it'll be 100% useful when reading this code next time.

arkid15r (approved):
💪 LGTM

Issue #1514, opened 2023-10-13
Feature request: add option to warm-up country import cache


eliasdorneles:
Hello,

Thanks for this library, nice work!

I'm using it in a web application that receives lots of requests, and recently had to debug an issue of thread contention caused by the `importlib.import_module` here: https://github.com/vacanza/python-holidays/blob/600ac6940a1d2112153773c9241d4f32066acedf/holidays/registry.py#L213

The threads were getting stuck trying to acquire the import lock, the call stack (obtained with `py-spy`) looked like this:

```
Thread 1525 (active): "AnyIO worker thread"
    _get_module_lock (<frozen importlib._bootstrap>:185)
    __enter__ (<frozen importlib._bootstrap>:170)
    _find_and_load (<frozen importlib._bootstrap>:1173)
    _gcd_import (<frozen importlib._bootstrap>:1204)
    import_module (importlib/__init__.py:126)
    get_entity (holidays/registry.py:211)
    __call__ (holidays/registry.py:193)
    country_holidays (holidays/utils.py:187)
```

The fix for it was to call `country_holidays` in my code at import time for every country i care about, thus avoiding the `importlib.import_module` from being called from inside the request lifecycle as the `entity` attribute will be already loaded.

What do you think of adding an option to do that by default?

For example, something like:

```
if os.getenv('PYTHON_HOLIDAYS_CACHE_WARMUP_ENABLED') == 'true':
    # preload it here
```

arkid15r:
Hi @eliasdorneles,
I know we don't have an issue template yet so I blame myself only for the fact I can't figure out what exactly caused the issue and what the preferred fix would be. In general, I'd avoid introducing env variable dependency but the PH code is totally open for contributions that fix existing users headaches. 

Having said that, I'd like to request some clarification on this (multi-threading?) issue. It'd be ideal to have a PR (draft is okay) to understand the problem and assess the fix implementation approach.

Thank you!

eliasdorneles:
Hello,

Right, I understand, I will try to get a minimal example that shows the problem.
I believe it's an issue related to `importlib.import_module` not being thread-safe.

The web application I mentioned (which I unfortunately cannot share the code as-is) is a FastAPI app, with endpoints defined as "sync" ([using regular `def` and not `async def`](https://fastapi.tiangolo.com/tutorial/sql-databases/#about-def-vs-async-def)), so the requests are served by anyio worker threads.

I will come back with more details later!

Izzette:
[...]

PR #1794, opened 2024-05-16
Synchronize entity dynamic imports

Izzette:
## Proposed change

Perform dynamic imports in registry in dedicated worker thread to avoid deadlocks from concurrent calls to `importlib.import_module` for the same module from different threads.  This may not be sufficient if other code outside of python-holdays is importing these moduels directly.  I believe this cpython issue is related https://bugs.python.org/issue38884.  This should resolve (through a different implementation) #1514.

Here is a flamegraph of the deadlock in action on one thread:
image

My main concern with this PR is that it may introduce a slight performance degradation.  I'm not sure adding a test will be feasible, as the deadlock is inherently racy.  Making a reliable production in the python-holidays code would require adding a lot of synchronization points within python-holidays to instrument such a test and hammering the `importlib.import_module` with many threads and than calling `importlib.invalidate_caches` in a loop until the issue is reproduced.

## Type of change

## Checklist

### DIFF
--- a/holidays/registry.py
+++ b/holidays/registry.py
@@ -11,7 +11,7 @@
 #  License: MIT (see LICENSE file)
 
 import importlib
-from concurrent.futures import ThreadPoolExecutor
+from threading import RLock
 from typing import Any, Dict, Iterable, Optional, Tuple, Union
 
 from holidays.holiday_base import HolidayBase
@@ -179,7 +179,7 @@ FINANCIAL: RegistryDict = {
     "ny_stock_exchange": ("NewYorkStockExchange", "NYSE", "XNYS"),
 }
 
-IMPORT_ENSURE_EXECUTOR = ThreadPoolExecutor(max_workers=1, thread_name_prefix="[HoliDynImp] ")
+IMPORT_LOCK = RLock()
 
 
 class EntityLoader:
@@ -228,10 +228,8 @@ class EntityLoader:
         if self.entity is None:
             # Avoid deadlock due to importlib.import_module not being thread-safe by caching all
             # the first imports in a dedicated thread.
-            ensure_futr = IMPORT_ENSURE_EXECUTOR.submit(importlib.import_module, self.module_name)
-            ensure_futr.result()  # We can discard the result, as it is now cached.
-
-            self.entity = getattr(importlib.import_module(self.module_name), self.entity_name)
+            with IMPORT_LOCK:
+                self.entity = getattr(importlib.import_module(self.module_name), self.entity_name)
 
         return self.entity
 
--- a/holidays/registry.py
+++ b/holidays/registry.py
@@ -179,6 +179,9 @@ FINANCIAL: RegistryDict = {
     "ny_stock_exchange": ("NewYorkStockExchange", "NYSE", "XNYS"),
 }
 
+# A re-entrant lock. Once a thread has acquired a re-entrant lock,
+# the same thread may acquire it again without blocking.
+# https://docs.python.org/3/library/threading.html#rlock-objects
 IMPORT_LOCK = RLock()
 
 
--- a/holidays/registry.py
+++ b/holidays/registry.py
@@ -11,6 +11,7 @@
 #  License: MIT (see LICENSE file)
 
 import importlib
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any, Dict, Iterable, Optional, Tuple, Union
 
 from holidays.holiday_base import HolidayBase
@@ -178,6 +179,8 @@ FINANCIAL: RegistryDict = {
     "ny_stock_exchange": ("NewYorkStockExchange", "NYSE", "XNYS"),
 }
 
+IMPORT_ENSURE_EXECUTOR = ThreadPoolExecutor(max_workers=1, thread_name_prefix="[HoliDynImp] ")
+
 
 class EntityLoader:
     """Country and financial holidays entities lazy loader."""
@@ -223,6 +226,11 @@ class EntityLoader:
     def get_entity(self) -> Optional[HolidayBase]:
         """Return lazy-loaded entity."""
         if self.entity is None:
+            # Avoid deadlock due to importlib.import_module not being thread-safe by caching all
+            # the first imports in a dedicated thread.
+            ensure_futr = IMPORT_ENSURE_EXECUTOR.submit(importlib.import_module, self.module_name)
+            ensure_futr.result()  # We can discard the result, as it is now cached.
+
             self.entity = getattr(importlib.import_module(self.module_name), self.entity_name)
 
         return self.entity
--- a/AUTHORS
+++ b/AUTHORS
@@ -52,6 +52,7 @@ Heikki Orsila
 Henrik Sozzi
 Hugh McNamara
 Hugo van Kemenade
+Isabelle COWAN-BERGMAN
 Jacky Han
 Jacob Punter
 Jaemin Kim


holidays/registry.py
```diff
@@ -177,6 +178,8 @@
     "ny_stock_exchange": ("NewYorkStockExchange", "NYSE", "XNYS"),
 }
 
+IMPORT_ENSURE_EXECUTOR = ThreadPoolExecutor(max_workers=1, thread_name_prefix="[HoliDynImp] ")
```
Izzette, line 181: In linux, as well as other *nix systems, thread names have a pretty show maximum length, in linux it's[ 16-bytes (including trailing \0)](https://man7.org/linux/man-pages/man3/pthread_getname_np.3.html#:~:text=whose%20length%20is%20restricted%20to%2016%0A%20%20%20%20%20%20%20characters%2C%20including%20the%20terminating%20null%20byte%20(%27%5C0%27)).  So i've abbreviated to 13 bytes in the prefix, which is already rather long.  and will only allow for the first two bytes of the python binary name `py` to show up.  Feel free to suggest a different abbreviation.

holidays/registry.py
```diff
@@ -222,6 +225,11 @@ def __str__(self) -> str:
     def get_entity(self) -> Optional[HolidayBase]:
         """Return lazy-loaded entity."""
         if self.entity is None:
+            # Avoid deadlock due to importlib.import_module not being thread-safe by caching all
+            # the first imports in a dedicated thread.
+            ensure_futr = IMPORT_ENSURE_EXECUTOR.submit(importlib.import_module, self.module_name)
+            ensure_futr.result()  # We can discard the result, as it is now cached.
```
Izzette, line 231: This will be called on every import, but the lock is only held on the first import.

Izzette:
Ok, I've managed to get what I believe is a minimal reproduction, but on my M2 with Python 3.12.3 the deadlock detection in importlib seems to be working correctly, so I don't reproduce a "true" deadlock, but I suspect that this is the same deadlock that is affecting us in production.  My code also does some benchmarking, but because the code that loads the module caches it, it doesn't really seem like it's likely to affect anyone ... and on top of it, it's only a degradation of about 8 µs (`Average duration 4276.57 ns` to `Average duration 12868.65 ns`).

Here's the deadlock detection in action:
```
Traceback (most recent call last):
  File "/Users/izzi/src/python-holidays/perf.py", line 125, in main
    t.join(30.0)
  File "/Users/izzi/src/python-holidays/perf.py", line 86, in join
    raise self.error
  File "/Users/izzi/src/python-holidays/perf.py", line 66, in run
    self._run_inner()
  File "/Users/izzi/src/python-holidays/perf.py", line 81, in _run_inner
    country_holidays(vals[1])
  File "/Users/izzi/src/python-holidays/holidays/utils.py", line 188, in country_holidays
    return getattr(holidays, country)(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/izzi/src/python-holidays/holidays/registry.py", line 207, in __call__
    cls = self.get_entity()
          ^^^^^^^^^^^^^^^^^
  File "/Users/izzi/src/python-holidays/holidays/registry.py", line 225, in get_entity
    self.entity = getattr(importlib.import_module(self.module_name), self.entity_name)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/Users/izzi/src/python-holidays/holidays/countries/isle_of_man.py", line 16, in <module>
    from holidays.countries.united_kingdom import UnitedKingdom, UnitedKingdomStaticHolidays
  File "<frozen importlib._bootstrap>", line 1357, in _find_and_load
  File "<frozen importlib._bootstrap>", line 418, in __enter__
  File "<frozen importlib._bootstrap>", line 346, in acquire
_frozen_importlib._DeadlockError: deadlock detected by _ModuleLock('holidays.countries.united_kingdom') at 4397028400
```

Alarmingly I see two other bugs that seem to be race conditions due to the dynamic import:
```
Traceback (most recent call last):
  File "/Users/izzi/src/python-holidays/perf.py", line 125, in main
    t.join(30.0)
  File "/Users/izzi/src/python-holidays/perf.py", line 86, in join
    raise self.error
  File "/Users/izzi/src/python-holidays/perf.py", line 66, in run
    self._run_inner()
  File "/Users/izzi/src/python-holidays/perf.py", line 81, in _run_inner
    country_holidays(vals[1])
  File "/Users/izzi/src/python-holidays/holidays/utils.py", line 188, in country_holidays
    return getattr(holidays, country)(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/izzi/src/python-holidays/holidays/registry.py", line 207, in __call__
    cls = self.get_entity()
          ^^^^^^^^^^^^^^^^^
  File "/Users/izzi/src/python-holidays/holidays/registry.py", line 225, in get_entity
    self.entity = getattr(importlib.import_module(self.module_name), self.entity_name)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1310, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/Users/izzi/src/python-holidays/holidays/countries/__init__.py", line 80, in <module>
    from .isle_of_man import IsleOfMan, IM, IMN
ImportError: cannot import name 'IsleOfMan' from partially initialized module 'holidays.countries.isle_of_man' (most likely due to a circular import) (/Users/izzi/src/python-holidays/holidays/countries/isle_of_man.py)
```

And:
```
Traceback (most recent call last):
  File "/Users/izzi/src/python-holidays/perf.py", line 125, in main
    t.join(30.0)
  File "/Users/izzi/src/python-holidays/perf.py", line 86, in join
    raise self.error
  File "/Users/izzi/src/python-holidays/perf.py", line 66, in run
    self._run_inner()
  File "/Users/izzi/src/python-holidays/perf.py", line 81, in _run_inner
    country_holidays(vals[1])
  File "/Users/izzi/src/python-holidays/holidays/utils.py", line 188, in country_holidays
    return getattr(holidays, country)(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/izzi/src/python-holidays/holidays/registry.py", line 207, in __call__
    cls = self.get_entity()
          ^^^^^^^^^^^^^^^^^
  File "/Users/izzi/src/python-holidays/holidays/registry.py", line 225, in get_entity
    self.entity = getattr(importlib.import_module(self.module_name), self.entity_name)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1337, in _find_and_load_unlocked
KeyError: 'holidays.countries'
```

These errors seem to happen more frequently on the countries that import other countries (Isle of Man, American Samoa, etc.).  I'm not surprised to see these errors, as while one thread is importing one module, another may be trying to access the partially initialized module and end up with either this KeyError or an ImportError due to using directly a partially initialized module.

We've also been able to confirm that this PR is working for us in our app when running locally, and doesn't degrade performance to any measurable extent.

Here's the code for the minimal-ish reproduction.  I've done some extra tuning with conditions to try to get the reproduction to be more reliable, but of course it's inherently probabilistic.  It's kind of messy, but it was super helpful for me to confirm the theory of our bug, and test this fix.

```python
#!/usr/bin/env python

from typing import Optional
from types import TracebackType, FrameType
import time
import sys
import traceback
from threading import Thread, Condition, Lock
import importlib
import random


def bench():
    from holidays.registry import EntityLoader

    iters = 10_000
    start = time.monotonic_ns()
    for _ in range(iters):
        # Raw entity loader performance, but because the `HolidayBase` get cached in `holidays`
        # module, this ultimately should get called only once per country, pre-process, unless this
        # API is used directly.
        EntityLoader("holidays.countries.australia").get_entity()

    end = time.monotonic_ns()
    avg_duration = (end - start) / iters

    print(f"Average duration {round(avg_duration, 2):.2f} ns")


class Counter:
    cond: Condition
    count: int

    def __init__(self):
        self.cond = Condition(Lock())
        self.count = 0

    def inc(self):
        with self.cond:
            self.count += 1
            self.cond.notify_all()

    def wait(self, expected_count: int):
        while True:
            with self.cond:
                if self.count >= expected_count:
                    return

                self.cond.wait()


class Loader(Thread):
    counter: Counter
    cond: Condition
    error: Optional[Exception]

    def __init__(self, counter: Counter, cond: Condition, *args, **kwargs):

        self.counter = counter
        self.cond = cond
        self.error = None

    def run(self):
        try:
            self._run_inner()
        except Exception as error:
            self.error = error

    def _run_inner(self):
        # Reload the things we need to test for a deadlock
        from holidays.utils import country_holidays
        from holidays.registry import COUNTRIES

        # Sync all threads to get them to execute the entity loader at the same time
        with self.cond:
            self.counter.inc()
            self.cond.wait()

        for vals in random.choices(list(COUNTRIES.values()), k=100):
            country_holidays(vals[1])

    def join(self, *args, **kwargs):
        super().join(*args, **kwargs)
        if not self.is_alive() and self.error is not None:
            raise self.error


# Fake exception class used to format the traceback for a "true" deadlock condition.
class FakeException(BaseException):
    pass


def main():
    bench()

    print("Reproducing deadlock / race conditions (this will take awhile) ...")
    reprod_attempts = 1_000
    reprod_threads = 100

    error_types_seen = set()

    for _ in range(reprod_attempts):
        # Clear the cached modules between each reproduction attempt
        for k in list(sys.modules.keys()):
            if k.split(".")[0] == "holidays":
                del sys.modules[k]

        importlib.invalidate_caches()

        counter = Counter()
        cond = Condition()
        ts: list[Thread] = []
        for _ in range(reprod_threads):
            t = Loader(counter, cond)
            t.start()
            ts.append(t)

        counter.wait(reprod_threads)
        with cond:
            cond.notify_all()

        for t in ts:
            try:
                t.join(30.0)
            except Exception as error:
                error_type = type(error)
                if error_type not in error_types_seen:
                    traceback.print_exception(error)
                    error_types_seen |= {error_type}

            if t.is_alive():
                frame: FrameType = sys._current_frames()[t.ident]
                tb = None
                while frame is not None:
                    tb = TracebackType(
                        tb, frame, frame.f_lasti, frame.f_lineno
                    )
                    frame = frame.f_back

                error = FakeException().with_traceback(tb)
                raise TimeoutError("Deadlock occurred") from error


if __name__ == "__main__":
    main()
```

Based on the above reproduction code, it seems that this PR fixes all 3 bugs caused by racy-ness in dynamic imports when using multithreading.

Izzette:
I've updated the code to use a re-entrant lock instead of a thread pool.  It has one main advantage over the thread pool strategy: it's re-entry safe.  With a thread pool, if the code were ever to be updated so that importing a module using the lazy loader could lazy load another module (I'm not sure why we'd do this), it could deadlock depending on import order when waiting on the future for the work to be executed on the same thread.  I don't think this will add any other deadlock potential, and it still passes the minimal reproduction check I shared.  For bonus points it reduces the already negligible performance impact of the thread pool strategy to undetectably negilgable (I got `Average duration 4621.12 ns` with RLock) and it doesn't spawn a new thread, consuming RLIMIT_NPROC limit or the kernel task limit.  I think the gopher in me just go carried away ...

arkid15r:
Hi @Izzette,

Thanks for going above and beyond in order to identify and fix the issue. I have to admit this isn't a kind of PRs we normally receive. I'm going to look into it this weekend and do my best to squeeze the changes into our May 20th release.

Again, I really appreciate the detailed description.

arkid15r (commented):
This looks stable, let's see how it goes in the wild.
I had just a minor addition to the PR -- a comment describing `RLock`. 

@Izzette sorry I couldn't process this earlier and include into v0.49

arkid15r (commented):
LGTM

@KJhellico @PPsyrius any comments/concerns on this?

KJhellico:
LGTM.

PPsyrius:
LGTM

arkid15r:
Alright, merging it then.

@Izzette thanks for your contribution!

arkid15r (approved):

PR #1859, opened 2024-07-09
 v1: Refactor entity name/location based on ISO standards

arkid15r:
## Proposed change

Update v1 code base with the new structure and entity name templates.

Some context to consider:

 - The uppercase format for entity files is not compliant with PEP8. In order to make it consistent I went this way as you can't import `as`, `in`, and `is` modules
 - The entities holding directory can be named per lowercase recommendation (e.g. `iso_3166` instead ov `ISO_3166`). I kept it consistent with its contents naming approach (I'm open to discussion of pros/cons here -- we still able to use the lowercase option)
 - The entity name is now camel-cased based on its standard entity code (e.g. `UsHolidays`, `UaHolidays`)
 - The ISO 3166 entities available by both alpha-2 and alpha-3 codes directly from `holidays` package only
 - All entities have `name` attribute and a standardized docstring
 - Some code still has references to old-style country/market instead of entity -- this will be fixed in following PRs
 - Test are also consolidated based on ISO standard entity code
 - The documentation and examples are going to be inconsistent until a separate docs related PR is merged
 - The snapshot directory still has old-style structure and will be updated later

I'm probably missing something so don't hesitate to comment while reviewing.

GitHub won't show the files renaming history even though it's visible via diff locally (please use local tool for better review experience)

## Type of change

## Checklist


KJhellico (approved):
LGTM. Just great!

PR #1895, opened 2024-08-02
Add documentation build test to CI/CD

KJhellico:
## Proposed change

Add documentation build test to CI/CD.

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -170,7 +170,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          python -m pip install --upgrade --requirement requirements/docs.txt
+          python -m pip install --requirement requirements/docs.txt
           python -m pip install .
 
       - name: Build docs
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -153,6 +153,30 @@ jobs:
...
+          python-version: '3.12'
+
+      - name: Install Dependencies
+        run: |
+          python -m pip install --requirement requirements/docs.txt
+          python -m pip install .
+
+      - name: Build docs
+        run: |
+          python -m sphinx -E -T -W -b html -D language=en -j auto -q docs/source docs/build
```
arkid15r, line 178: It looks like we need to adjust `make doc` target and use it here instead.

arkid15r (commented):
LGTM

.github/workflows/ci-cd.yml
KJhellico: As I understand, this is actually the same command.

arkid15r: Cool, so we can just use it as is then.

arkid15r (approved):
👍

PR #1931, opened 2024-08-18
Fix misleading France subdivision

MichaelMure:
## Proposed change
The France subdivision is misleading as those are **not** department (as in, sub-area in France mainland) but DOM (département d'outre-mer) or TOM (territoire d'outre-mer):  territories outside of France mainland, more or less administratively connected to France.

I don't know if that matters here, but selecting one should be considered optional as most people would live in France mainland, which is none of those choices.

See https://forum.hacf.fr/t/installer-lintegration-vacances/42799/7
## Type of change

- [X] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)

## Checklist

- [ ] I've run `make pre-commit`, it didn't generate any changes
- [ ] I've run `make test`, all tests passed locally

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -427,7 +427,7 @@ All other default values are highlighted with bold:
      - UNOFFICIAL
    * - France
      - FR
-     - Départements: BL, GES, GP, GY, MF, MQ, NC, PF, RE, WF, YT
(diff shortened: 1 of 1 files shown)


arkid15r (approved):
Thanks for the fix @MichaelMure!

PR #1952, opened 2024-08-24
Add New Zealand Subdivision Aliases (`EN`, `MI`, HASC)

PPsyrius:
## Proposed change

Migrates deprecated New Zealand subdivisions to new subdivision aliases system, also adding Maori name (ISO) and HASC code as optional aliases.

The "`South Canterbury`" subregion is undeprecated as they're still in active use as **Regional Holiday Anniversaries**: https://www.govt.nz/browse/work/public-holidays-and-work/public-holidays-and-anniversary-dates/ . As these territories were never given any acronyms by the New Zealand government, the full name is getting used here. I'm not sure about its name, though; maybe a possible rename to "`Canterbury (South)`" to match the source above?

"`Waitangi`" is removed from the `_deprecated_subdivisions` list altogether as they're mistakenly added there in the first place in #608. It isn't remapped into any other subdivision at the moment, so they should be safe to remove here.

## Type of change

## Checklist


PPsyrius:
This should hopefully be the last PR for "`_deprecated_subdivisions` migration to `subdivisions_aliases`" right now unless I'm mistaken. 🙏

holidays/countries/new_zealand.py
```diff
@@ -45,27 +46,71 @@ class NewZealand(ObservedHolidayBase, ChristianHolidays, InternationalHolidays,
         "WGN",  # Greater Wellington / Te Pane Matua Taiao
         "WKO",  # Waikato
         "WTC",  # West Coast / Te Tai o Poutini
+        # Subregions:
+        # https://www.govt.nz/browse/work/public-holidays-and-work/public-holidays-and-anniversary-dates/
+        "South Canterbury",
```
KJhellico, line 51: Unofficial code "STC" was previously used. If we have holidays for this province anyway, maybe we could use it? @arkid15r, how strictly do we follow ISO 3166-2?

arkid15r: The current version is not that strictly follows the standard. I hope this will change in`v1`. We probably should think of a way to keep it flexible to allow "extras" built on top of official ISO compliant entities.

arkid15r (approved):
@PPsyrius you've done a tremendous work migrating deprecated subdivisions to subdivisions aliases. Thank you!

I'm merging this as is for the next release so any other possible changes go in a separate PR(s).

PR #1956, opened 2024-08-28
Add Dominica holidays

### LINKED ISSUE #1172: Add Dominica holidays
Useful links:
  - https://en.wikipedia.org/wiki/Public_holidays_in_Dominica
  - https://www.timeanddate.com/holidays/dominica/
  - http://dominica.gov.dm/about-dominica/public-holidays

PPsyrius:
## Proposed change

Add Dominica holidays.

Resolves #1172 .

> [!NOTE]
> While Dominica's [Public Holidays Act](https://www.dominica.gov.dm/laws/chapters/chap19-10.pdf) itself states that Labour Day falls on `May, 1st`, this has been __*de facto*__ been observed on `1st Monday of May` since at least [2010](https://www.dominica-weekly.com/images/dominica-calendar-2010/1600-1280.jpg) continuously to the present (2024). As we cannot find the relevant amendments that caused that, we'll assume that the change started in 2010 for our current implementation.

## Type of change

## Checklist


tests/countries/test_dominica.py
```diff
@@ -0,0 +1,378 @@
...
+            "2009-09-03",
+            "2010-01-04",
+            "2019-09-19",
+        )
+
+    def test_labour_day(self):
+        name = "Labour Day"
+
+        # May, 1st.
+        self.assertNoHoliday("1989-05-01")
```
KJhellico, line 42: I think this check is redundant a bit, because we have already checked that there are no holidays at all in 1989.

tests/countries/test_dominica.py
```diff
@@ -0,0 +1,378 @@
...
+            "2017-08-07",
+            "2018-08-06",
+            "2019-08-05",
+            "2020-08-03",
+            "2021-08-02",
+            "2022-08-01",
+            "2023-08-07",
+            "2024-08-05",
+        )
+        self.assertNoHoliday("1989-05-01")
```
KJhellico, line 90: And here as well.

PPsyrius: Fixed.

tests/countries/test_dominica.py
PPsyrius: Fixed.

tests/countries/test_dominica.py
```diff
@@ -0,0 +1,376 @@
...
+            "2021-08-02",
+            "2022-08-01",
+            "2023-08-07",
+            "2024-08-05",
+        )
+        self.assertHolidayName(name_emancipation_day, dt)
+        self.assertNoHolidayName(name_first_mon_aug, range(1998, 2051))
+        self.assertNoHolidayName(name_emancipation_day, range(1990, 1998))
+
+    def test_2010_public_holiday(self):
```
arkid15r, line 93: Here and below:
```suggestion
    def test_2010_public_holidays(self):
```

arkid15r (commented):
Yay, another country!

LGTM, just the naming suggestion:

tests/countries/test_dominica.py
PPsyrius: Fixed.

arkid15r (approved):

PR #1979, opened 2024-09-05
Update Germany holidays: add `th` localization, deprecated non-standard ISO code, add subdivision aliases

PPsyrius:
## Proposed change

This PR includes 3 changes for Germany 🇩🇪  holidays:
- Deprecate the unofficial `BYP` code. Catholic municipalities in Bavaria will instead need to include the `CATHOLIC` category to mimic the previously implemented behavior.
  - Maybe we can do something similar to #1952 by adding "`Bayern (Katholisch)`" as a subregion too? I'm open to all suggestions here.
- Add Subdivision Aliases for all Germany subdivisions.
- Add `th` localization. (Checked with https://bangkok.diplo.de/th-th/botschaft/oeffnungszeiten/1670002 )

## Type of change

## Checklist


holidays/constants.py
```diff
@@ -53,3 +53,4 @@
 HEBREW = "hebrew"
 HINDU = "hindu"
 ISLAMIC = "islamic"
+PROTESTANT = "protestant"
```
KJhellico, line 56: It would be more logical to add the category CATHOLIC. Then Assumption Day will belong to `by_catholic_holidays`.

PPsyrius: Hmm, true - my only concern is that it changes the default behavior for `BY` subdivision.

PPsyrius: Implemented.

PPsyrius:
Hmm, seems like we might be able to consolidate this a bit further:
image

KJhellico (approved):
LGTM.

arkid15r (commented):
🇩🇪 LGTM

arkid15r (approved):

PR #2013, opened 2024-09-23
Update Japan holidays

KJhellico:
## Proposed change

Update Japan holidays:
- refactor substitute and national holidays calculation
- add Bank Holiday on Jan 1
- update tests

## Type of change

## Checklist


holidays/countries/japan.py
```diff
@@ -51,29 +51,34 @@ def __init__(self, *args, **kwargs) -> None:
...
+            self._add_observed(dt, name=tr("振替休日"), show_observed_label=False)
+
+        # A weekday between national holidays becomes a holiday too (国民の休日) -
+        # national holiday.
+        # In 1986-2006 it was only May 4 (between Constitution Day and Children's Day).
+        # Since 2006, it may be only the day between Respect for the Aged Day and
+        # Autumnal Equinox Day (in September).
+        if self._year <= 1985:
+            return None
+        if self._year <= 2006:
```
PPsyrius, line 71:
```suggestion
        elif 1986 <= self._year <= 2006:
```

KJhellico: But what will we gain from it? 👀

PPsyrius: Nvm, I thought it was supposed to be part of the same if-else as this section above:
```python
        if self._year <= 1985:
            return None
        if self._year <= 2006:
            may_4 = (MAY, 4)
            if not (self._is_monday(may_4) or self._is_sunday(may_4)):
                # National Holiday.
                self._add_holiday(tr("国民の休日"), may_4)
        else:
            for dt in dts:
                if dt.month == SEP and _timedelta(dt, +2) in dts:
                    # National Holiday.
                    self._add_holiday(tr("国民の休日"), _timedelta(dt, +1))
                    break
```

holidays/countries/japan.py
```diff
@@ -51,29 +51,34 @@ def __init__(self, *args, **kwargs) -> None:
...
+            may_4 = (MAY, 4)
+            if not (self._is_monday(may_4) or self._is_sunday(may_4)):
```
arkid15r, line 73: up to you: this version might be a bit more performant and (personally for me) easier to read.
```suggestion
            if not self._is_monday(may_4) and not self._is_sunday(may_4):
```

arkid15r (dismissed):
LGTM

arkid15r (approved):

PR #2014, opened 2024-09-23
Update United States holidays: fix Washington's Birthday in GA & IN

KJhellico:
## Proposed change

Update United States holidays: fix Washington's Birthday in GA.

## Type of change

## Checklist


holidays/countries/united_states.py
```diff
@@ -23,6 +22,7 @@
     SUN_TO_NEXT_MON,
     SAT_SUN_TO_PREV_FRI,
     SAT_SUN_TO_NEXT_MON,
+    US_GA_WASHINGTON_BIRTHDAY,
```
arkid15r, line 25: It seems this rule should be defined right in this scope just without `US_` prefix as it's going to be used by the US entity only.

PPsyrius: From some quick check the State of Indiana also uses the same calculation method as Georgia - at least for the year 2023-2025.

- https://www.mountvernon.org/library/digitalhistory/digital-encyclopedia/article/commemoration-of-george-washingtons-birthday
- https://www.in.gov/spd/benefits/state-holidays/

arkid15r (approved):
LGTM

PR #2048, opened 2024-10-11
Update Johor (Malaysia) and United Arab Emirates `weekend` observance, cleanup others

PPsyrius:
## Proposed change

This updates Johor (Malaysia) weekend observance from `FRI-SAT` back to `SAT-SUN` [as previously used between 1995-2013](https://forum.lowyat.net/topic/3035452), starting Jan 1, 2025 next year [as decreed on Oct 8, 2024](https://www.thestar.com.my/news/nation/2024/10/08/johor-realigns-its-weekend).

The same goes for the [United Arab Emirates's weekend observance change](https://time.com/6126260/uae-working-days-weekend/) from `FRI-SAT` to `SAT-SUN` starting Jan 1, 2022 (we missed this somehow in the last library-wide special weekend implementations).

I've also standardised the weekend observance code a bit for other countries to match the existing format used by Saudi Arabia.

## Type of change

## Checklist


KJhellico:
Need to add [here](https://github.com/vacanza/holidays/blob/3535df8108bd30ebc1d6bc92b545891d03f8172a/holidays/countries/malaysia.py#L230):
```python
        else:
            self._observed_rule = SUN_TO_NEXT_WORKDAY
            self.weekend = {SAT, SUN}
```
Then it will work correctly when `years` includes years with different rules (as in snapshots).

holidays/countries/algeria.py
```diff
@@ -32,6 +32,7 @@ class Algeria(HolidayBase, InternationalHolidays, IslamicHolidays):
     # %s (estimated).
     estimated_label = tr("(تقدير) %s")
     supported_languages = ("ar", "en_US", "fr")
+    weekend = {FRI, SAT}
```
KJhellico, line 35: But value is set in `_populate_public_holidays()`.

PPsyrius: I must've accidentally put it there while reviewing other entries, fixed.

KJhellico (approved):
LGTM.

PR #2025, opened 2024-09-27
Add Vietnamese translation of VietNam holidays

vunhatchuong:
## Proposed change

Add `vi` translation of Vietnam holidays.

This translation has been checked by me who was born and currently lives in Viet Nam and crossed-checked with the state [source](https://hethongphapluat.com/bo-luat-lao-dong-2019/dieu-112).

## Type of change

- [ ] New country/market holidays support (thank you!)
- [x] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new python-holidays functionality in general)

## Checklist

- [x] I've followed the [contributing guidelines][contributing-guidelines]
- [x] I've run `make pre-commit`, it didn't generate any changes
- [x] I've run `make test`, all tests passed locally

[contributing-guidelines]: https://github.com/vacanza/python-holidays/blob/dev/CONTRIBUTING.rst
[docs]: https://github.com/vacanza/python-holidays/tree/dev/docs/source

PR #2049, opened 2024-10-11
Update Vietnam holidays

KJhellico:
## Proposed change

Update Vietnam holidays:
- update Lunar New Year holidays calculation
- add substituted holidays

## Type of change

## Checklist


holidays/countries/vietnam.py
```diff
@@ -32,47 +64,81 @@ class Vietnam(ObservedHolidayBase, ChineseCalendarHolidays, InternationalHoliday
...
+        kwargs.setdefault("observed_since", 1995)
         super().__init__(*args, **kwargs)
 
+    def _add_lunar_new_year_observed(self, dt_lny: date) -> None:
+        if self._year <= 1994:
+            return None
+
+        day_names = {
+            # 29 of Lunar New Year.
+            -2: tr("29 Tết"),
```
KJhellico, line 78: This name (and its translation) are based on [this comment](https://github.com/vacanza/holidays/pull/2025#issuecomment-2387386364). 

To be honest, I don't quite understand the logic behind this name (why the second-to-last day of the year is called the 29th of something), so I'm asking @vunhatchuong to explain in more detail if possible.

vunhatchuong: I name it this way because there are no official names for it and this is a common way websites also refer to it.

The 29th could either be the second-to-last day of the year (2024) or the last day of the year (2022) because of how the Lunar calendar works so you also can't simply call it "Lunar New Year's Eve".

When you talk about Tết, people only keep track of at most the first 5 days so there are no misunderstandings of the name "29 Tết".

KJhellico: @vunhatchuong , thank you for your answer! So let's use this name.

PPsyrius (approved):
LGTM 🇻🇳

PR #2083, opened 2024-10-26
Update United States holidays: Alaska state holidays correction

PPsyrius:
## Proposed change

According to the ["Alaska Dispatch"](https://web.archive.org/web/20120502232826/http://www.alaskadispatch.com/article/happy-alaska-day-great-land):
> The changing of the colors and the guard is commemorated in Alaska each Oct. 18, designated "Alaska Day," an official state holiday in 1917 by the third Territorial Legislature. That same year, territorial lawmakers also created Seward's Day, also a state holiday, to mark on each final Monday in March the signing of the purchase treaty.

Thus, the start date for this should be 1917 instead of 1867 as in the current implementation.

As for `Seward's Day`, a local newspaper mentioned that [this wasn't observed in 1921](https://chroniclingamerica.loc.gov/lccn/sn86072239/1922-03-29/ed-1/seq-8/):
image

> [!NOTE]
> There's currently 2 state-level checks being used in United States test files at the moment, between 
> ```python3
> self.assertNoHolidayName(name, self.state_hols["AK"], range(1865, 1917))
> ```
> and
> ```python3
> self.assertNoHolidayName(name, UnitedStates(subdiv="AK", years=range(1865, 1917)))
> ```
> with negligible performance difference between the two - though we should probably standardize our test code to either one of them.

## Type of change

## Checklist

### DIFF
--- a/tests/countries/test_united_states.py
+++ b/tests/countries/test_united_states.py
@@ -1541,9 +1541,9 @@ class TestUnitedStates(CommonCountryTests, TestCase):
         name = "Alaska Day"
         self.assertNoHolidayName(name)
         self.assertHolidayName(
-            name, self.state_hols["AK"], (f"{year}-10-18" for year in range(1867, 2050))
+            name, self.state_hols["AK"], (f"{year}-10-18" for year in range(1917, 2050))
         )
-        self.assertNoHolidayName(name, self.state_hols["AK"], range(1865, 1867))
+        self.assertNoHolidayName(name, self.state_hols["AK"], range(1865, 1917))
         obs_dt = (
             "2009-10-19",
             "2014-10-17",
--- a/tests/countries/test_united_states.py
+++ b/tests/countries/test_united_states.py
@@ -921,8 +921,7 @@ class TestUnitedStates(CommonCountryTests, TestCase):
             (f"{year}-03-30" for year in range(1918, 1921)),
             (f"{year}-03-30" for year in range(1922, 1955)),
         )
-        self.assertNoHolidayName(name, self.state_hols["AK"], range(1865, 1918))
-        self.assertNoHolidayName(name, self.state_hols["AK"], 1921)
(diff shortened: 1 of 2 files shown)


tests/countries/test_united_states.py
```diff
@@ -916,9 +916,13 @@ def test_sewards_day(self):
...
         self.assertNoHolidayName(name)
         self.assertHolidayName(
-            name, self.state_hols["AK"], (f"{year}-03-30" for year in range(1918, 1955))
+            name,
+            self.state_hols["AK"],
+            (f"{year}-03-30" for year in range(1918, 1921)),
+            (f"{year}-03-30" for year in range(1922, 1955)),
         )
         self.assertNoHolidayName(name, self.state_hols["AK"], range(1865, 1918))
+        self.assertNoHolidayName(name, self.state_hols["AK"], 1921)
```
KJhellico, line 925:
```suggestion
        self.assertNoHolidayName(name, self.state_hols["AK"], range(1865, 1918), 1921)
```

KJhellico:
> There's currently 2 state-level checks being used in United States test files at the moment, between
> 
> ```python
> self.assertNoHolidayName(name, self.state_hols["AK"], range(1865, 1917))
> ```
> 
> and
> 
> ```python
> self.assertNoHolidayName(name, UnitedStates(subdiv="AK", years=range(1865, 1917)))
> ```

But option with `UnitedStates(subdiv=` is used only for cases with `observed=False` or for years before 1865.

KJhellico (approved):
LGTM.

arkid15r (commented):
Could you include the relevant sources into the code?

// I've clarified the contributing instructions as they weren't clear on how it should be done.

arkid15r (commented):
👍 looks great!

arkid15r (approved):

PR #2085, opened 2024-10-26
Update CONTRIBUTING.rst

arkid15r:
## Proposed change

Clarify source reference requirements.

## Type of change

## Checklist

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -11,9 +11,10 @@ Basics
 
 When contributing with fixes and new features, please start forking/branching
 from the `dev branch`_ to work on the latest code and reduce merging issues.
-If you add/change holiday official dates or names you must include references to
+If you add/change holiday official dates or names your code must include references to
 all sources (government sites, archived web pages, wiki pages, etc) you've used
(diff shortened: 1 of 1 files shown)


PPsyrius (commented):
That is perfectly understandable, LGTM.

In certain cases, would it be possible to simply reference vacanza repo PRs - or do we want the absolute source in all cases? It's just that non-UTF-8 links are sometimes unwieldy long, so I would like to shorten those references if possible.

KJhellico (approved):
LGTM.

arkid15r:
> That is perfectly understandable, LGTM.
> 
> In certain cases, would it be possible to simply reference vacanza repo PRs - or do we want the absolute source in all cases? It's just that non-UTF-8 links are sometimes unwieldy long, so I would like to shorten those references if possible.

The links should be fine -- as far as I remember ruff skips them while linting the code. So the absolute source URL is what should be included for now (however, I like the idea of a PR as a buffer/shortener, we might consider it later).

Yeah, they may be weird looking but it's the price we're willing to pay for having all references in one place. I expect some docs improvements soon as we want to migrate to `mkdocs`. I hope this move will change our documentation's state for better.

PR #2109, opened 2024-11-06
Add "Brasil, Bolsa, Balcão" ([B]³) holidays

araggohnxd:
## Proposed change

Add "Brasil, Bolsa, Balcão" - Brazil Stock Exchange and Over-the-Counter Market (B3) - to the list of available financial markets holidays.

This is still a work in progress, so I'm opening a draft PR just in case anyone else feels like contributing.

## Type of change

- [X] New country/market holidays support (thank you!)

## Checklist


PPsyrius:
@araggohnxd I've created a PR with some suggestions inside (testcases refractors, add your name to `AUTHORS` list, add snapshot, ...), nice to see new market support :)

KJhellico:
@araggohnxd, thank you for your contribution to Holidays!

As I can see, this list of BVMF holidays is based on the [current list of Brazil holidays](https://github.com/vacanza/holidays/blob/dev/holidays/countries/brazil.py). I have just created PR #2113, where I made some changes to holidays of the first half of the 20th century (based on available legal sources). If you have the time and opportunity, I would be grateful if you could review these changes, perhaps I have missed something or misunderstood. Also, you can use info from this PR to add localization of BVMF holidays if you would like to.

PPsyrius:
@KJhellico AFAIK I don't think our existing `make l10n` even have localization support for markets yet

araggohnxd:
Yeah, I think the next step is indeed adding localization, even though I remember not quite understanding how to do it for financial markets. I'll check out the related PR to see if it can throw some light on the issue for me.

araggohnxd:
@KJhellico Good job on spotting the Our Lady of Aparecida related law on your PR. I also added some Tiradentes' day sources on this PR, you may want to check that out.

I think the feature is pratically done, the only thing I'm not really sure about is the Holy Thursday, because I couldn't find any source backing up its existence as a holiday, other than [this resolution from the central bank](https://www.bcb.gov.br/pre/normativos/res/1998/pdf/res_2516_v2_P.pdf):

> PT: Parágrafo único. A Quinta-Feira da Semana Santa será considerada dia útil somente a partir do ano 2000.
> EN: Sole paragraph. The Holy Thursday will be considered a business day only from the year 2000 onwards.

About localization, I couldn't really make it work, so any help is appreciated.

snapshots/financial/B3.json
```diff
@@ -0,0 +1,1260 @@
+{
```
PPsyrius, line 1: @arkid15r @KJhellico I don't think we can get `B3` snapshot to be generated via `make snapshot` normally on top of `BVMF` ones as it's treated as a shorter aliases at the moment instead of "alternative codes" like `ECB`/`TAR` case.

Do we wanna update the `make snapshot` to handle this or remove this manually added copy instead?

araggohnxd: https://github.com/vacanza/holidays/blob/a4fcbc720a5fb458ee790b7cc1c56c329937f539/holidays/registry.py#L278

Is there even a reason why it limits the alias name to between 3 and 4 characters?

KJhellico: But there is no need for several snapshots for one entity. I think just one with name that matches the country/market value is enough.

araggohnxd: I guess I just did it because I saw the other financial markets do the same, didn't really think it through

PPsyrius: @KJhellico I think we should add 2-letter financial code support either way for consistency's sake though. 👀

scripts/generate_snapshots.py
```diff
@@ -66,11 +74,13 @@ def generate_country_snapshots(self):
...
             return None
 
-        supported_countries = list_supported_countries()
+        supported_countries = list_supported_countries(include_aliases=False)
         country_list = self.args.country or supported_countries
         if unknown_countries := set(country_list).difference(supported_countries.keys()):
             raise ValueError(f"Countries {', '.join(unknown_countries)} not available")
 
+        snapshot_path = "snapshots/countries"
+        self.prepare_snapshot_directory(snapshot_path)
```
KJhellico, line 83:
```suggestion
        if not self.args.country:
            self.prepare_snapshot_directory(snapshot_path)
```

scripts/generate_snapshots.py
```diff
@@ -92,19 +102,21 @@ def generate_financial_snapshots(self):
...
             return None
 
-        supported_markets = list_supported_financial()
+        supported_markets = list_supported_financial(include_aliases=False)
         market_list = self.args.market or supported_markets
         if unknown_markets := set(market_list).difference(supported_markets.keys()):
             raise ValueError(f"Markets {', '.join(unknown_markets)} not available")
 
+        snapshot_path = "snapshots/financial"
+        self.prepare_snapshot_directory(snapshot_path)
```
KJhellico, line 111:
```suggestion
        if not self.args.market:
            self.prepare_snapshot_directory(snapshot_path)
```

KJhellico (approved):
LGTM!

PR #2121, opened 2024-11-09
Update CH holidays: rename Neujahrestag to Neujahrstag

kurtkeller:
## Proposed change

Like in other German speaking countries, January 01 is called "Neujahr" in Switzerland, not "Neujahrestag". "Neujahr*e*stag" would be wrong anyway, it would be "Neujahrstag".

## Type of change

## Checklist

- [ ] I've followed the [contributing guidelines][contributing-guidelines]
- [ ] I've successfully run `make check`, all checks and tests are green

### DIFF
--- a/holidays/countries/switzerland.py
+++ b/holidays/countries/switzerland.py
@@ -66,7 +66,7 @@ class Switzerland(ObservedHolidayBase, ChristianHolidays, InternationalHolidays)
 
     def _populate_public_holidays(self):
         # New Year's Day.
-        self._add_new_years_day(tr("Neujahr"))
+        self._add_new_years_day(tr("Neujahrstag"))
 
         # Ascension Day.
         self._add_ascension_thursday(tr("Auffahrt"))
--- a/holidays/countries/switzerland.py
+++ b/holidays/countries/switzerland.py
@@ -66,7 +66,7 @@ class Switzerland(ObservedHolidayBase, ChristianHolidays, InternationalHolidays)
 
     def _populate_public_holidays(self):
         # New Year's Day.
-        self._add_new_years_day(tr("Neujahrestag"))
+        self._add_new_years_day(tr("Neujahr"))
 
         # Ascension Day.
         self._add_ascension_thursday(tr("Auffahrt"))
--- a/holidays/locale/de/LC_MESSAGES/CH.po
+++ b/holidays/locale/de/LC_MESSAGES/CH.po
@@ -27,7 +27,7 @@ msgstr ""
 "X-Generator: Poedit 3.4\n"
 
 #. New Year's Day.
-msgid "Neujahrestag"
+msgid "Neujahrstag"
 msgstr ""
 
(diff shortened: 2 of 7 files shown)


KJhellico:
[Here](https://www.bj.admin.ch/dam/bj/de/data/publiservice/service/zivilprozessrecht/kant-feiertage.pdf) is "Neujahrstag". This is an official source, as far as I understand.

kurtkeller:
> [Here](https://www.bj.admin.ch/dam/bj/de/data/publiservice/service/zivilprozessrecht/kant-feiertage.pdf) is "Neujahrstag". This is an official source, as far as I understand.

That's correct, I'm familiar with that document. It is a document which specifies which days are counted as holidays for legal matters, for example to calculate whether you filed claims within the specified period of time.

short answer:
The current spelling of "Neujahrestag" (check the 2nd "e") in the python file is wrong. "Neujahr" is what people actually use. "Neujahrstag" is understood and would be the correct spelling you also find in the referenced document. 

long answer:
You won't find "Neujahrestag" (as in the python file) anywhere in that document. The document uses "Neujahrstag", not "Neujahr  e  stag". Nevertheless what we really call it is "Neujahr", same as for Germany, Austria, Liechtenstein. The document also uses "Fahrtsfest" for Glarus (...subdiv_gl_public...) but we call it "Näfelser Fahrt" (the way it is in the python file). Furthermore the PDF document uses "Mariä Himmelfahrt" and "Mariä Empfängnis" in some places and in others "Maria ..." (ä vs a). The python file uses the traditional spelling throughout which is much more consistent. The PDF document uses the Italian names for Ticino (..._ti_...), French names for the French-speaking cantons and both French and German for the three cantons officially speaking both languages. The canton of Graubünden (..._gr_...) officially uses 3 languages: German, Italian and Romansh, but the document only uses German and Italian. (see: https://www.eda.admin.ch/aboutswitzerland/en/home/gesellschaft/sprachen/mehrsprachigkeit.html and https://www.fedlex.admin.ch/eli/cc/2009/821/en#art_21) The python file consistently uses the German names throughout. Switching to localized and even multi-language names could become messy. Besides, the PDF document also clearly states on page 2 that only August 1st is a holiday defined on national level. The other three holidays configured in _populate_public_holidays actually are are not national holiday but holidays defined on cantonal (subdiv) level, albeit they are holidays in all 26 cantons. Also here, changing that in the python file might be counter-productive.

KJhellico:
Yes, I mean, let's change it to "Neujahrstag".

arkid15r:
@kurtkeller does "Neujahrstag" look like a meet in the middle option?
Thanks for your native language experience input!

kurtkeller:
Yes, "Neujahrstag" is a good enough compromise.

arkid15r (approved):
Thanks for spotting and fixing that @kurtkeller!

KJhellico (approved):
LGTM!

Issue #1852, opened 2024-06-30
Refactor Curaçao as NL/CW subdivision


arkid15r:
### Description

Make sure Curaçao is available as a subdiv of Netherlands (based on https://www.iso.org/obp/ui/en/#iso:code:3166:CW).

### Related Issues

N/A

### Proposed Solution

Move the `CW` holidays logic to `NL` class. Refactor Curaçao logic to be based on `NL`'s subdiv `CW`

### Additional context

N/A

KJhellico:
It seems logical to do the same for [Aruba](https://www.iso.org/obp/ui/en/#iso:code:3166:AW).

arkid15r:
> It seems logical to do the same for [Aruba](https://www.iso.org/obp/ui/en/#iso:code:3166:AW).

Absolutely! I've updated the issue. Thanks!

PPsyrius:
Quick question here: Isle of Man, Jersey, and Guernsey were previously callable as Great Britain subdivisions before 2007: https://en.wikipedia.org/wiki/ISO_3166-2:GB as `GB-IOM,` `GB-GSY,` and `GB-JSY` respectively. Should these entities get the same treatment, too?

Ideally, we should have ways to call up the appropriate holiday codes rather than folding them into the main entities. That would be great here.

PanderMusubi:
@arkid15r if they have extra or other holidays on Aruba and Curaçao, please also add it to https://commenthol.github.io/date-holidays/#nl/2025/NL Thanks.

arkid15r:
> [@arkid15r](https://github.com/arkid15r) if they have extra or other holidays on Aruba and Curaçao, please also add it to https://commenthol.github.io/date-holidays/#nl/2025/NL Thanks.

Yes, we normally do that (`CW` already has multiple categories supported) -- feel free to provide any official sources you're aware of.
Thank you!

PR #2141, opened 2024-11-26
Introduce start/end year for entity

KJhellico:
## Proposed change

Introduce start/end year of period in which entity holidays are defined (with default values 1901 and 2100).

## Type of change

## Checklist


holidays/countries/thailand.py
```diff
@@ -682,10 +680,10 @@ def _populate_workday_holidays(self):
 
         # วันลอยกระทง
         # Status: In-Use.
+        # Started in 1941.
```
PPsyrius, line 683:
```suggestion
```
This is just a cut-off date due to the Thai LuniSolar calendar support range, not its start date.

PPsyrius (commented):
I'm glad this is finally getting implemented ❤️ 

This might be something for later, but having subdivision-level start dates would be nice for countries such as Australia where each state has a different holiday start date the United States where the state level exists much earlier, but US federal holidays observance only started in [1870](https://en.wikipedia.org/wiki/Federal_holidays_in_the_United_States) - or a future #1852 implementation. :eyes:

holidays/countries/finland.py
```diff
@@ -34,6 +34,7 @@ class Finland(HolidayBase, ChristianHolidays, InternationalHolidays):
     default_language = "fi"
     supported_languages = ("en_US", "fi", "sv_FI", "uk")
     supported_categories = (PUBLIC, UNOFFICIAL)
+    start_year = 1853
```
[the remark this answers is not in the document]
KJhellico: Just value before the earliest appearance of holiday (Runeberg Day in 1854). 🙂

holidays/countries/thailand.py
KJhellico: Yes, but I consider this attribute ( `start_year`) as just a technical, not as "start" in a juridical sense. In fact, "just a cut-off".

PPsyrius (approved):
LGTM 🕚

PPsyrius (approved):

PR #2186, opened 2024-12-20
Add `HebrewCalendarHolidays` to holidays groups, refactor Israel holidays

KJhellico:
## Proposed change

Add `HebrewCalendarHolidays` to holidays groups (for use in other countries with Hebrew holidays); refactor Israel holidays.

## Type of change

## Checklist


arkid15r (commented):
Yeah, this refactoring is long overdue. Thanks for addressing it!

holidays/groups/hebrew.py
arkid15r, line 1: I don't see the comment I thought I had left yesterday. Neither here nor in the mail. I didn't send it somehow -- sorry.

So the suggestion was about the other layer of abstraction we have for calendars. The methods like `def _yom_kippur_date():` should be there and in `groups/hebrew.py` you could operate w/ methods like `def _add_yom_kippur()`.

So in `israel.py` you could use `self._add_yom_kippur(tr("יום כיפור")) directly from the Hebrew holidays group.

KJhellico: Yes, initially I started with functions like `_add_yom_kippur()`. But in Israel, most holidays will require date value anyway (to handle with "observed" or add the Nth day of holiday).

PPsyrius: @KJhellico understandable, although I prefer @arkid15r's standardization approach more as that's what we've implemented for Islamic and Christian holiday groups.

Keep in mind that [Argentina also observes these officially as well](https://en.wikipedia.org/wiki/Public_holidays_in_Argentina) (Passover, Rosh Hashanah, and Yom Kippur)

KJhellico: Yes, I started with this when I started working on Montenegro.

PPsyrius (approved):
LGTM

arkid15r (commented):
LGTM, thanks for keeping it well-structured.

PR #2197, opened 2024-12-27
Update working days calculation

### LINKED ISSUE #2196: Incorrect working days calculation in case of substituted holidays from another year
Some countries (e.g. Ukraine, China) have working days moved to the previous or next year (via _substituted holidays_ scheme). If only the year **from which** non-working day is moved is specified, this change is not reflected.

For example, Ukraine has a non-working day moved from December 28, 1996 (Saturday) to January 2, 1997 (Thursday after New Year). Therefore, December 28, 1996 should be a working day. However, it is not:

```python
>>> import holidays
>>> ua = holidays.country_holidays("UA", years=1996)
>>> ua.is_working_day("1996-12-28")
False
```

KJhellico:
## Proposed change

Update working days calculation (if there is substituted holidays from another year) .
Close #2196.

## Type of change

## Checklist


PPsyrius (approved):
LGTM

holidays/groups/custom.py
```diff
@@ -18,7 +22,7 @@ class StaticHolidays:
     an external class.
     """
 
-    def __init__(self, cls) -> None:
+    def __init__(self, cls, inter_year: bool = False) -> None:
```
arkid15r: I don't think the overhead worth introducing the flag, we can check it for all entities.

holidays/groups/custom.py
```diff
@@ -33,3 +37,13 @@ def __init__(self, cls) -> None:
...
+
+            # Has substituted holidays from another year.
+            if inter_year:
+                self.weekend_workdays = set()
+                for subst_data in getattr(self, "special_public_holidays", {}).values():
+                    for data in _normalize_tuple(subst_data):
+                        if len(data) == 5:
+                            _, _, from_month, from_day, from_year = data
+                            from_date = date(from_year, from_month, from_day)
+                            self.weekend_workdays.add(from_date)
```
arkid15r: But this has to be out of the loop.

arkid15r (commented):
@KJhellico I've updated the code as per your request.
Please review the changes when you get a chance.

arkid15r (commented):
Adding to the queue.
@PPsyrius please review when you get a chance.

PPsyrius (approved):
LGTM

PR #2199, opened 2024-12-30
Update Ukraine holidays in 1991

KJhellico:
## Proposed change

Update Ukraine holidays in 1991 and fix substituted holidays in some years.

## Type of change

## Checklist


holidays/countries/ukraine.py
```diff
@@ -72,9 +73,8 @@ def _populate_common(self, is_martial_law: bool = False):
             dts_observed.add(self._add_womens_day(tr("Міжнародний жіночий день")))
 
         if (self._year >= 2022) == is_martial_law:
```
arkid15r, line 75: We should use `is` for this type of checks, but for that `is_martial_law` needs to be renamed to something like `during_martial_law`
```suggestion
        if (self._year >= 2022) is is_martial_law:
```

arkid15r (commented):
LGTM

arkid15r (approved):

PR #2227, opened 2025-01-13
Add faster date parsing from strings in ISO 8601 format

KJhellico:
## Proposed change

Add faster date parsing from strings in ISO 8601 format ("YYYY-MM-DD" or "YYYYMMDD"): use `date.fromisoformat()` instead of `dateutil.parser.parse()`.
Speed comparison:

```shell
python -m timeit -s "from dateutil.parser import parse" -n 100000 "d = parse('2024-12-25')"
100000 loops, best of 5: 20.7 usec per loop

python -m timeit -s "from datetime import date" -n 10000 "d = date.fromisoformat('2024-12-25')"
10000 loops, best of 5: 127 nsec per loop
```

## Type of change

## Checklist

### DIFF
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -593,6 +593,7 @@ class HolidayBase(dict[date, str]):
 
         # Key is `str` instance.
         elif isinstance(key, str):
+            # key possibly contains a date in YYYY-MM-DD or YYYYMMDD format.
             if len(key) in {8, 10}:
                 try:
                     dt = date.fromisoformat(key)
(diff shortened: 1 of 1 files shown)


holidays/holiday_base.py
```diff
@@ -592,10 +593,16 @@ def __keytransform__(self, key: DateLike) -> date:
 
         # Key is `str` instance.
         elif isinstance(key, str):
-            try:
-                dt = parse(key).date()
-            except (OverflowError, ValueError):
-                raise ValueError(f"Cannot parse date from string '{key}'")
+            if len(key) in {8, 10}:
```
arkid15r, line 596: Could you add a comment with key examples for 8 and 10 length.

arkid15r (commented):
It looks like a great performance improvement!
Perhaps we can find more in our `dateutli` use cases.

LGTM 👍

arkid15r (commented):
LGTM

PPsyrius (approved):
LGTM

PR #2273, opened 2025-02-06
Check .mo rather than .po files in TestListLocalizedEntities

cjwatson:
## Proposed change

Now that `.po` files are no longer included in distribution bundles (https://github.com/vacanza/holidays/pull/2243), having this test look at `.po` files makes things slightly more difficult for downstream distributors such as Debian, since we run the tests on the equivalent of the built wheel.

## Type of change

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,6 +29,7 @@ Chanran Kim
 Chris McKeague
 Chris Turra
 Christian Alexander
+Colin Watson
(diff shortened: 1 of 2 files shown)


KJhellico (dismissed):
LGTM!

arkid15r (approved):
Thanks for fixing that @cjwatson

PR #2285, opened 2025-02-10
Unify certain holidays names translation

KJhellico:
## Proposed change

Unify certain holidays names translation: Labor Day, Children's Day (in several variations) etc.

## Type of change

## Checklist


holidays/countries/greenland.py
```diff
@@ -75,7 +75,7 @@ def _populate_optional_holidays(self):
         # Epiphany.
         self._add_epiphany_day(tr("Mitaarneq"))
 
-        # International Workers' Day.
+        # Workers' Day.
```
arkid15r, line 78: I don't mind simplifying, just checked Wikipedia -- it has `Intertnational` in the name https://en.wikipedia.org/wiki/International_Workers%27_Day

holidays/financial/brasil_bolsa_balcao.py
```diff
@@ -70,7 +70,7 @@ def _populate_public_holidays(self):
             self._add_holiday_apr_21(tr("Tiradentes"))
 
         if self._year >= 1925:
-            # Workers' Day.
+            # Worker's Day.
```
arkid15r, line 73: But not here :) ?

arkid15r (commented):
LGTM
A couple of questions to clarify:

holidays/countries/greenland.py
KJhellico: Hmm, looks like first I need to review Scandinavian holidays entirely.

holidays/financial/brasil_bolsa_balcao.py
KJhellico: This Brazilian holiday is originally named "Dia do Trabalhador", and translated as "Worker's Day" in [Brazil holidays](https://github.com/vacanza/holidays/blob/d55c5b4d7b126c1f493e86d9303f3fc6404bff51/holidays/countries/brazil.py#L119). "Workers' Day" would be "Dia dos Trabalhadores".

holidays/countries/greenland.py
KJhellico: Although Greenlandic name for this holiday is "Sulisartut ulluat" (and there is no word “international”), the translation into all Scandinavian languages includes this word. So let's leave it in the English version as well. 🤷‍♂️

holidays/financial/brasil_bolsa_balcao.py
arkid15r: Maybe it's more about localization vs translation? They can use their own name for that but as long as it's May 1st I treat it as "International Workers' Day". 

It also can be a wrong opinion :) so we should find a solution in a discussion. Probably, something to think about later.

arkid15r (commented):
LGTM 👍

PPsyrius (approved):
LGTM 🌐

PR #2299, opened 2025-02-17
Update subdivision names for Presidents' Day in the United States

parkerbxyz:
## Proposed change

Many subdivisions in the United States use a different name for “Washington's Birthday” federal holiday. This pull request adds the official name used by each subdivision for this holiday.

## Type of change

## Checklist


arkid15r (changes requested):
@parkerbxyz thanks for the hard work you've done to gather those pieces of standardization information!

It seems we need to address some duplicate holidays first (see `AK`, `ID` or `MD` snapshots)

holidays/countries/united_states.py
```diff
@@ -49,11 +49,34 @@ class UnitedStates(ObservedHolidayBase, ChristianHolidays, InternationalHolidays
     Frances Xavier Cabrini Day:
         - https://leg.colorado.gov/sites/default/files/2020a_1031_signed.pdf
 
-    Washington's Birthday (GA):
-        - https://www.gasupreme.us/court-information/holidays-2/
-
-    Washington's Birthday (IN):
-        - https://www.in.gov/spd/benefits/state-holidays/
```
KJhellico, line 56: @parkerbxyz, thank you for such a wide list of official sources! It would also be useful to keep these two links, especially for GA (which is not listed in your sources).

holidays/countries/united_states.py
```diff
@@ -208,7 +231,31 @@ def _populate_subdiv_holidays(self):
...
+                "PR",
+                "SC",
+                "TN",
+                "TX",
+                "UT",
+                "VA",
+                "VT",
+                "WA",
+                "WV",
+                "WY",
```
KJhellico, line 257: I think these lines should be moved to condition above (L222).

arkid15r (commented):
LGTM 👍 

// @KJhellico and I made the requested changes in order to include this PR into today's release.

KJhellico (approved):
LGTM.

arkid15r (approved):

holidays/countries/united_states.py
parkerbxyz: The reason I did it the way I did is to prevent removing “Washington's Birthday” from each subdivision pre-1971. By moving these lines up to the condition above, it removes the holiday before the rename, which is I don't think is correct.

PR #2310, opened 2025-02-25
Fix 2025 Chinese Calendar's Birthday of Buddha

PPsyrius:
## Proposed change

Fixed the wrong pre-calculated results for the 2025 Chinese Calendar-based "Birthday of Buddha" - this should've been on `MAY, 5` instead of `MAY, 4` as confirmed by [Hong Kong's](https://www.gov.hk/en/about/abouthk/holiday/2025.htm), [South Korea's](https://english.visitkorea.or.kr/svc/contents/contentsView.do?vcontsId=140038) and [Macao's](https://www.gov.mo/en/public-holidays/year-2025/) calendar - the only 3 entities that observes them as holidays at the moment.

I came across this by chance while sorting out test case errors for my WIP Macao implementation, and I'm surprised this wasn't spotted in our recent HK l10n implementation.

## Type of change

## Checklist


PPsyrius:
Sorry for the multiple commits, the test keeps failing due to artifacts from local Macao branch so `make check` is more or less unusable on my end at the moment 😔

KJhellico:
@PPsyrius, good catch! Indeed, the new moon is April 28, so 8th day of 4th lunar month is May 5. And this is both in the Korean and Chinese time zones. It looks like the source of our Chinese calendar needs to be checked in more detail...

KJhellico (approved):
LGTM!

arkid15r (approved):
LGTM 👍

PR #2332, opened 2025-03-06
Update Philippines holidays: 2025 Eid'l Fitr and Eid'l Adha Tentative Date

PPsyrius:
## Proposed change

- Adds 2025 Tentative Date of Eid'l Fitr and Eid'l Adha for the Philippines.

## Type of change

## Checklist


holidays/countries/philippines.py
```diff
@@ -188,24 +189,6 @@ class PHL(Philippines):
     pass
 
 
-class PhilippinesChineseHolidays(_CustomChineseHolidays):
```
KJhellico, line 191: Wouldn't it be better to keep it as list of verified dates? 👀

holidays/countries/philippines.py
```diff
@@ -223,6 +206,7 @@ class PhilippinesIslamicHolidays(_CustomIslamicHolidays):
         2022: (JUL, 9),
         2023: (JUN, 28),
         2024: (JUN, 17),
+        2025: (JUN, 6),
```
KJhellico, line 209: Should we wait for an official President's proclamation?

PPsyrius: Our default Islamic calendar implementation (Saudi Arabia's Umm al-Qura) has the Eid'l Fitr estimated date as `MAR, 30` - which is different from the Philippines-based tentative date of `MAR, 31` - which is why I want to get this fix out before the next release cycle.

Long term-wise we might want to consider pre-filling ASEAN countries Islamic holiday data though since their tentative date is often very different from Arab Gulf states.

Brunei Darussalam, Malaysia, Singapore, and Indonesia started the adoption of the [`New MABIMS Crescent Visibility Criteria`](https://journal.uinjkt.ac.id/index.php/ahkam/article/viewFile/22275/10513) back in 2024 to better unify their holidays arrangements. I think we can probably use data from https://accuhijri.github.io/ as a separate `islamic_mabims` calendar.

KJhellico: I fully agree with you. AHC -🔥!

holidays/countries/philippines.py
PPsyrius: `PhilippinesChineseHolidays` removal is now reverted in https://github.com/vacanza/holidays/pull/2332/commits/c37454606c0f54b30c057eb9d8db08a5c147b576, alongside a new 2025 entry 👍

KJhellico (approved):
LGTM.

arkid15r (approved):
🇵🇭 LGTM

PR #2759, opened 2025-07-26
Implement confirmed years range support for custom calendars

KJhellico:
## Proposed change

Countries with Eastern holidays have lists of dates (mappings) in their custom calendars that only _partially differ_ from the reference dates. I propose the concept of a "confirmed years range". If the year falls within this range (for a certain holiday), the date of this holiday will be considered "exact" regardless of its presence in mapping.

Azerbaijan has been updated as implementation example.

## Type of change

## Checklist


PPsyrius (commented):
Nice concept, I think this can be extended to other calendars as well i.e. Chinese & Hindu calendars

KJhellico:
Yes, this can be applied to other Eastern calendars of similar type. Islamic is just an example.

holidays/countries/azerbaijan.py
```diff
@@ -206,58 +206,25 @@ class AZE(Azerbaijan):
 
 
 class AzerbaijanIslamicHolidays(_CustomIslamicHolidays):
+    EID_AL_ADHA_CHECKED_RANGE = (2002, 2025)
```
arkid15r: I'd use something more similar in format and maybe `CONFIRMED` instead of `CHECKED`. Probably

```python
EID_AL_ADHA_DATES_CONFIRMED_YEARS = 
EID_AL_ADHA_DATES = 
```

holidays/calendars/islamic.py
```diff
@@ -3973,9 +3973,13 @@ class _IslamicLunar:
     def _get_holiday(self, holiday: str, year: int) -> Iterable[tuple[date, bool]]:
         estimated_dates = getattr(self, f"{holiday}_DATES", {})
         exact_dates = getattr(self, f"{holiday}_DATES_{_CustomCalendar.CUSTOM_ATTR_POSTFIX}", {})
+        checked_range = getattr(
+            self, f"{holiday}_CHECKED_RANGE_{_CustomCalendar.CUSTOM_ATTR_POSTFIX}", None
+        )
         for year in (year - 1, year):
             for dt in _normalize_tuple(exact_dates.get(year, estimated_dates.get(year, ()))):
-                yield date(year, *dt), year not in exact_dates
+                checked = checked_range and checked_range[0] <= year <= checked_range[1]
```
arkid15r: This approach always expects an adjacent block of years. Do we want to extend it to support one-offs or multiple ranges, e.g. something like (1990, 1995, range(2000, 2006), 2008)

arkid15r: Use `is_` prefix for bools
```suggestion
                is_checked = checked_range and checked_range[0] <= year <= checked_range[1]
```

PPsyrius: This could be useful for my WIP Tanzania holidays refactor as we got intermittent range of checked years there 👀

KJhellico: My main goal was to make it as simple as possible. :) Individual year values would be better to just include in `...DATES`. Separate year ranges - hmm, need to think about it...

KJhellico: Now support ranges like `((2000, 2010), (2018, 2024))`.

holidays/calendars/islamic.py
```diff
@@ -3973,9 +3973,15 @@ class _IslamicLunar:
...
+        checked_range = getattr(
+            self, f"{holiday}_DATES_CONFIRMED_YEARS_{_CustomCalendar.CUSTOM_ATTR_POSTFIX}", None
+        )
         for year in (year - 1, year):
             for dt in _normalize_tuple(exact_dates.get(year, estimated_dates.get(year, ()))):
-                yield date(year, *dt), year not in exact_dates
+                is_checked = checked_range and any(
+                    yfrom <= year <= yto for yfrom, yto in _normalize_tuple(checked_range)
+                )
+                yield date(year, *dt), not (year in exact_dates or is_checked)
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, there is already a "is_" prefix.

holidays/countries/azerbaijan.py
```diff
@@ -206,58 +206,25 @@ class AZE(Azerbaijan):
...
         2017: (JUN, 26),
-        2018: (JUN, 15),
         2019: (JUN, 5),
-        2020: (MAY, 24),
-        2021: (MAY, 13),
-        2022: (MAY, 2),
-        2023: (APR, 21),
-        2024: (APR, 10),
-        2025: (MAR, 30),
     }
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, look at `_CustomCalendarType::__new__` code.

arkid15r (commented):
Looks good 👍

PPsyrius (approved):
LGTM 👍 

Hindu holidays could probably use a similar treatment as well, but let's do that in a separate PR.

PR #2770, opened 2025-07-31
Input handling improvements for `get_nth_working_day`

### LINKED ISSUE #2758: `get_nth_working_day` with `n=0`
The current method seems to care only for positive and negative values of `n`, not 0. When we pass `n=0`, it returns the day passed.

When `n=0`, if the passed day is not a working day, we should return the next working day (similar to `n=1`). This is reasonable because judging by the name of the method, it should always return a working day.

For example, if Jul 26 is a Saturday,
**Current**
| Date passed | `n` | Date returned |
|--------|--------|--------|
| 26 Jul | 0 | 26 Jul |
| 26 Jul | 1 | 28 Jul |
| 28 Jul | 0 | 28 Jul |

**Desired**
| Date passed | `n` | Date returned |
|--------|--------|--------|
| 26 Jul | 0 | **28 Jul** |
| 26 Jul | 1 | 28 Jul |
| 28 Jul | 0 | 28 Jul |

PPsyrius:
## Proposed change

If `n` input for `get_nth_working_day` is 0 and the current date isn't a working day, this will now returns the next working day instead.

Resolves #2758 

## Type of change

## Checklist

### DIFF
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -1129,14 +1129,14 @@ class HolidayBase(dict[date, str]):
         Returns:
             The calculated working day after shifting by n working days.
         """
+        direction = +1 if n >= 0 else -1
         dt = self.__keytransform__(key)
         # Special case: n is 0 return today if working day, else get next working day.
         if n == 0:
             while not self.is_working_day(dt):
-                dt = _timedelta(dt, 1)
+                dt = _timedelta(dt, direction)
             return dt
         # General case.
-        direction = +1 if n >= 0 else -1
         for _ in range(abs(n)):
             dt = _timedelta(dt, direction)
             while not self.is_working_day(dt):
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -1116,6 +1116,7 @@ class HolidayBase(dict[date, str]):
         """Find the n-th working day from a given date.
 
         Moves forward if n is positive, or backward if n is negative.
+        If n is 0, will return today if a working day, else get next working day.
 
         Args:
             key:
@@ -1128,8 +1129,14 @@ class HolidayBase(dict[date, str]):
         Returns:
             The calculated working day after shifting by n working days.
         """
-        direction = +1 if n > 0 else -1
         dt = self.__keytransform__(key)
+        # Special case: n is 0 return today if working day, else get next working day.
+        if n == 0:
+            while not self.is_working_day(dt):
+                dt = _timedelta(dt, 1)
+            return dt
+        # General case.
+        direction = +1 if n >= 0 else -1
         for _ in range(abs(n)):
             dt = _timedelta(dt, direction)
             while not self.is_working_day(dt):
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -1131,12 +1131,9 @@ class HolidayBase(dict[date, str]):
         """
         direction = +1 if n >= 0 else -1
         dt = self.__keytransform__(key)
-        if n == 0:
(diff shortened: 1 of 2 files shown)


holidays/holiday_base.py
```diff
@@ -1128,8 +1129,14 @@ def get_nth_working_day(self, key: DateLike, n: int) -> date:
...
             The calculated working day after shifting by n working days.
         """
-        direction = +1 if n > 0 else -1
+        direction = +1 if n >= 0 else -1
         dt = self.__keytransform__(key)
+        # Special case: n is 0 return today if working day, else get next working day.
+        if n == 0:
+            while not self.is_working_day(dt):
+                dt = _timedelta(dt, direction)
+            return dt
```
KJhellico: It would be good to avoid duplicating this code, but I can't currently find a better way to do it...

holidays/holiday_base.py
```diff
@@ -1128,8 +1129,14 @@ def get_nth_working_day(self, key: DateLike, n: int) -> date:
         Returns:
             The calculated working day after shifting by n working days.
         """
-        direction = +1 if n > 0 else -1
+        direction = +1 if n >= 0 else -1
         dt = self.__keytransform__(key)
+        # Special case: n is 0 return today if working day, else get next working day.
```
KJhellico: This comment in fact repeats the line from docstring above.

holidays/holiday_base.py
PPsyrius: IMO this works, but I don't think this is more maintainable personally
```python
        direction = +1 if n >= 0 else -1
        dt = self.__keytransform__(key)
        for _ in range(abs(n) if n != 0 else 1):
            if n != 0:
                dt = _timedelta(dt, direction)
            while not self.is_working_day(dt):
                dt = _timedelta(dt, direction)
        return dt
```

holidays/holiday_base.py
```diff
@@ -1128,10 +1129,11 @@
...
             The calculated working day after shifting by n working days.
         """
-        direction = +1 if n > 0 else -1
+        direction = +1 if n >= 0 else -1
         dt = self.__keytransform__(key)
-        for _ in range(abs(n)):
-            dt = _timedelta(dt, direction)
+        for _ in range(abs(n) if n != 0 else 1):
+            if n != 0:
+                dt = _timedelta(dt, direction)
```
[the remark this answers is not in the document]
PPsyrius: Good suggestion, but IMO this might be more concise:
```python
        direction = +1 if n >= 0 else -1
        dt = self.__keytransform__(key)
        for _ in range(abs(n) or 1):
            if n:
                dt = _timedelta(dt, direction)
```
There's no need to put `steps` separately if it's only used once after all

holidays/holiday_base.py
```diff
@@ -1128,10 +1129,11 @@ def get_nth_working_day(self, key: DateLike, n: int) -> date:
...
+        direction = +1 if n >= 0 else -1
         dt = self.__keytransform__(key)
-        for _ in range(abs(n)):
-            dt = _timedelta(dt, direction)
+        for _ in range(abs(n) or 1):
+            if n:
+                dt = _timedelta(dt, direction)
             while not self.is_working_day(dt):
                 dt = _timedelta(dt, direction)
         return dt
```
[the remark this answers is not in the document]
PPsyrius: @coderabbitai Wouldn't this means the `while not self.is_working_day(dt):` for n==0 case is skipped altogether instead? `for _ in range(abs(0)):` won't even run in the first place

KJhellico (approved):
LGTM!

PR #2873, opened 2025-08-30
Add `common::TestCase::assertHolidayNameCount` method

KJhellico:
## Proposed change

Add `common::TestCase::assertHolidayNameCount` (and `assertNonObservedHolidayNameCount`) methods, for checking that a holiday with specified name occurs a specified number of times in every year.

## Type of change

## Checklist

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -235,6 +235,28 @@ class TestCase:
         """Assert non-observed holidays exactly match expected holidays."""
         self._assertHolidays("holidays_non_observed", *args)
 
+    def _assertHolidayCountPerYear(self, name, count, instance_name, *args):  # noqa: N802
+        """Helper: assert number of holidays with a specific name in every year match expected."""
+        holidays, items = self._parse_arguments(args, instance_name=instance_name)
+
+        holiday_dates = holidays.get_named(name, lookup="exact")
+        for year in items:
+            holiday_count = sum(1 for dt in holiday_dates if dt.year == year)
+            self.assertEqual(
+                count,
+                holiday_count,
+                f"`{name}` occurs {holiday_count} times in year {year}, should be {count}",
+            )
+
+    def assertHolidayCountPerYear(self, name, count, *args):  # noqa: N802
+        """Assert number of holidays with a specific name in every year match expected."""
+        self._assertHolidayCountPerYear(name, count, "holidays", *args)
+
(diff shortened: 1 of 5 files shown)


PPsyrius (dismissed):
LGTM 🛠️

tests/common.py
```diff
@@ -235,6 +236,35 @@ def assertNonObservedHolidays(self, *args):  # noqa: N802
         """Assert non-observed holidays exactly match expected holidays."""
         self._assertHolidays("holidays_non_observed", *args)
 
+    def _assertHolidayCountPerYear(self, name, count, instance_name, *args):  # noqa: N802
```
arkid15r: Let's simplify naming by dropping the`PerYear` part everywhere?

arkid15r: It also can benefit from `_assertHolidayNameCount`

tests/common.py
```diff
@@ -235,6 +236,35 @@
...
+            holiday_counts[dt.year] += 1
+
+        for year in items:
+            holiday_count = holiday_counts.get(year, 0)
+            self.assertEqual(
+                count,
+                holiday_count,
+                f"`{name}` occurs {holiday_count} times in year {year}, should be {count}",
+            )
+
```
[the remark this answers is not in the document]
KJhellico: I compared the performance of both methods (Counter/DefaultDict) using the following script:
```python
from collections import Counter, defaultdict
from timeit import timeit
import holidays


NAME = "Eid al-Adha"
hols = holidays.country_holidays("AZ", years=range(1990, 2050), language="en_US")


def with_counter():
    holiday_counts = Counter(dt.year for dt in hols.get_named(NAME, lookup="exact"))
    return holiday_counts


def with_defaultdict():
    holiday_counts = defaultdict(int)
    for dt in hols.get_named(NAME, lookup="exact"):
        holiday_counts[dt.year] += 1
    return holiday_counts


def main():
    num = 50_000
    t1 = timeit(with_counter, number=num)
    t2 = timeit(with_defaultdict, number=num)
    print(f"Counter: {t1:.5f} DefaulDict: {t2:.5f} delta={(1 - t2 / t1) * 100:.2f}%")


if __name__ == "__main__":
    main()
```
and got this result:

`Counter: 12.80548 DefaulDict: 12.41126 delta=3.08%
`

So DefaulDict is slightly faster.

arkid15r (approved):
LGTM

PR #2885, opened 2025-09-03
Fix: Corrects one-day discrepancy in estimated Islamic holidays for Iran

jntdst:
Proposed change
This PR addresses and fixes a one-day discrepancy for estimated Islamic (lunar) holidays in Iran.

Currently, future lunar holidays calculated by the library for Iran are consistently off by one day compared to the official Iranian calendar. For example, an event officially on a Tuesday is often incorrectly listed as Monday (estimated). This causes reliability issues for users planning based on future dates.

This change implements a structural fix directly within the Iran class in holidays/countries/iran.py:

The _add_islamic_calendar_holiday method is overridden: This method now intercepts Islamic holidays before they are added.

Automatic Date Correction: If a holiday is flagged as is_estimated, the method automatically adds timedelta(days=1) to its date, aligning it with the official calendar.

Removal of "(estimated)" Label: The islamic_show_estimated flag is now forced to False in the class constructor. Since the dates are corrected, displaying the "estimated" label is no longer necessary and would be confusing.

This approach provides a permanent and transparent solution, ensuring that all future holiday calculations for Iran are accurate without requiring any post-processing by the end-user.

Type of change
[ ] New country/market holidays support (thank you!)

[x] Supported country/market holidays update (calendar discrepancy fix, localization)

[ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)

[ ] Dependency update (version deprecation/pin/upgrade)

[ ] Bugfix (non-breaking change which fixes an issue)

[ ] Breaking change (a code change causing existing functionality to break)

[ ] New feature (new holidays functionality in general)

Checklist
[* ] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).

[* ] I've run make check locally; all checks and tests passed.


holidays/countries/iran.py
```diff
@@ -51,19 +52,41 @@ class Iran(HolidayBase, IslamicHolidays, PersianCalendarHolidays):
...
-    def __init__(self, *args, islamic_show_estimated: bool = True, **kwargs):
+    def __init__(self, *args, **kwargs):
         """
-        Args:
-            islamic_show_estimated:
-                Whether to add "estimated" label to Islamic holidays name
-                if holiday date is estimated.
+        Initializes the Iran holidays class.
+        The `islamic_show_estimated` parameter is removed and forced to False
+        to handle date corrections internally.
```
KJhellico: No need to remove this parameter, `False` as default value is enough.

holidays/countries/iran.py
```diff
@@ -51,19 +52,41 @@ class Iran(HolidayBase, IslamicHolidays, PersianCalendarHolidays):
...
+    ) -> set[date]:
+        """
+        Overrides the parent method to apply a one-day correction to estimated
+        Islamic holidays, based on the logic from the provided fix script.
+        """
+        corrected_dates = []
+        for dt, is_estimated in dates:
+            if is_estimated:
+                # Add one day to estimated dates to correct them.
+                corrected_dates.append((dt + timedelta(days=1), is_estimated))
```
KJhellico, line 83: We have `_timedelta` function in `holidays.calendars.gregorian` for that.

KJhellico (changes requested):

KJhellico:
Implemented in #2981, for Iran and other similar countries. @jntdst, thank you for pointing this specific thing out to me.

Issue #1712, opened 2024-03-02
Consider making holiday object a first-class citizen


arkid15r:
The current PH version uses simple date/str pair as an item of dict-based holidays container.

Do we need/want to extend it to something more complex that would allow us having additional attributes on holiday object level (not only name, but also type/category, observance, other flags, etc)? Would the flexibility worth the efficiency trade-off (at this moment I can think of at least the new object size)?

This is based on an idea from https://github.com/vacanza/python-holidays/issues/784#issuecomment-1431376193. The interface example -- https://github.com/commenthol/date-holidays/blob/4596fb6c83c1339316529ae73c78e39646739488/types/index.d.ts#L55

PR #2963, opened 2025-09-26
Add festival descriptions and info retrieval method

yadav-krish:
Added descriptions for various Hindu festivals and a method to retrieve festival information.

## Proposed change

## Type of change

## Checklist

### DIFF
--- a/holidays/calendars/hindu.py
+++ b/holidays/calendars/hindu.py
@@ -18,6 +18,8 @@ from holidays.calendars.custom import _CustomCalendar
 from holidays.calendars.gregorian import JAN, FEB, MAR, APR, MAY, AUG, SEP, OCT, NOV, DEC
 from holidays.helpers import _normalize_tuple
 
+import holidays
+
 BUDDHA_PURNIMA = "BUDDHA_PURNIMA"
 CHHATH_PUJA = "CHHATH_PUJA"
 DIWALI = "DIWALI"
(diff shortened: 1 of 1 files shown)


KJhellico:
First of all, I doubt the utility of this feature. And even if it's to be implemented, the calendar class is not the right place for it.

PPsyrius:
As @KJhellico says - I'm not really sure about the current implementation as well.

The concept itself is not that bad, I can see them being utilized by iCal or Home Assistant integration for example - but this could probably use a more standardized approach (i.e. `DIWALI` and `DIWALI_INDIA` should've yielded the same output as they're the same festival, just held on different date) and some test cases at the very least.

KJhellico:
I am thinking about the concept of HOLIDAY_ID as one of the attributes of the Holiday object (within the context of #1712). For example, it will have value NEW_YEARS_DAY for the holiday on January 1 (and, if present, January 2); LUNAR_NEW_YEAR for all days of the Chinese New Year; EID_AL_FITR for all days of Eid al-Fitr; possibly INDEPENDENCE_DAY for the country's independence holiday, regardless of its date; DIWALI for Diwali, regardless of its date, etc. Then, by creating a mapping {HOLIDAY_ID: HOLIDAY_DESCRIPTION}, it will be convenient to obtain a description of the holiday for iCal and other applications.

yadav-krish:
I have made the changes like removing the line "import holidays" which i had done globally. 
I have also included self parameter inside the get_festival_info function which I had forgotten earlier.

KJhellico:
@yadav-krish, can you show a example of using your code?

arkid15r:
Abandoned.

PR #2981, opened 2025-10-03
Update Islamic holidays calculation

KJhellico:
## Proposed change

Looking at the dates of Islamic holidays, we can see that some countries have relatively stable difference from our base Islamic calendar (Umm al-Qura). So, it would be a good idea to add the option to specify the value of this difference and consider it during the calculation.

This PR inspired by #2885 and actually supersede it.

## Type of change

## Checklist


holidays/calendars/islamic.py
```diff
@@ -3970,24 +3970,37 @@ class _IslamicLunar:
...
+                    year_from <= check_year <= year_to for year_from, year_to in confirmed_years
+                )
+
+                hol_date = date(check_year, *dt)
+                yield (
+                    _timedelta(hol_date, calendar_shift)
+                    if calendar_shift and not is_checked_year
+                    else hol_date,
+                    not is_confirmed,
                 )
```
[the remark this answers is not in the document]
KJhellico: When year is not in `confirmed_dates`, `hol_date` will have a value from `estimated_dates` (default Umm al-Qura value), and it needs to be shifted.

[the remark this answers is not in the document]
KJhellico: > years within `confirmed_years` represent officially verified years where the Umm al-Qura calculation is already correct

Now it's actually "where [Umm al-Qura calculation + shift] is already correct".

PPsyrius (dismissed):
LGTM 👍

holidays/calendars/islamic.py
```diff
@@ -3970,24 +3970,37 @@ class _IslamicLunar:
         2076: (DEC, 5),
     }
 
+    def __init__(self, calendar_shift: int = 0) -> None:
```
arkid15r: I'd like the param to have `_days` in the name. Also consider using `calendar_delta_days` and adding a docstring 👍

holidays/calendars/islamic.py
```diff
@@ -3970,24 +3970,37 @@ class _IslamicLunar:
...
                 confirmed_dates.get(check_year, estimated_dates.get(check_year, ()))
             ):
-                is_confirmed = check_year in confirmed_dates or any(
-                    year_from <= check_year <= year_to
-                    for year_from, year_to in _normalize_tuple(confirmed_years)
+                is_confirmed = is_checked_year or any(
+                    year_from <= check_year <= year_to for year_from, year_to in confirmed_years
+                )
+
+                hol_date = date(check_year, *dt)
```
arkid15r: Still stick to `hol_`s :) ?
Let's make it `holiday_date`

holidays/calendars/islamic.py
```diff
@@ -3970,24 +3970,37 @@ class _IslamicLunar:
...
+        confirmed_years = _normalize_tuple(
+            getattr(
+                self, f"{holiday}_DATES_CONFIRMED_YEARS_{_CustomCalendar.CUSTOM_ATTR_POSTFIX}", ()
+            )
         )
         estimated_dates = getattr(self, f"{holiday}_DATES", {})
+        calendar_shift = self.__calendar_shift
 
         for check_year in (year - 1, year):
+            is_checked_year = check_year in confirmed_dates
```
arkid15r: Shouldn't it be `confirmed_year`?

arkid15r (commented):
Awesome implementation! 
Let's take care of naming and it's good to go 👍

PPsyrius (approved):

arkid15r (commented):
👍

PR #2986, opened 2025-10-06
Add `_add_observed`/`_move_holiday`'s `force_observed` support

PPsyrius:
## Proposed change

Based on PR discussion in #2984.
> In some ways, the bot is right. When working on some countries (in Latin America, I think), I also wondered if there was a difference between the following cases:
>
> ```
> N) ... "Some holiday" - October 12, ...
> ...
> P) the holidays specified in paragraph N are observed on the following Monday
> ```
> and
> ```
> N) ... "Some holiday" - the first Monday after October 12
> ````
> It is logical to assume that in the second case, October 12 is simply a calculation base and should not be a holiday even when `observed=False`.
>
> Perhaps we should add a parameter to `_add_observed` and `_move_holiday` that would allow holidays to be moved unconditionally?

> Something like `force=True` or `force_observed=True`...

- Add `_add_observed`/`_move_holiday`'s `force_observed` support.
- Add `_move_holiday_forced( ... )` as shorthand alias for `_move_holiday( ..., force_observed=True, show_observed_label=False)`.
- Refactor `TestFiji` to the latest test case standards.
- Add `_add_saint_patricks_day` for `ChristianHolidays` Holiday Group.

## Type of change

## Checklist


holidays/observed_holiday_base.py
```diff
@@ -155,7 +156,7 @@ def _add_observed(
         # Convert to date: (m, d) → use self._year; (y, m, d) → use directly.
         dt = dt if isinstance(dt, date) else date(self._year, *dt) if len(dt) == 2 else date(*dt)
 
-        if not self.observed or not self._is_observed(dt):
+        if not force_observed and (not self.observed or not self._is_observed(dt)):
```
KJhellico:
```suggestion
        if not (force_observed or (self.observed and self._is_observed(dt))):
```

holidays/observed_holiday_base.py
```diff
@@ -196,15 +197,29 @@ def _add_observed(
...
 
+    def _move_holiday_forced(
+        self, dt: date, rule: Optional[ObservedRule] = None
+    ) -> tuple[bool, Optional[date]]:
+        return self._move_holiday(
+            dt,
+            rule=rule,
+            force_observed=True,
+            show_observed_label=False,
+        )
```
KJhellico:
```suggestion
        return self._move_holiday(dt, rule=rule, force_observed=True, show_observed_label=False)
```

KJhellico (commented):
Great refactoring! Especially Canada and New Zealand.

KJhellico (approved):
LGTM! 👍

arkid15r (commented):
LGTM 👍 

Smart `ObservedRule` application.

PR #2998, opened 2025-10-09
Drop Python 3.9 support

PPsyrius:
## Proposed change

- Drops Python 3.9 support [due to its impending EOL status as of October 2025](https://devguide.python.org/versions/) - I'm waiting for the final deprecation notice before marking this as ready.
- Upgrades syntax to Python 3.10 standard (mainly [PEP 604](https://peps.python.org/pep-0604/) and dropping `importlib-metadata` which is now built-in as `importlib.metadata`).
- Remove `zipp>=3.19.1` pinning in `dev` and `test` requirements (from what I understand, this was previously only used by `importlib-metadata`).

## Type of change

- [x] Dependency update (version deprecation/pin/upgrade)

## Checklist


arkid15r (commented):
It seems this one is ready for merging when time comes.
Will take another look closer to the next release date.

PPsyrius:
IMO, considering that https://devguide.python.org/versions/ won't specify the specific dates, and that our next release won't be out until November anyway - it's probably safe to merge this in now, PTAL 🙏

KJhellico (approved):
LGTM.

arkid15r (commented):
> IMO, considering that https://devguide.python.org/versions/ won't specify the specific dates, and that our next release won't be out until November anyway - it's probably safe to merge this in now, PTAL 🙏

Sounds reasonable 👍

PR #2992, opened 2025-10-08
feat: Add NYSE half-day holidays

Kartik-kodes:
## Proposed change

This pull request adds support for half-day holidays for the New York Stock Exchange (NYSE), including the day before Independence Day, the day after Thanksgiving, and Christmas Eve.
Resolves #2976.

## Type of change

- [ ] New country/market holidays support (thank you!)
- [x] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [ ] I've run `make check` locally; all checks and tests passed

PR #2999, opened 2025-10-09
Introduce `HolidayBase::_is_weekday` method

PPsyrius:
## Proposed change

Add `_is_weekday` method as an alias for `not self._is_weekend(*args)`.

Inspired by #2992 - on one hand, its use case seems pretty narrow, but on the other hand, why not 🤷 - so I'm marking this as a draft PR for now.

## Type of change

## Checklist

### DIFF
--- a/tests/test_holiday_base.py
+++ b/tests/test_holiday_base.py
@@ -567,23 +567,31 @@ class TestHelperMethods(unittest.TestCase):
         self.hb.weekend = {MON, TUE}
         for dt in dts:
             self.assertTrue(self.hb._is_weekend(dt))
+            self.assertFalse(self.hb._is_weekday(dt))
 
         self.hb.weekend = {}
         for dt in dts:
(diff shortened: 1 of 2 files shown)


KJhellico:
Why not? 🙂

KJhellico (dismissed):
LGTM.

tests/test_holiday_base.py
```diff
@@ -567,23 +567,31 @@ def test_is_weekend(self):
         self.hb.weekend = {MON, TUE}
         for dt in dts:
             self.assertTrue(self.hb._is_weekend(dt))
+            self.assertFalse(self.hb._is_weekday(dt))
```
arkid15r, line 570: I couldn't figure out a better name for this method 🤷‍♂️ 
Let's keep it as is.

arkid15r (approved):

PR #3008, opened 2025-10-15
Add USGS and USNY financial market holiday calendars

### LINKED ISSUE #852: Hi,  would it be possible to add support for USGS and USNY as they have in [OpenGamma](https://strata.opengamma.io/apidocs/com/opengamma/strata/basics/date/HolidayCalendarIds.html)?
Hi,  would it be possible to add support for USGS and USNY as they have in [OpenGamma](https://strata.opengamma.io/apidocs/com/opengamma/strata/basics/date/HolidayCalendarIds.html)?

_Originally posted by @mg3146 in https://github.com/dr-prodigy/python-holidays/discussions/690_

ryleymao:
Implements support for US Government Securities (USGS) and US New York (USNY) financial market holiday calendars following SIFMA recommendations.

Resolves #852.

## Changes
- Added `USGovernmentSecurities` class with full bond market holidays including Good Friday, Columbus Day, and Veterans Day
- Added `USNewYork` class with standard New York financial market holidays
- Proper observed date handling (SAT_TO_PREV_FRI + SUN_TO_NEXT_MON)
- Comprehensive test coverage with 44 passing tests across both calendars
- Support for holidays from 1950 onwards with proper historical transitions
- Juneteenth support starting 2021, MLK Day starting 1998

## Testing
All tests pass:
```bash
python3 -m pytest tests/financial/test_us_government_securities.py tests/financial/test_us_new_york.py -v
# 44 passed in 0.08s
```


holidays/financial/us_government_securities.py
```diff
@@ -0,0 +1,94 @@
...
+#  License: MIT (see LICENSE file)
+
+from holidays.groups import ChristianHolidays, InternationalHolidays
+from holidays.observed_holiday_base import ObservedHolidayBase, SAT_TO_PREV_FRI, SUN_TO_NEXT_MON
+
+
+class USGovernmentSecurities(ObservedHolidayBase, ChristianHolidays, InternationalHolidays):
+    """US Government Securities (USGS) bond market holidays.
+
+    The US Government Securities market follows SIFMA (Securities Industry and Financial Markets
```
KJhellico, line 19: 1. Considering the similarity of the list of holidays for these markets, I think it would be advisable to implement a separate class (like `SIFMAHolidays`) and inherit from it.
2. It's advisable to implement the additional holidays denoted as "Early Close" as `HALF_DAY` holidays.
3. What is the source of information about USNY holidays (i.e., how do they differ from USGS holidays)?

ryleymao: Hi @KJhellico, I have acknowledged your feedback and made changes:

1. **Created `SIFMAHolidays` base class** - Both USGS and USNY now inherit from this shared class, eliminating code duplication
2. **Documented USNY vs USGS differences** - Added clear documentation in the docstrings. USNY excludes Good Friday, Columbus Day, and Veterans Day (which USGS observes as full closures)
3. **HALF_DAY holidays** - I haven't implemented the early close days yet. The base structure is in place and can be implemented in a future update, would you like me to add those in this PR, or would you prefer that as a separate follow-up PR?

This PR completes the core requirements of issue #852 (USGS and USNY calendars). Let me know if you'd like me to add the HALF_DAY holidays.

KJhellico: I think it's better to make it completely at once, if we already have all the necessary information.

> What is the source of information about USNY holidays (i.e., how do they differ from USGS holidays)?

?

ryleymao: Hi @KJhellico. Sounds good, thank you for clarifying. 

**Source for USNY vs USGS differences:**
USNY represents the NYSE (New York Stock Exchange) holiday calendar, while USGS follows SIFMA bond market recommendations. key differences:
- **NYSE (USNY)**: According to the official NYSE holiday calendar (https://www.nyse.com/markets/hours-calendars): the stock exchange closes on Good Friday but remains **open** on Columbus Day and 
Veterans Day
- **Bond Market (USGS)**: SIFMA recommendations: bond markets close on Good Friday, Columbus Day, AND Veterans Day
This is well documented in financial markets, the stock exchanges (NYSE/NASDAQ) trade on Columbus Day and Veterans Day, while bond markets are closed on those days.

I will now work on adding the HALF_DAY (early close) holidays as recommended by SIFMA, and push the update for another review.

KJhellico: Then I don't quite understand what USNY is? Since we already have support for [NYSE holidays](https://github.com/vacanza/holidays/blob/dev/holidays/financial/ny_stock_exchange.py).

ryleymao: I see where I went wrong. I went on to research the OpenGamma Strata source code and ISDA definitions and found that USNY and NYSE serve different purposes:

**USNY (US New York):**
- Represents general "New York business days" used in financial contracts (ISDA derivatives, OTC instruments)
- Referenced by OpenGamma Strata for financial calculations
- Includes standard US federal holidays: New Year's, MLK Day, Presidents Day, Memorial Day, Independence Day, Labor Day, Columbus Day, Veterans Day, Thanksgiving, Christmas
- Does NOT include Good Friday (unlike NYSE)

**NYSE (New York Stock Exchange):**
- Represents the actual stock exchange trading calendar
- Includes Good Friday (stock exchange closes)
- Does NOT include Columbus Day or Veterans Day in modern times (exchange stays open)
- Includes special one off closures (presidential funerals, disasters, etc.)

**Key Difference:**
USNY follows standard federal holidays including Columbus Day and Veterans Day, while NYSE doesn't observe these as trading holidays. Conversely, NYSE closes for Good Friday while USNY doesn't treat it as a business day closure.

**I found an error in my implementation:** I incorrectly excluded Good Friday, Columbus Day, AND Veterans Day from USNY. According to OpenGamma's source code, USNY should only exclude Good
Friday (keeping Columbus Day and Veterans Day).
I'll push a fix to keep Columbus Day and Veterans Day in USNY shortly.

**Sources:**
- OpenGamma Strata implementation: https://github.com/OpenGamma/Strata/blob/main/modules/basics/src/main/java/com/opengamma/strata/basics/date/GlobalHolidayCalendars.java
- OpenGamma API docs: https://strata.opengamma.io/apidocs/com/opengamma/strata/basics/date/HolidayCalendarIds.html

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,124 @@
...
+        # New Year's Day.
+        self._move_holiday(self._add_new_years_day("New Year's Day"))
+
+        # Martin Luther King Jr. Day (3rd Monday of January).
+        # Established as federal holiday in 1986, bond markets observed since 1998.
+        if self._year >= 1998:
+            self._add_holiday_3rd_mon_of_jan("Martin Luther King Jr. Day")
+
+        # Washington's Birthday (Presidents Day) - 3rd Monday of February.
+        self._add_holiday_3rd_mon_of_feb("Washington's Birthday")
```
KJhellico: Washington's Birthday, Memorial Day, Columbus Day, Veterans Day were changed by Uniform Monday Holiday Act.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,124 @@
...
+        # Christmas Day (December 25).
+        self._move_holiday(self._add_christmas_day("Christmas Day"))
+
+    def _populate_half_day_holidays(self):
+        # Early close days at 2:00 PM Eastern Time as recommended by SIFMA.
+        # Markets close at 2:00 PM ET (%s).
+        early_close_label = "Markets close at 2:00 PM ET (%s)"
+
+        # Helper to add early close day relative to a holiday.
+        def add_early_close(holiday_name, delta_days):
```
KJhellico: Categories must have independent implementation. `HALF_DAY` code cannot rely on the existence of holidays previously created by `PUBLIC` code.

ryleymao: Hi @KJhellico, I have just addressed your comments in my latest commit:
1. Added Uniform Monday Holiday Act comments for Washington's Birthday, Memorial Day, Columbus Day, and Veterans Day
2. Rewrote `_populate_half_day_holidays()` to calculate dates independently without relying on PUBLIC holiday lookups
I also added comprehensive HALF_DAY tests to achieve 100% coverage.

tests/financial/test_us_new_york.py
```diff
@@ -0,0 +1,267 @@
...
+from holidays.constants import HALF_DAY, PUBLIC
+from holidays.financial.us_new_york import USNewYork, USNY, USNewYorkFinancial
+from tests.common import CommonFinancialTests
+
+
+class TestUSNewYork(CommonFinancialTests, TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass(USNewYork)
+
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, stop to suggest return types annotations. Currently, they are not present everywhere in the project, and this will be resolved later for all parts of the code together.

holidays/financial/_sifma.py
KJhellico: Uniform Monday Holiday Act requires no commens, but the implementation of changes in the code (since we have `start_year` == 1950).

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+        self._add_holiday_4th_thu_of_nov("Thanksgiving Day")
+
+        # Christmas Day (December 25).
+        self._move_holiday(self._add_christmas_day("Christmas Day"))
+
+    def _populate_half_day_holidays(self):
+        # Early close days at 2:00 PM Eastern Time as recommended by SIFMA.
+        # HALF_DAY holidays must be independently calculated and cannot rely on PUBLIC holidays.
+
+        from holidays.calendars.gregorian import (
```
KJhellico: We don't use local imports.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+            MAY,
+            MON,
+            NOV,
+            THU,
+            _get_nth_weekday_of_month,
+            date,
+        )
+
+        # Calculate dates without adding them to the dictionary.
+        # Use property methods and date calculations to avoid duplicating holidays.
```
KJhellico: And excessive comments too.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Day before New Year's Day (if it's a weekday).
+        new_years = date(self._year, JAN, 1)
+        if self._is_weekday(new_years):
+            dt = _timedelta(new_years, -1)
```
KJhellico: Here you will get the date from _previous_ year.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+            if self._is_weekday(dt):
+                self._add_holiday("Markets close at 2:00 PM ET (New Year's Day)", dt)
+
+        # Day before Good Friday (Thursday before Easter).
+        # Good Friday is 2 days before Easter Sunday.
+        good_friday = _timedelta(self._easter_sunday, -2)
+        dt = _timedelta(good_friday, -1)
+        if self._is_weekday(dt):
+            self._add_holiday("Markets close at 2:00 PM ET (Good Friday)", dt)
```
KJhellico: It's Maundy Thursday. We have `_add_holy_thursday` for that.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Day before Memorial Day (Friday before last Monday of May).
+        # Memorial Day is the last Monday of May.
+        memorial_day = _get_nth_weekday_of_month(-1, MON, MAY, self._year)
+        dt = _timedelta(memorial_day, -1)
+        if self._is_weekday(dt):
+            self._add_holiday("Markets close at 2:00 PM ET (Memorial Day)", dt)
```
KJhellico: Yoou need `_add_holiday_3_days_prior_last_mon_of_may`.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Day before Independence Day (if it's a weekday).
+        independence_day = date(self._year, JUL, 4)
+        if self._is_weekday(independence_day):
+            dt = _timedelta(independence_day, -1)
+            if self._is_weekday(dt):
+                self._add_holiday("Markets close at 2:00 PM ET (Independence Day)", dt)
```
KJhellico: You need to check if Jul 4 is Tue-Fri actually.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Day after Thanksgiving (Black Friday).
+        # Thanksgiving is the 4th Thursday of November.
+        thanksgiving = _get_nth_weekday_of_month(4, THU, NOV, self._year)
+        dt = _timedelta(thanksgiving, +1)
+        if self._is_weekday(dt):
+            self._add_holiday("Markets close at 2:00 PM ET (Thanksgiving Day)", dt)
```
KJhellico: `_add_holiday_1_day_past_4th_thu_of_nov`.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Day before Christmas Day (if it's a weekday).
+        christmas = self._christmas_day
+        if self._is_weekday(christmas):
+            dt = _timedelta(christmas, -1)
+            if self._is_weekday(dt):
+                self._add_holiday("Markets close at 2:00 PM ET (Christmas Day)", dt)
```
KJhellico: You need to check if Dec 25 is Tue-Fri.

holidays/financial/_sifma.py
ryleymao: I moved all imports to top of the file

holidays/financial/_sifma.py
ryleymao: i removed comments keeping only necessary comments leaving methods cleaner

holidays/financial/_sifma.py
ryleymao: using helper methods now:
- _add_holy_thursday 
- _add_holiday_3_days_prior_last_mon_of_may 
- _add_holiday_1_day_past_4th_thu_of_nov

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+
+        # Day before Christmas (if Christmas is Tue-Fri).
+        if (
+            self._is_tuesday(self._christmas_day)
+            or self._is_wednesday(self._christmas_day)
+            or self._is_thursday(self._christmas_day)
+            or self._is_friday(self._christmas_day)
+        ):
+            self._add_holiday_dec_24("Markets close at 2:00 PM ET (Christmas Day)")
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, Yes, your remark is correct (2021 data confirms this), but the proposed implementation is not optimal.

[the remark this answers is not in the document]
KJhellico: Categories are independent; `HALF_DAY` category can be requested separately.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,136 @@
...
+        self._move_holiday(self._add_christmas_day("Christmas Day"))
+
+    def _populate_half_day_holidays(self):
+        # Day before Good Friday (Maundy Thursday).
+        self._add_holy_thursday("Markets close at 2:00 PM ET (Good Friday)")
+
+        # Friday before Memorial Day (3 days prior to last Monday of May).
+        self._add_holiday_3_days_prior_last_mon_of_may(
+            "Markets close at 2:00 PM ET (Memorial Day)"
+        )
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, your remark about the difference "before and after 1971" is correct, but attributes like `half_day_start_year` are currently not supported in the project.

ryleymao:
Hi @KJhellico, I see you're discussing the observed date logic with @coderabbitai. Please let me know if there's anything you need from me at this time, or if I should wait. Thank you

KJhellico:
> Please let me know if there's anything you need from me at this time, or if I should wait.

As you can see, the bot provides useful recommendations regarding the Uniform Monday Holiday Act (as I mentioned [earlier](https://github.com/vacanza/holidays/pull/3008#discussion_r2447565500)) and drew my attention to the fact that the logic behind calculating "early close dates" is more complex than I initially thought.
Therefore, I recommend first implementing the changes for "before/after 1971" for the relevant holidays. As for "early close dates," the approach [used for Washington’s Birthday in Georgia and Indiana](https://github.com/vacanza/holidays/blob/ee0a215c35c44df00a7d07c006811b502922d695/holidays/countries/united_states.py#L692) (custom observed rule) seems most suitable here.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,151 @@
...
+from holidays.observed_holiday_base import (
+    ObservedHolidayBase,
+    ObservedRule,
+    SAT_TO_PREV_FRI,
+    SUN_TO_NEXT_MON,
+)
+
+# Custom observed rule for early close dates.
+# Early close occurs on the day before an observed holiday if that observed date falls Tue-Fri.
+# If observed date is Mon, Sat, or Sun, there is no early close.
```
KJhellico: Look at 2021-2023. ;)

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,151 @@
...
+        if self._year >= 1998:
+            self._add_holiday_3rd_mon_of_jan("Martin Luther King Jr. Day")
+
+        # Washington's Birthday (Presidents Day).
+        # Observed on February 22 until 1970, then moved to 3rd Monday of February
+        # starting in 1971 by the Uniform Monday Holiday Act.
+        if self._year >= 1971:
+            self._add_holiday_3rd_mon_of_feb("Washington's Birthday")
+        else:
+            self._move_holiday(self._add_holiday_feb_22("Washington's Birthday"))
```
KJhellico:
```suggestion
        name = "Washington's Birthday"
        if self._year >= 1971:
            self._add_holiday_3rd_mon_of_feb(name)
        else:
            self._move_holiday(self._add_holiday_feb_22(name))
```
And the same for similar cases below.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Christmas Day (December 25).
+        self._move_holiday(self._add_christmas_day("Christmas Day"))
+
+    def _populate_half_day_holidays(self):
+        # Day before Good Friday (Maundy Thursday).
+        self._add_holy_thursday("Markets close at 2:00 PM ET (Good Friday)")
+
+        # Friday before Memorial Day (3 days prior to last Monday of May).
+        self._add_holiday_3_days_prior_last_mon_of_may(
```
KJhellico: 1971?...

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,157 @@
...
+        # Friday before Memorial Day (3 days prior to last Monday of May).
+        # Only applies from 1971 onwards when Memorial Day moved to last Monday of May.
+        if self._year >= 1971:
+            self._add_holiday_3_days_prior_last_mon_of_may(
+                "Markets close at 2:00 PM ET (Memorial Day)"
+            )
+
+        # Day before Independence Day (if observed Independence Day is Tue-Fri).
+        # Uses custom observed rule to calculate early close based on observed date.
+        jul_4 = date(self._year, 7, 4)
```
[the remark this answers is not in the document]
ryleymao: fixed

ryleymao:
@KJhellico I made some changes:
1. **Comment fix**: Removed the misleading Sat/Sun reference. After weekend observance rules, the observed date is never Sat/Sun, so the comment now correctly says "If observed date is Monday then there is no early close."
2. **DRY refactor**: Extracted holiday names to variables for Washington's Birthday, Memorial Day, Columbus Day, and Veterans Day.
3. **Memorial Day early close bug**: Added `if self._year >= 1971` check. Before 1971, Memorial Day was May 30, so the "3 days prior to last Monday" logic didn't make sense.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,157 @@
...
+            "Markets close at 2:00 PM ET (Thanksgiving Day)"
+        )
+
+        # Day before Christmas (if observed Christmas is Tue-Fri).
+        # Uses custom observed rule to calculate early close based on observed date.
+        dec_25 = date(self._year, DEC, 25)
+        observed_christmas = self._get_observed_date(dec_25, rule=SAT_TO_PREV_FRI + SUN_TO_NEXT_MON)
+        early_close_christmas = self._get_observed_date(observed_christmas, rule=SIFMA_EARLY_CLOSE)
+        if early_close_christmas != observed_christmas:  # Check if early close applies
+            self._add_holiday("Markets close at 2:00 PM ET (Christmas Day)", early_close_christmas)
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, stop to suggest adding return type annotation.

KJhellico:
> 1. If observed date is Monday then there is no early close.

> Look at 2021-2023.

KJhellico:
> Added `if self._year >= 1971` check.

Why do you think that before 1971 there was no early close before Memorial Day?

ryleymao:
@KJhellico, I removed the year restriction. Let me know if there's a different approach needed for before 1971 Memorial Day early closes. I had added it because 
Before 1971: Memorial Day = May 30 (fixed date)
From 1971: Memorial Day = Last Monday of May

KJhellico:
> Let me know if there's a different approach needed

Yes, it is needed.

> Before 1971: Memorial Day = May 30 (fixed date)

Where is the implementation of early close for this period?

ryleymao:
> Where is the implementation of early close for this period?
@KJhellico, this is implemented at `holidays/financial/_sifma.py:138-148`:

**Pre-1971 (lines 138-148):**
```
else:
      # Calculate early close based on observed May 30
      may_30 = date(self._year, MAY, 30)
      observed_may_30 = self._get_observed_date(
          may_30, rule=SAT_TO_PREV_FRI + SUN_TO_NEXT_MON
      )
      early_close_memorial = self._get_observed_date(
          observed_may_30, rule=SIFMA_EARLY_CLOSE
      )
      if early_close_memorial != observed_may_30:
          self._add_holiday("Markets close at 2:00 PM ET (Memorial Day)", early_close_memorial)
```

**1971+ (lines 134-137):** 
```
if self._year >= 1971:
     self._add_holiday_3_days_prior_last_mon_of_may(
         "Markets close at 2:00 PM ET (Memorial Day)"
     )
```

For years before 1971 this calculates the early close based on the observed May 30 date using the SIFMA_EARLY_CLOSE

KJhellico:
> Pre-1971 (lines 138-148):
> 1971+ (lines 134-137):

You don't have to explain in detail, I can see the code. :)

KJhellico:
OK, next step. I think that when finding early_close, it's simpler not to calculate `observed_...`, but to rely on the date of the holiday itself.

ryleymao:
> OK, next step. I think that when finding early_close, it's simpler not to calculate `observed_...`, but to rely on the date of the holiday itself.

just fixed this in the SIFMA_EARLY_CLOSE and removed intermediate observed date calculations :)

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,165 @@
...
+    SUN_TO_NEXT_MON,
+)
+
+# Custom observed rule for early close dates.
+# Early close occurs on the day before the observed holiday.
+# MON: no early close (holiday observed on Monday)
+# TUE-FRI: early close the day before (-1)
+# SAT: early close on Friday, 2 days before (-2, since Saturday observed on Friday)
+# SUN: no early close (holiday observed on Monday)
+SIFMA_EARLY_CLOSE = ObservedRule({MON: 0, TUE: -1, WED: -1, THU: -1, FRI: -1, SAT: -2, SUN: 0})
```
KJhellico: Again, look on 2022 and 2023 Christmas (Sunday and Monday).

ryleymao: @KJhellico Got it, I just fixed this by removed all references to "observed" dates since the rule is now applied directly to the actual holiday date.

KJhellico: > Again, look on 2022 and 2023 Christmas (Sunday and Monday).

❗

ryleymao: @KJhellico, Looking at 2022 and 2023 Christmas, I'm trying to understand:
- 2022: Dec 25 is Sunday (observed Monday Dec 26). Should there be an early close on Friday Dec 23 since markets are closed Monday Dec 26?
- 2023: Dec 25 is Monday. No early close since the day before (Sunday) is a weekend.
Is the issue that Sunday holidays (when observed on Monday) should have an early close on Friday? Or am I still missing the point?

KJhellico: No need to guess, there is a [source](https://www.sifma.org/resources/general/us-holiday-archive/).

ryleymao: I see what I was missing, the `SIFMA_EARLY_CLOSE` rule had `MON: 0, SUN: 0` which was incorrect.
I've updated the rule and now both Sunday AND Monday holidays now correctly get Friday early closes. I also updated the tests to match the actual SIFMA data and all pass.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,164 @@
...
+    SAT_TO_PREV_FRI,
+    SUN_TO_NEXT_MON,
+)
+
+# Custom observed rule for early close dates.
+# Applied directly to the actual holiday date to determine if there's an early close.
+# MON: holiday on Monday → early close on Friday (-3)
+# TUE-FRI: holiday on Tue-Fri → early close the day before (-1)
+# SAT: holiday on Saturday → early close on Friday (-2)
+# SUN: holiday on Sunday → early close on Friday (-2)
```
KJhellico:
```suggestion
```
No need for such comments.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,164 @@
...
+    This base class implements the common holidays observed across SIFMA-recommended
+    calendars, including both full closures (PUBLIC category) and early closes
+    (HALF_DAY category).
+
+    References:
+        * <https://www.sifma.org/resources/general/holiday-schedule/>
+        * <https://web.archive.org/web/20250210040000/https://www.sifma.org/resources/general/holiday-schedule/>
+
+    Historical references:
+        * <https://web.archive.org/web/20250210040000/https://www.sifma.org/resources/general/us-holiday-archive/>
```
KJhellico:
```suggestion
    References:
        * <https://web.archive.org/web/20250210040000/https://www.sifma.org/resources/general/holiday-schedule/>
        * [2015-2024](https://web.archive.org/web/20250210040000/https://www.sifma.org/resources/general/us-holiday-archive/)
        * [1996-2017](https://web.archive.org/web/20221206175502/https://www.sifma.org/wp-content/uploads/2017/06/misc-us-historical-holiday-market-recommendations-sifma.pdf)
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,164 @@
...
+        InternationalHolidays.__init__(self)
+        kwargs.setdefault("observed_rule", SAT_TO_PREV_FRI + SUN_TO_NEXT_MON)
+        super().__init__(*args, **kwargs)
+
+    def _populate_public_holidays(self):
+        # New Year's Day.
+        self._move_holiday(self._add_new_years_day("New Year's Day"))
+
+        # Martin Luther King Jr. Day (3rd Monday of January).
+        # Established as federal holiday in 1986, bond markets observed since 1998.
```
KJhellico:
```suggestion
        # Martin Luther King Jr. Day.
```
And the same for all holidays below (no need for such comments, except Good Friday note).

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Thanksgiving Day (4th Thursday of November).
+        self._add_holiday_4th_thu_of_nov("Thanksgiving Day")
+
+        # Christmas Day (December 25).
+        self._move_holiday(self._add_christmas_day("Christmas Day"))
+
+    def _populate_half_day_holidays(self):
+        # Day before Good Friday (Maundy Thursday).
+        self._add_holy_thursday("Markets close at 2:00 PM ET (Good Friday)")
```
KJhellico:
```suggestion
        begin_time_label = "Markets close at 2:00 PM ET (%s)"

        # Day before Good Friday.
        self._add_holy_thursday(begin_time_label % "Good Friday")
```
And use `begin_time_label` in rest of holidays' names.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,164 @@
...
+            self._move_holiday(self._add_remembrance_day(name))
+
+        # Thanksgiving Day (4th Thursday of November).
+        self._add_holiday_4th_thu_of_nov("Thanksgiving Day")
+
+        # Christmas Day (December 25).
+        self._move_holiday(self._add_christmas_day("Christmas Day"))
+
+    def _populate_half_day_holidays(self):
+        # Day before Good Friday (Maundy Thursday).
```
KJhellico: Where is Day before New Year's Day?

ryleymao: added now

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+#  specific sets of holidays on the fly. It aims to make determining whether a
+#  specific date is a holiday as fast and flexible as possible.
+#
+#  Authors: Vacanza Team and individual contributors (see CONTRIBUTORS file)
+#           dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
+#           ryanss <ryanssdev@icloud.com> (c) 2014-2017
+#  Website: https://github.com/vacanza/holidays
+#  License: MIT (see LICENSE file)
+
+from holidays.calendars.gregorian import DEC, FRI, JUL, MAY, MON, SAT, SUN, THU, TUE, WED, date
```
KJhellico:
```suggestion
from datetime import date

from holidays.calendars.gregorian import MAY, JUL, MON, TUE, WED, THU, FRI, SAT, SUN
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+        # Washington's Birthday.
+        name = "Washington's Birthday"
+        if self._year >= 1971:
+            self._add_holiday_3rd_mon_of_feb(name)
+        else:
+            self._move_holiday(self._add_holiday_feb_22(name))
+
+        # Good Friday.
+        # Bond markets close on Good Friday, though it is not a federal holiday.
+        # Note: Some SIFMA markets may treat this as an early close instead of full closure.
```
KJhellico:
```suggestion
        # Note: Some SIFMA markets may treat this as an early close instead of full closure.

        # Good Friday.
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+        if self._year >= 1971:
+            self._add_holiday_last_mon_of_may(name)
+        else:
+            self._move_holiday(self._add_holiday_may_30(name))
+
+        # Juneteenth National Independence Day.
+        if self._year >= 2021:
+            self._move_holiday(self._add_holiday_jun_19("Juneteenth National Independence Day"))
+
+        # Independence Day (July 4).
```
KJhellico:
```suggestion
        # Independence Day.
```
And the same for the rest.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Labor Day (1st Monday of September).
+        self._add_holiday_1st_mon_of_sep("Labor Day")
+
+        # Columbus Day.
+        name = "Columbus Day"
+        if self._year >= 1971:
+            self._add_holiday_2nd_mon_of_oct(name)
+        else:
+            self._move_holiday(self._add_holiday_oct_12(name))
```
KJhellico:
```suggestion
            self._move_holiday(self._add_columbus_day(name))
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def _populate_half_day_holidays(self):
+        begin_time_label = "Markets close at 2:00 PM ET (%s)"
+
+        # Day before New Year's Day.
+        # Apply SIFMA_EARLY_CLOSE rule to next year's January 1.
+        jan_1_next = date(self._year + 1, 1, 1)
+        early_close_nye = self._get_observed_date(jan_1_next, rule=SIFMA_EARLY_CLOSE)
+        if early_close_nye:
+            self._add_holiday(begin_time_label % "New Year's Day", early_close_nye)
```
KJhellico:
```suggestion
        self._add_holiday(
            begin_time_label % "New Year's Day",
            self._get_observed_date(self._next_year_new_years_day, rule=SIFMA_EARLY_CLOSE),
        )
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Day before Good Friday (Maundy Thursday).
```
KJhellico:
```suggestion
        # Day before Good Friday.
```
For style consistency.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+        # Friday before Memorial Day.
+        # Pre-1971: Memorial Day was May 30, so calculate early close from May 30.
+        # 1971+: Memorial Day is last Monday of May, so 3 days prior.
+        if self._year >= 1971:
+            self._add_holiday_3_days_prior_last_mon_of_may(begin_time_label % "Memorial Day")
+        else:
+            # Calculate early close based on May 30
+            may_30 = date(self._year, MAY, 30)
+            early_close_memorial = self._get_observed_date(may_30, rule=SIFMA_EARLY_CLOSE)
+            self._add_holiday(begin_time_label % "Memorial Day", early_close_memorial)
```
KJhellico:
```suggestion
        # Day before Memorial Day.
        name = begin_time_label % "Memorial Day"
        if self._year >= 1971:
            self._add_holiday_3_days_prior_last_mon_of_may(name)
        else:
            self._add_holiday(
                name,
                self._get_observed_date(date(self._year, MAY, 30), rule=SIFMA_EARLY_CLOSE),
            )
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Day before Independence Day.
+        # Uses custom observed rule to calculate early close based on holiday date.
+        jul_4 = date(self._year, JUL, 4)
+        early_close_jul_4 = self._get_observed_date(jul_4, rule=SIFMA_EARLY_CLOSE)
+        self._add_holiday(begin_time_label % "Independence Day", early_close_jul_4)
```
KJhellico:
```suggestion
        # Day before Independence Day.
        self._add_holiday(
            begin_time_label % "Independence Day",
            self._get_observed_date(date(self._year, JUL, 4), rule=SIFMA_EARLY_CLOSE),
        )
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Day after Thanksgiving (Black Friday).
```
KJhellico:
```suggestion
        # Day after Thanksgiving Day.
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+        self._add_holiday_1_day_past_4th_thu_of_nov(begin_time_label % "Thanksgiving Day")
+
+        # Day before Christmas.
+        # Uses custom observed rule to calculate early close based on holiday date.
+        dec_25 = date(self._year, DEC, 25)
+        early_close_christmas = self._get_observed_date(dec_25, rule=SIFMA_EARLY_CLOSE)
+        self._add_holiday(begin_time_label % "Christmas Day", early_close_christmas)
```
KJhellico:
```suggestion
        # Day before Christmas Day.
        self._add_holiday(
            begin_time_label % "Christmas Day",
            self._get_observed_date(self._christmas_day, rule=SIFMA_EARLY_CLOSE),
        )
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def __init__(self, *args, **kwargs):
+        ChristianHolidays.__init__(self)
+        InternationalHolidays.__init__(self)
+        kwargs.setdefault("observed_rule", SAT_TO_PREV_FRI + SUN_TO_NEXT_MON)
+        super().__init__(*args, **kwargs)
+
+    def _populate_public_holidays(self):
+        # New Year's Day.
+        self._move_holiday(self._add_new_years_day("New Year's Day"))
```
KJhellico:
```suggestion
        name = "New Year's Day"
        self._move_holiday(self._add_new_years_day(name))
        self._add_observed(self._next_year_new_years_day, name=name, show_observed_label=False)
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+        # Day after Thanksgiving Day.
+        self._add_holiday_1_day_past_4th_thu_of_nov(begin_time_label % "Thanksgiving Day")
+
+        # Day before Christmas Day.
+        self._add_holiday(
+            begin_time_label % "Christmas Day",
+            self._get_observed_date(self._christmas_day, rule=SIFMA_EARLY_CLOSE),
+        )
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, `self._christmas_day` defined in `ChristianHolidays`.

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,150 @@
...
+            self._move_holiday(self._add_holiday_may_30(name))
+
+        # Juneteenth National Independence Day.
+        if self._year >= 2021:
+            self._move_holiday(self._add_holiday_jun_19("Juneteenth National Independence Day"))
+
+        # Independence Day.
+        self._move_holiday(self._add_holiday_jul_4("Independence Day"))
+
+        # Labor Day (1st Monday of September).
```
KJhellico:
```suggestion
        # Labor Day.
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,150 @@
...
+            self._move_holiday(self._add_columbus_day(name))
+
+        # Veterans Day.
+        name = "Veterans Day"
+        if 1971 <= self._year <= 1977:
+            self._add_holiday_4th_mon_of_oct(name)
+        else:
+            self._move_holiday(self._add_remembrance_day(name))
+
+        # Thanksgiving Day (4th Thursday of November).
```
KJhellico:
```suggestion
        # Thanksgiving Day.
```

holidays/financial/_sifma.py
```diff
@@ -0,0 +1,150 @@
...
+        self._add_holiday_4th_thu_of_nov("Thanksgiving Day")
+
+        # Christmas Day (December 25).
```
KJhellico:
```suggestion
        # Christmas Day.
```

holidays/financial/us_new_york.py
```diff
@@ -0,0 +1,77 @@
...
+    Historical references:
+        * <https://web.archive.org/web/20250210040000/https://www.sifma.org/resources/general/us-holiday-archive/>
+    """
+
+    market = "USNY"
+
+    def _populate_public_holidays(self):
+        # Call parent to get all SIFMA holidays
+        super()._populate_public_holidays()
+
```
KJhellico, line 67: It is more logical to take a less complex entity as a basis and add elements to it than to take a more complex one and then remove elements. In general, if you do not plan to add any other markets based on SIFMA, it would be optimal to implement the USNY class and, based on it, USGS (adding Good Friday), without a separate SIFMA class.

ryleymao: I see what you are saying and agree with you, building from USNY (simpler) up to USGS (more complex) is more logical than the current approach.

I am wondering, given that all previous review comments have been addressed and tests are passing, would you be open to merging this PR as is and addressing the refactoring in a follow up issue? I'm happy to take on that refactoring work and be assigned to it. If you'd prefer the refactoring done now before merge, I can do that too. Just let me know your preference.

KJhellico: I think it's better to do everything properly from the start.

BTW, about tests: I don't review the tests in detail until the code for the holidays themselves is fully agreed upon, but looking at them now, I see that they don't quite meet our latest standards for tests.

holidays/financial/us_government_securities.py
```diff
@@ -0,0 +1,67 @@
...
+    2:00 PM Eastern Time (HALF_DAY holidays) as recommended by SIFMA.
+
+    The USGS calendar is referenced by the 2006 ISDA definitions 1.11 and is used for
+    instruments such as SOFR (Secured Overnight Financing Rate) fixings.
+
+    Key difference from USNY:
+        - USGS observes Good Friday as a full closure (USNY does not)
+
+    References:
+        * <https://www.sifma.org/resources/general/holiday-schedule/>
```
KJhellico, line 35:
```suggestion
```

holidays/financial/us_government_securities.py
```diff
@@ -0,0 +1,67 @@
...
+        self._add_good_friday("Good Friday")
+
+    def _populate_half_day_holidays(self):
+        # Get all USNY early closes first
+        super()._populate_half_day_holidays()
+
+        # Add Good Friday early close.
+        # Day before Good Friday (Maundy Thursday).
+        begin_time_label = "Markets close at 2:00 PM ET (%s)"
+        self._add_holy_thursday(begin_time_label % "Good Friday")
```
KJhellico, line 59:
```suggestion
    def _populate_public_holidays(self):
        super()._populate_public_holidays()

        # Good Friday.
        self._add_good_friday("Good Friday")

    def _populate_half_day_holidays(self):
        super()._populate_half_day_holidays()

        begin_time_label = "Markets close at 2:00 PM ET (%s)"

        # Day before Good Friday.
        self._add_holy_thursday(begin_time_label % "Good Friday")
```

holidays/financial/us_government_securities.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class USGS(USGovernmentSecurities):
+    pass
+
+
+class USBondMarket(USGovernmentSecurities):
+    pass
```
KJhellico, line 67: How standard is `USBondMarket` name?

holidays/financial/us_new_york.py
```diff
@@ -0,0 +1,162 @@
...
+    The USNY calendar represents general "New York business days" used in financial contracts
+    (ISDA derivatives, OTC instruments). While it follows SIFMA (Securities Industry and
+    Financial Markets Association) recommendations for most holidays, it differs from both
+    USGS (US Government Securities) and NYSE calendars.
+
+    Key differences:
+        - vs USGS: USNY does NOT observe Good Friday (USGS does)
+        - vs NYSE: USNY DOES observe Columbus Day and Veterans Day (NYSE does not in modern times)
+        - vs NYSE: USNY does NOT include special one-off closures (NYSE does)
```
KJhellico, line 39: As I can see, USNY is actually New York Banking Holidays, and don't follow SIFMA. If you know of sources that confirm otherwise, please cite them.

tests/test_utils.py
```diff
@@ -216,6 +216,8 @@ def test_list_supported_financial(self):
         self.assertIsInstance(xnys_subdivisions, list)
 
         financial_count = sum(
-            1 for path in Path("holidays/financial").glob("*.py") if path.stem != "__init__"
+            1
+            for path in Path("holidays/financial").glob("*.py")
+            if path.stem != "__init__" and not path.stem.startswith("_")
```
KJhellico, line 221:
```suggestion
            1 for path in Path("holidays/financial").glob("*.py") if path.stem != "__init__"
```

KJhellico:
@ryleymao, are you not planning to continue working on this PR?

Issue #3065, opened 2025-11-18
Test Case Refactor for remaining countries to New Testing Standard


PPsyrius:
A-L countries (except India) and all existing markets are already migrated in #2853, #3088, #3187

The rest here are roughly 35 countries each per group:
- [ ] M (22 entities)
- [ ] N-O, W-Z (21 entities)
- [ ] P-R, V (21 entities)
- [ ] S (33 entities)
- [ ] T-U (23 entities)

See mentioned PR above for implementation examples

VaishnaviBardapure:
Hi! I would like to work on migrating Mozambique test to the new standard as a part of this issue.

VaishnaviBardapure:
Since the Mozambique migration pattern worked well, I would like to take care of migrating the remaining countries to keep codebase consistent. 
Let me know it it's cool if I open a PR for those!

PPsyrius:
> Since the Mozambique migration pattern worked well, I would like to take care of migrating the remaining countries to keep codebase consistent. 
> Let me know it it's cool if I open a PR for those!

Yes, you're welcome to work on those 🙂

Let's do it on a letter-by-letter basis for now per PR i.e. letter M, N, O, ... countries - if you feel a particular batch is too large, you can further split those up as needed

PR #3090, opened 2025-11-28
Add `assert{variant}HolidayDatesInYear`, `assert{variant}HolidaysInYear` test methods

PPsyrius:
## Proposed change

I tried to better utilise the now centralised test setups implemented in #2881 by switching to `assert{variant}HolidayDatesInYear` and `assert{variant}HolidaysInYear` test methods to avoid duplicate test setups.

While not as blazingly fast as before that PR was implemented, which was ~55-70s (as we've yet to port all remaining countries' tests to the new system - see #3065 ), this should help us a bit.
```Python
### Pre-Optimized
110.33s 
111.99s
104.15s
Avg Time: 108.82s

### Post-`InYear` Tests Implementation
99.01s
92.54s
96.70s
Avg Time: 96.08s
```

Note that this still excludes cases where we did year-specific checks for multiple categories at the same time.

## Type of change

## Checklist


tests/common.py
```diff
@@ -242,7 +251,15 @@ def setUpClass(
...
-            setattr(cls, attr_name, cls.test_class(**init_kwargs))
+            instance = cls.test_class(**init_kwargs)
+
+            # Per-year index for InYear test methods.
+            year_index = defaultdict(list)
+            for dt, name in instance.items():
+                year_index[dt.year].append((dt, name))
+            instance._year_index = dict(year_index)
+
+            setattr(cls, attr_name, instance)
```
KJhellico: Would it be optimal to create dictionaries for 100 years for all variants (with subdivisions) and then use only the values for a few years?

KJhellico (commented):
Cool idea! 👍

tests/common.py
PPsyrius: Hmm,  I think we can probably narrow this down to 2000-2030, and force outlier ones to use the old filtering method 👀

KJhellico: I think, your first version (with `for dt, name in holidays.items() if dt.year == year`)  will still be the best balance between optimality and simplicity.

tests/common.py
```diff
@@ -385,6 +394,26 @@ def _assertHolidayDates(self, instance_name, *args):
...
 
+    # HolidayDatesInYear.
+    def _assertHolidayDatesInYear(self, year, instance_name, *args):
+        """Helper: assert holiday dates exactly match expected dates in the given year."""
+        holidays, dates = self._parse_arguments(args, instance_name=instance_name)
+        self._verify_type(holidays)
+
+        filtered_holidays = {
+            dt.strftime("%Y-%m-%d"): name for dt, name in holidays.items() if dt.year == year
+        }
```
KJhellico:
```suggestion
        filtered_holidays = {dt.strftime("%Y-%m-%d") for dt in holidays.keys() if dt.year == year}
```
We can optimize a bit because we don't need the names here.

tests/common.py
```diff
@@ -385,6 +394,26 @@ def _assertHolidayDates(self, instance_name, *args):
...
+
+        # Check one by one for descriptive error messages.
+        for dt in dates:
+            self.assertIn(dt, filtered_holidays, dt)
+
+        self.assertEqual(
+            len(dates),
+            len(filtered_holidays.keys()),
+            set(dates).difference(filtered_holidays.keys()),
```
KJhellico:
```suggestion
            len(filtered_holidays),
            set(dates).difference(filtered_holidays),
```

KJhellico:
```suggestion
            len(dates), len(filtered_holidays), set(dates).difference(filtered_holidays)
```

tests/common.py
PPsyrius: Good catch

KJhellico (dismissed):
LGTM! 👍

tests/common.py
```diff
@@ -75,6 +75,11 @@ def _method(self, *args, **kwargs):
                         raise TypeError(f"Expected (name, ...) arguments for {instance_name}")
                     name, *rest = args
                     return helper_func(self, name, instance_name, *rest, **kwargs)
+                elif method_type == "year":
+                    if not args:
+                        raise TypeError(f"Expected (year, ...) arguments for {instance_name}")
+                    year, *rest = args
```
arkid15r, line 81: Maybe we could simplify this by removing `year` from the method signature -- figure out year from the args data, just make sure the year is consistent for all provided dates (single year value).

Just something to think about for future.

arkid15r (dismissed):
Great improvement 👍

KJhellico (approved):

PR #3094, opened 2025-11-29
Update documentation: remove mixins from summary

KJhellico:
## Proposed change

Reorganize summary a bit: remove Mixins, move Changelog up in the list.

## Type of change

## Checklist

### DIFF
--- a/docs/mixins.md
+++ b/docs/mixins.md
@@ -1,3 +1,0 @@
-# Mixins
-
-::: holidays.mixins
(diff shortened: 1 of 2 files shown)


PPsyrius (dismissed):

arkid15r (commented):
I'm not sure how useful it'd be for end users as it's our internal helper.

@KJhellico @PPsyrius if you both think it should be visible via API docs I'm okay with adding it.

KJhellico:
> I'm not sure how useful it'd be for end users as it's our internal helper.

Hmm, it sounds rational. Then, wouldn't it be better to remove it from the documentation?

arkid15r:
> Hmm, it sounds rational. Then, wouldn't it be better to remove it from the documentation?

This seems to be a better way to handle this.

PPsyrius (approved):
LGTM🗒️

PR #3195, opened 2026-01-06
Add SCHOOL category and subdivision hooks for Belgium

Upanshi-Mittal:
## Proposed change

This PR introduces initial infrastructure for Belgian school holidays.

It adds support for the `SCHOOL` holiday category in Belgium and defines subdivision-specific hooks for the three Belgian school communities:
French, Flemish, and German-speaking.

No placeholder or speculative dates are included; holidays with unclear or unstable definitions are intentionally left unimplemented.

This PR follows the guidance discussed in #2506 and prepares the codebase for future, data-backed implementations.

## Type of change

## Checklist

- [x] I've run relevant local checks (sanity check for `SCHOOL` category initialization).


KJhellico:
@Upanshi-Mittal, what review are you waiting for on this draft? %)

Upanshi-Mittal:
@KJhellico this PR is intentionally limited to structure/scaffolding only.
It introduces the SCHOOL category for Belgium and defines subdivision hooks for the three communities, without adding any actual school holiday logic or dates yet.

The idea is to first align on the approach before following up with a separate PR that implements the real holiday rules and data.
I’m not expecting this to be merged as a complete feature at this stage, and I’m fine keeping it as a draft until then.

KJhellico:
I don't see the point in separate PR. Just fill this one with code.

Upanshi-Mittal:
Got it. I’ll extend this PR and add the actual Belgian school holiday logic here instead of creating a follow-up PR.

Upanshi-Mittal:
This PR now includes full implementation of Belgian school holidays for VLG, WBR and GER subdivisions along with documentation examples.

holidays/countries/belgium.py
```diff
@@ -12,78 +12,213 @@
...
-
-class Belgium(HolidayBase, ChristianHolidays, InternationalHolidays):
-    """Belgium holidays.
-
-    References:
-        * <https://en.wikipedia.org/wiki/Public_holidays_in_Belgium>
-        * <https://web.archive.org/web/20250331001402/https://www.belgium.be/nl/over_belgie/land/belgie_in_een_notendop/feestdagen>
-        * <https://nl.wikipedia.org/wiki/Feestdagen_in_België>
-        * <https://web.archive.org/web/20240816004739/https://www.nbb.be/en/about-national-bank/national-bank-belgium/public-holidays>
+def nth_weekday_of_month(year, month, weekday, n):
```
KJhellico: As I [already noted](https://github.com/vacanza/holidays/issues/2506#issuecomment-3715019244),

> Make sure you have an understanding of the established code style and approaches used in project.

The code currently doesn't meet the project requirements at all.

Upanshi-Mittal:
@KJhellico Thanks for pointing that out.

I realize now that I implemented the logic without properly following the existing country implementations and code style. I’ll refactor this PR to align with the established patterns used in the project, remove unnecessary custom helpers, and simplify the holiday calculations.

I’ll push an updated version shortly.

Upanshi-Mittal:
@KJhellico one small question before I go ahead and refactor more.

Paasvakantie is the part that is confusing me the most. I tried to derive it only from Easter-based rules, but for multiple years (like 2026, 2027, 2029) it doesn’t match the official Flemish school calendars. So I’m not sure what approach is preferred here.

Should I:

1. keep it fully rule-based even if it’s not perfectly accurate,
2. use an exception mapping with officially published dates for Paasvakantie,
3. or leave Paasvakantie unimplemented for now with a TODO?

I just want to make sure I’m following the project’s direction before continuing.

KJhellico:
> I tried to derive it only from Easter-based rules

Try to define it in the simplest possible rule (rules) with the fewest exceptions.

KJhellico:
[Official rules (Art. 4)](https://codex.vlaanderen.be/zoeken/Document.aspx?DID=1000373&param=inhoud):
> 5° de paasvakantie begint de eerste maandag van april en duurt twee weken.
> Indien Pasen in de maand maart valt, dan begint de paasvakantie de maandag na Pasen.
> Indien Pasen na 15 april valt, begint de paasvakantie de tweede maandag vóór Pasen.

holidays/countries/belgium.py
```diff
@@ -11,27 +11,45 @@
...
 
     References:
         * <https://en.wikipedia.org/wiki/Public_holidays_in_Belgium>
         * <https://web.archive.org/web/20250331001402/https://www.belgium.be/nl/over_belgie/land/belgie_in_een_notendop/feestdagen>
         * <https://nl.wikipedia.org/wiki/Feestdagen_in_België>
         * <https://web.archive.org/web/20240816004739/https://www.nbb.be/en/about-national-bank/national-bank-belgium/public-holidays>
+        * Official education calendars:
+            - Flemish Community (Vlaanderen)
+            - French Community (Fédération Wallonie-Bruxelles)
+            - German-speaking Community (Deutschsprachige Gemeinschaft)
```
KJhellico: There should be links here, with MD syntax.

holidays/countries/belgium.py
```diff
@@ -84,10 +102,190 @@ def _populate_bank_holidays(self):
         # Bank Holiday.
         self._add_christmas_day_two(tr("Banksluitingsdag"))
 
+    def _populate_school_holidays(self):
+        # School holidays vary by community (subdivision).
```
KJhellico: This method is redundant.

holidays/countries/belgium.py
```diff
@@ -84,10 +102,190 @@ def _populate_bank_holidays(self):
...
+    def _populate_subdiv_vlg_school_holidays(self):
+        """
+        School holidays for the Flemish Community (Vlaamse Gemeenschap).
+
+        Most vacation periods are rule-based and can be calculated
+        algorithmically, therefore they are implemented here.
+        """
+        year = self._year
+        easter = self._easter_sunday
+
```
KJhellico, line 139: We have `holidays.calendars.gregorian._timedelta` instead of `datetime.timedelta`. Also there is `_get_nth_weekday_from` and `_get_nth_weekday_of_month`.

holidays/countries/belgium.py
```diff
@@ -84,10 +102,190 @@ def _populate_bank_holidays(self):
...
+        # Current year's Christmas
+        christmas = date(year, 12, 25)
+        # if Christmas is on Monday, holiday will start that day.
+        # if Christmas is on Tue-Fri, start the previous Monday.
+        # if Christmas is a Sat/Sun, start the following Monday.
+        if christmas.weekday() <= 4: 
+            start = christmas - timedelta(days=christmas.weekday()+1)
+        else:
+            start = christmas + timedelta(days=7 - christmas.weekday()-1)
+        self._add_multiday_holiday(start, 13, name=tr("Kerstvakantie"))
```
KJhellico: It would be useful to override `_add_multiday_holiday` method to start adding holidays directly from the `start_date` rather than from the next day (this slightly simplifies dates calculations). Perhaps we will add this as a parameter later.

Upanshi-Mittal: Would you like me to override  _add_multiday_holiday in the Belgium class for this PR, or should we keep this for a follow-up change once a parameter is added?

KJhellico: > override _add_multiday_holiday in the Belgium class

Yes, that will be simpler for now.

holidays/countries/belgium.py
```diff
@@ -12,26 +12,55 @@
 
 from gettext import gettext as tr
 
-from holidays.constants import BANK, PUBLIC
+from holidays.calendars.gregorian import (
+    date,
```
KJhellico: Have you seen `date` in `holidays.calendars.gregorian`?

holidays/countries/belgium.py
```diff
@@ -84,6 +113,179 @@ def _populate_bank_holidays(self):
...
+    ) -> set[date]:
+        """Override to start adding holidays directly from start_date."""
+        if (holiday_name := name or self.get(start_date)) is None:
+            raise ValueError(f"Cannot infer holiday name for date {start_date!r}.")
+
+        return {
+            d
+            for delta in range(duration_days)
+            if (d := self._add_holiday(holiday_name, _timedelta(start_date, delta)))
+        }
```
KJhellico:
```suggestion
        return super()._add_multiday_holiday(
            _timedelta(start_date, -1), duration_days=duration_days, name=name
        )
```
Something like that is what I meant.

holidays/countries/belgium.py
```diff
@@ -84,6 +113,179 @@ def _populate_bank_holidays(self):
...
+        elif easter > date(year, 4, 15):
+            # second Monday before Easter
+            paas_start = _get_nth_weekday_from(-2, 0, easter)
+            paas_end = _timedelta(paas_start, days=15)
+        else:
+            paas_start = _get_nth_weekday_of_month(1, 0, 4, year)
+            paas_end = _timedelta(paas_start, days=14)
+        self._add_multiday_holiday(
+            paas_start, (paas_end - paas_start).days, name=tr("Paasvakantie")
+        )
```
KJhellico:
```suggestion
        easter_start = _get_nth_weekday_of_month(1, MON, APR, self._year)
        easter_duration = 14
        if easter.month == 3:
            easter_start = _timedelta(easter, -6)
        elif easter.day >= 16:
            easter_start = _timedelta(easter, -13)
            easter_duration = 15
        # Easter Holidays.
        self._add_multiday_holiday(easter_start, easter_duration, name=tr("Paasvakantie"))
```
This is more consistent with project established code style. Calculations should be simplified wherever possible.
All translated strings must have l10n comment in previous line.

holidays/countries/belgium.py
```diff
@@ -84,6 +113,179 @@ def _populate_bank_holidays(self):
...
+    def _populate_subdiv_wbr_school_holidays(self):
+        """
+        School holidays for Wallonia (French Community + German-speaking schools).
+        Only fixed and well-defined days are implemented.
+        Vacation periods are defined yearly by ministerial decree and are intentionally
+        left unimplemented.
+        """
+        year = self._year
+        # Armistice Day
+        self._add_remembrance_day(tr("Jour de l'Armistice"))
```
KJhellico: Belgium default language is Dutch (nl), so all strings should be in Dutch. And translations - in the corresponding .po files.

holidays/countries/belgium.py
```diff
@@ -84,6 +113,179 @@ def _populate_bank_holidays(self):
...
+        # Easter Monday.
+        self._add_easter_monday(tr("Lundi de Pâques"))
+
+        # Ascension Day.
+        self._add_ascension_thursday(tr("Jeudi de l'Ascension"))
+
+        # Whit Monday.
+        self._add_whit_monday(tr("Lundi de Pentecôte"))
+
+        # NOT IMPLEMENTED:
```
KJhellico: Why?

KJhellico (commented):
Since pre-commit check has already passed, we can take a look at the code. :) I also recommend starting to write tests.

holidays/countries/belgium.py
Upanshi-Mittal: That makes sense, using the parent implementation with a shifted start date is much cleaner. I’ll refactor _add_multiday_holiday accordingly.

holidays/countries/belgium.py
Upanshi-Mittal: This version is much clearer and matches existing patterns. I’ll switch to this structure.

KJhellico:
@Upanshi-Mittal, are you planning to complete this PR?

Upanshi-Mittal:
@KJhellico 
Yes, I’m planning to complete it.
I’m currently reviewing the remaining changes needed to align fully with the project style and will push an update this week.

holidays/countries/belgium.py
Upanshi-Mittal: The reason I initially left them unimplemented is that the school holiday periods for those communities are officially published each academic year and can vary due to government adjustments.
However, I understand that the project prefers rule-based implementations wherever possible. I will revisit the official decrees to identify the underlying rules and try to implement them with minimal exceptions if required.

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+        Most vacation periods are rule-based and can be calculated
+        algorithmically, therefore they are implemented here.
+        """
+        year = self._year
+        easter = self._easter_sunday
+
+        # Christmas holidays (week of 25 December).
+        christmas = date(year, DEC, 25)
+        start = _get_nth_weekday_from(-1 if christmas.weekday() <= 4 else 1, 0, christmas)
+        self._add_multiday_holiday(start, 13, name=tr("Kerstvakantie"))
```
KJhellico:
```suggestion
        # Christmas Holidays.
        name = tr("Kerstvakantie")
        christmas = self._christmas_day
        christmas_start = _get_nth_weekday_from(
            1 if self._is_weekend(christmas) else -1, MON, christmas
        )
        self._add_multiday_holiday(christmas_start, 32 - christmas_start.day, name=name)
```
Although `_add_holiday` correctly handles going beyond the current year, it's better to calculate the correct duration directly. And try to implement this for January vacation period as well.

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+            -1 if prev_christmas.weekday() <= 4 else 1,
+            0,
+            prev_christmas,
+        )
+        self._add_multiday_holiday(prev_start, 14, name=tr("Kerstvakantie"))
+
+        # Carnival holidays.
+        krokus_raw = _timedelta(easter, days=-47)
+        krokus_start = _get_nth_weekday_from(-1, MON, krokus_raw)
+        self._add_multiday_holiday(krokus_start, 7, name=tr("Krokusvakantie"))
```
KJhellico:
```suggestion
        # Carnival Holidays: Carnival Monday (Easter - 48 days) → 1 week.

        # Carnival Holidays.
        self._add_multiday_holiday(_timedelta(easter, -48), 7, name=tr("Krokusvakantie"))
```

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+        # - Otherwise → first Monday of April.
+        easter_start = _get_nth_weekday_of_month(1, MON, APR, self._year)
+        easter_duration = 14
+
+        if easter.month == 3:
+            easter_start = _timedelta(easter, -6)
+        elif easter.day >= 16:
+            easter_start = _timedelta(easter, -13)
+            easter_duration = 15
+        self._add_multiday_holiday(easter_start, easter_duration, name=tr("Paasvakantie"))
```
KJhellico, line 183:
```suggestion
        # Easter Holidays.
        self._add_multiday_holiday(easter_start, easter_duration, name=tr("Paasvakantie"))
```

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+
+        # Labour Day.
```
KJhellico:
```suggestion
        # Labor Day.
```
l10n comments should be in en_US locale.

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+        self._add_labor_day(tr("Dag van de Arbeid"))
+
+        # Ascension Day.
+        self._add_ascension_thursday(tr("Hemelvaart"))
+
+        # Friday after Ascension.
+        ascension_friday = _timedelta(easter, days=40)
+        self._add_holiday(tr("Vrijdag na O. L. H. Hemelvaart"), ascension_friday)
```
KJhellico:
```suggestion
        # Friday after Ascension Day.
        self._add_holiday_40_days_past_easter(tr("Vrijdag na O. L. H. Hemelvaart"))
```

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+
+        # Whit Monday / Pentecost Monday.
```
KJhellico:
```suggestion
        # Whit Monday.
```

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+        self._add_whit_monday(tr("Pinkstermaandag"))
+
+        # Summer holidays.
+        start = date(year, JUL, 1)
+        end = date(year, SEP, 1)
+        self._add_multiday_holiday(start, (end - start).days, name=tr("Zomervakantie"))
```
KJhellico:
```suggestion
        # Summer Holidays: From Jul 1 to Aug 31 (fixed period).

        # Summer Holidays.
        self._add_multiday_holiday(date(self._year, JUL, 1), 62, name=tr("Zomervakantie"))
```

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+
+        # Autumn holidays (week of 1 November).
+        nov_1 = date(year, NOV, 1)
+        start = _get_nth_weekday_from(-1 if nov_1.weekday() <= 5 else 1, 0, nov_1)
+        self._add_multiday_holiday(start, 7, name=tr("Herfstvakantie"))
```
KJhellico:
```suggestion
        # Autumn Holidays: Monday of the week of Nov 1 → 1 week.
        # If Nov 1 is on Sun, start from Nov 2.

        nov_1 = date(self._year, NOV, 1)
        autumn_start = _get_nth_weekday_from(1 if self._is_sunday(nov_1) else -1, MON, nov_1)
        # Autumn Holidays.
        self._add_multiday_holiday(autumn_start, 7, name=tr("Herfstvakantie"))
```

holidays/countries/belgium.py
```diff
@@ -84,6 +121,225 @@ def _populate_bank_holidays(self):
...
+
+        # Armistice Day (End of World War I).
```
KJhellico:
```suggestion
        # Armistice Day.
```

KJhellico (commented):
Few suggestions on Flemish holidays. The rest of methods also requires similar changes in code style.

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+        self._add_multiday_holiday(date(self._year, JAN, 1), duration, name=name)
+
+        # Carnival holidays.
+        carnival_raw = _timedelta(easter, days=-49)
+        carnival_start = _get_nth_weekday_from(1, MON, carnival_raw)
+        self._add_multiday_holiday(carnival_start, 14, name=tr("Krokusvakantie"))
+
+        # Spring Holidays.
+        spring_start = _get_nth_weekday_of_month(4, MON, APR, year)
+        self._add_multiday_holiday(spring_start, 14, name=tr("Paasvakantie"))
```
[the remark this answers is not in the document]
Upanshi-Mittal: @KJhellico 
Since Paasvakantie is already used for Easter holidays in other subdivisions, should WBR use a separate msgid (e.g. Lentevakantie) for Spring Holidays, or keep Paasvakantie for consistency?

Upanshi-Mittal:
The duplication mainly comes from adding public holidays (e.g. Labor Day) inside each subdivision’s school holiday method. Since these are also public holidays, I wanted to confirm whether they should remain duplicated in the school methods or be handled only in `_populate_public_holidays()` to avoid duplication.

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+        """
+        School holidays for the Flemish Community (Vlaamse Gemeenschap).
+
+        Most vacation periods are rule-based and can be calculated
+        algorithmically, therefore they are implemented here.
+        """
+        year = self._year
+        easter = self._easter_sunday
+
+        # Christmas holidays.
```
KJhellico:
```suggestion
        # Christmas Holidays.
```
Pay closer attention to details, it's important.
And I find it useful to leave a short description of the rules for each vacation period before the corresponding block, as was done in your previous editions.

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+
+        # Previous year's Christmas (for January spillover).
+        prev_christmas = self._christmas_day.replace(year=self._year - 1)
+        prev_christmas_start = _get_nth_weekday_from(
+            1 if self._is_weekend(prev_christmas) else -1,
+            MON,
+            prev_christmas,
+        )
+        duration = 14 - (32 - prev_christmas_start.day)
+        self._add_multiday_holiday(date(self._year, JAN, 1), duration, name=name)
```
KJhellico:
```suggestion
        new_year = date(self._year, JAN, 1)
        christmas_last_part_duration = (4 - new_year.weekday()) % 7 + 3
        self._add_multiday_holiday(new_year, christmas_last_part_duration, name=name)
```
The logic is a little confusing, but it works. ;)

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+
+        # Carnival holidays.
+        carnival_raw = _timedelta(easter, days=-47)
+        carnival_start = _get_nth_weekday_from(-1, MON, carnival_raw)
+        self._add_multiday_holiday(carnival_start, 7, name=tr("Krokusvakantie"))
```
KJhellico:
```suggestion
        # Carnival Holidays: Carnival Monday (Easter - 48 days) → 1 week.

        # Carnival Holidays.
        self._add_multiday_holiday(_timedelta(easter, -48), 7, name=tr("Krokusvakantie"))
```
You missed it.

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+            easter_duration = 15
+        self._add_multiday_holiday(easter_start, easter_duration, name=tr("Paasvakantie"))
+
+        # Labor Day.
+        self._add_labor_day(tr("Dag van de Arbeid"))
+
+        # Ascension Day.
+        self._add_ascension_thursday(tr("O. L. H. Hemelvaart"))
+
+        # Friday after Ascension.
```
KJhellico:
```suggestion
        # Friday after Ascension Day.
```
❗

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+        self._add_holiday_40_days_past_easter(tr("Vrijdag na O. L. H. Hemelvaart"))
+
+        # Whit Monday.
+        self._add_whit_monday(tr("Pinkstermaandag"))
+
+        # Summer holidays.
```
KJhellico:
```suggestion
        # Summer Holidays.
```

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+        self._add_multiday_holiday(date(year, JUL, 1), 62, name=tr("Zomervakantie"))
+
+        # Autumn holidays.
+        nov_1 = date(year, NOV, 1)
+        autumn_start = _get_nth_weekday_from(1 if self._is_sunday(nov_1) else -1, MON, nov_1)
+        self._add_multiday_holiday(autumn_start, 7, name=tr("Herfstvakantie"))
```
KJhellico, line 212:
```suggestion
        # Autumn Holidays.
        self._add_multiday_holiday(autumn_start, 7, name=tr("Herfstvakantie"))
```
All of these are l10n comments.

holidays/countries/belgium.py
```diff
@@ -84,6 +119,222 @@ def _populate_bank_holidays(self):
...
+
+        # Armistice Day.
+        self._add_remembrance_day(tr("Wapenstilstand"))
+
+    def _populate_subdiv_wbr_school_holidays(self):
```
KJhellico, line 217: In 2022, the school year schedule in the French-speaking community has changed, which must be considered.

Upanshi-Mittal:
Hi! I was working on the WBR (Fédération Wallonie-Bruxelles) school holidays and noticed something after the 2022 reform. It doesn’t seem to follow a simple relation with Easter anymore. Carnival and Spring holidays don’t match a fixed offset like “Easter - 48 days” for all years (for example 2025), so it feels more like a cycle-based system (7 weeks school + 2 weeks break).

Also, older data (pre-2022) isn’t very consistent, so it’s hard to derive a single formula that works everywhere.
What do you think would be better here:

1. Using a cycle-based approach, or
2. Keeping predefined dates for recent years?

I can proceed with either approach.

KJhellico:
> older data (pre-2022) isn’t very consistent, so it’s hard to derive a single formula that works everywhere

Can you give an example?

Upanshi-Mittal:
> Can you give an example?

for example if we try something simple like “Carnival = Easter - 48 days”, it works for some years but not all.
Like in 2025, Easter is on April 20. If we subtract 48 days, we get March 3, but the actual Carnival holiday starts on February 24.
So this kind of fixed formula doesn’t always match the real dates.

And when I checked older (pre-2022) data, the holiday positions also don’t follow a consistent pattern with respect to Easter, so it’s hard to come up with one rule that works for every year.

KJhellico:
> Like in 2025

But the rules have changed since 2022!

KJhellico:
@Upanshi-Mittal, are you planning to finish this PR?

Upanshi-Mittal:
Yes, I’m finishing this PR today. I’ve updated the holiday logic, added test coverage for school holidays, ensured category isolation, and resolved all linting issues. I’ll push the latest changes shortly.

KJhellico:
I've updated the code for Flemish.
For American English, the term "Break" is more natural for these types of holidays, so it's best to use it (as has already been done for Germany) I've also standardized the style of comments describing each type of holiday; I recommend following it.
I haven’t looked at the code for other subdivisions yet, but the lack of a difference before/after 2022 for WBR is noticeable.

holidays/countries/belgium.py
```diff
@@ -84,6 +118,276 @@ def _populate_bank_holidays(self):
...
+        # The start date depends on the date of Easter:
+        # - If Easter falls in March, the break starts on the Monday before Easter.
+        # - If Easter falls on or after April 16, the break starts 13 days before Easter.
+        # - Otherwise, the break starts on the first Monday of April.
+        # The break normally lasts two weeks (14 days), extended to include Easter Monday
+        # when necessary.
+        easter_start = _get_nth_weekday_of_month(1, MON, APR, year)
+        easter_duration = 14
+        if easter.month == 3:
+            easter_start = _timedelta(easter, -6)
```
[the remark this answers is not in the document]
KJhellico: @cubic-dev-ai, however, you were right!

Upanshi-Mittal:
I noticed that after removing the category filter, some of the existing tests specific holidays are failing. It seems those tests rely on the holidays being restricted to Public -only categories.

Should we adjust the tests to account for mixed-category behavior, or is there a preferred way to handle Public-only holiday population in this case?

KJhellico:
I don't review tests until the main code is finished or nearly finished.

PR #3135, opened 2025-12-15
Update United States holidays: update Puerto Rico holidays

KJhellico:
## Proposed change

Update United States holidays:
- update Puerto Rico holidays
- remove GOVERNMENT category from United States territories with their own ISO 3166-1 codes

## Type of change

- [ ] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [x] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] I've run `make check` locally; all checks and tests passed.

PR #3201, opened 2026-01-08
Update README: update Available Countries table

PPsyrius:
## Proposed change

Removing redundant entry for U.S. Virgin Islands not handled earlier in #3135

## Type of change

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1804,7 +1804,7 @@ any) in brackets, available languages and additional holiday categories. All cou
 <td></td>
 <td>See Virgin Islands (U.S.)</td>
 <td></td>
-<td>GOVERNMENT, UNOFFICIAL</td>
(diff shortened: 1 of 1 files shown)


arkid15r (commented):
It looks like we need some tests.

PPsyrius:
> It looks like we need some tests.

AFAIK, it's the only odd one out at the moment and is already included as "Virgin Islands (U.S.)" as well. The reason it wasn't tested before was that it doesn't have an entity code (and thus wasn't covered by our existing test cases) - IMO, removing this particular table entry altogether might be better. 👀

KJhellico:
> emoving this particular table entry altogether might be better

Yes, I also don't understand why it's here, without a code.

KJhellico (approved):
LGTM.

PR #3218, opened 2026-01-17
Add archive-links Makefile target

arkid15r:
## Proposed change

Add archive-links Makefile target

Archive some URLs.

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,10 @@
+.DEFAULT_GOAL := help
+
+.PHONY: help
+
 help:
 	@echo "Usage: make <target>"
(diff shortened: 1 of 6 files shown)


PPsyrius (approved):
LGTM 🛠️

twilight93:
I think it would be worth adding an option to run the archiver on a single .py file (e.g., `python archive_links.py
countries/united_states.py`) instead of scanning the entire library every time. It would be especially helpful when we are just adding or updating a specific country and want to verify those references quickly.

Just a thought!

PR #3221, opened 2026-01-18
Update WM archiver script: add optional target parameter

### LINKED ISSUE #3220: Add an option to run the links archiver on a single .py file
I think it would be worth adding an option to run the archiver on a single .py file (e.g., `python archive_links.py
countries/united_states.py`) instead of scanning the entire library every time. It would be especially helpful when we are just adding or updating a specific country and want to verify those references quickly.

Just a thought!

_Originally posted by @twilight93 in https://github.com/vacanza/holidays/issues/3218#issuecomment-3762708187_

pareshjoshij:
## Proposed change

Closes #3220

I've updated `scripts/archive_links.py` to accept an optional `path` argument.

Right now, the script scans the whole `holidays` folder by default, which can take a while if we just want to verify references for a specific country file. With this update, you can point the archiver to a single file or sub-directory (e.g., `python scripts/archive_links.py holidays/countries/estonia.py`). If no path is provided, it still defaults to scanning everything, so existing workflows won't break.

**Note on rate limits:** I skipped adding a sleep timer to prioritize speed. If you'd prefer I add a delay to avoid potential API 429 errors, just let me know and I'll add it.

## Type of change

## Checklist


scripts/archive_links.py
```diff
@@ -405,6 +405,13 @@ def main() -> None:
         description="Find, check/archive via Wayback Machine, and replace URLs in project files.",
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
+    # New positional argument for specific file or directory
```
KJhellico: AI comments?

pareshjoshij: No, it isn't , However, if it doesn't meet our repo standards, I am happy to change or remove it

KJhellico: Then I wonder what "New" here means to the average reader of this code? %)

pareshjoshij: That's a fair point. It's only new right now in this PR

scripts/archive_links.py
```diff
@@ -430,8 +436,25 @@ def main() -> None:
...
     start_time_total = time.time()
-    directory = os.path.abspath(os.path.join(__file__, "..", "..", "holidays"))
-    files_to_urls_data, _ = scan_directory_for_links(directory, EXTENSIONS_TO_SCAN)
+
+    files_to_urls_data: dict[str, list[str]] = {}
+
+    if args.path:
+        target_path = os.path.abspath(args.path)
+        if os.path.isfile(target_path):
+            print(f"Targeting single file: {target_path}")
```
KJhellico:
```suggestion
```

scripts/archive_links.py
```diff
@@ -430,8 +436,25 @@ def main() -> None:
...
+            urls = find_hyperlinks_in_file(target_path)
+            if urls:
+                files_to_urls_data = {target_path: urls}
+        elif os.path.isdir(target_path):
+            files_to_urls_data, _ = scan_directory_for_links(target_path, EXTENSIONS_TO_SCAN)
+        else:
+            print(f"Error: Path not found: {target_path}", file=sys.stderr)
+            sys.exit(1)
+    else:
+        # Default behavior: Scan the standard holidays library location
```
KJhellico:
```suggestion
```

KJhellico:
Actually, this script needs more refactoring later (use `pathlib.Path`, class-based approach).

pareshjoshij:
@KJhellico sir, I’ve removed the extra comments and print statements !

KJhellico (dismissed):
LGTM.

PPsyrius (dismissed):
LGTM 🛠️ 

I tried this script earlier and it works perfectly

KJhellico (approved):

PPsyrius (approved):

PR #3226, opened 2026-01-18
Add test ensuring no duplicate holiday dates across a year

VKofficial:
## Proposed change

This PR adds a logic-level test asserting that requesting holidays across
multiple years does not produce duplicate date entries.

The test protects against subtle aggregation or caching regressions when
holiday data is generated per-year and merged into a single result.
No runtime behavior is changed.

## Type of change

## Checklist

- [x] I've run the relevant tests locally; all checks and tests passed.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -73,6 +73,25 @@ class TestCountryHolidays(unittest.TestCase):
             CountryHoliday("IT")
         self.assertIn("CountryHoliday is deprecated", str(ctx.warning))
 
+class TestNoDuplicateDatesAcrossYears(unittest.TestCase):
+    """Ensure no duplicate holiday dates are produced when multiple years are requested."""
+
+    @pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


PPsyrius (commented):
Given that:

- The generated results are already a subclass of Python `dict`, the only case there could be duplicate entries are when multiple supported categories (i.e. Hong Kong, Macau) or multiple subdivisions (i.e. various U.S. states & territories) are requested together - which it already deduplicates should the holiday entries in question share the same name.

- The generated results don't go across a Gregorian year limit (i.e. observed holidays for Thailand's New Year's Eve won't give you next year's result automatically unless the next year's data was also requested).

IMO this test is probably not required or overlapping with an existing one, but let's wait for @arkid15r and @KJhellico 's input.

KJhellico:
Yes, I completely agree - I don't see any possibility of duplicating `dict` keys.

arkid15r:
@VKofficial thanks for the PR
Please consider opening an issue to discuss first and potentially avoid unnecessary/duplicate work.

PR #3228, opened 2026-01-19
Update Brunei holidays: observed holidays, 2026 exact dates

KJhellico:
## Proposed change

Update Brunei holidays:
- observed holidays calculation
- Islamic holidays 2026 exact dates

## Type of change

## Checklist


holidays/countries/brunei.py
```diff
@@ -38,13 +44,12 @@
...
-    FRI_TO_NEXT_MON,
-    FRI_TO_NEXT_SAT,
-    SUN_TO_NEXT_TUE,
-    SUN_TO_NEXT_WED,
-    FRI_SUN_TO_NEXT_SAT_MON,
+    ObservedRule,
+    FRI_SUN_TO_NEXT_WORKDAY,
 )
 
+BN_WORKDAY_TO_NEXT_WORKDAY = ObservedRule({MON: +7, TUE: +7, WED: +7, THU: +7, SAT: +7})
```
arkid15r, line 51: We use `_RULE` suffix sometimes. Let's figure out our preferred approach for custom rules later.

arkid15r (commented):
LGTM

PPsyrius (approved):
LGTM 🇧🇳

holidays/countries/brunei.py
KJhellico: I named it based on common rule `WORKDAY_TO_NEXT_WORKDAY`.

arkid15r: Yeah, I see your point and it's perfectly fine. What we need to decide is either add _RULE suffix for all custom/per-country rules or get rid of it entirely and have a standardized naming.

PR #3244, opened 2026-01-30
Update Germany holidays: add catholic holidays to Augsburg

Estartu:
## Proposed change

There is a special subdivision for the city Augsburg. Augsburg is a catholic city. Therefore all Bavarian holidays for predominantly catholic regions must also be applied. 

## Type of change

## Checklist

### DIFF
--- a/holidays/countries/germany.py
+++ b/holidays/countries/germany.py
@@ -266,6 +266,7 @@ class Germany(HolidayBase, ChristianHolidays, InternationalHolidays, StaticHolid
 
     def _populate_subdiv_augsburg_public_holidays(self):
         self._populate_subdiv_by_public_holidays()
+        self._populate_subdiv_by_catholic_holidays()
 
         # Augsburg Peace Festival.
         self._add_holiday_aug_8(tr("Augsburger Hohes Friedensfest"))
--- a/holidays/countries/germany.py
+++ b/holidays/countries/germany.py
@@ -266,7 +266,9 @@ class Germany(HolidayBase, ChristianHolidays, InternationalHolidays, StaticHolid
 
     def _populate_subdiv_augsburg_public_holidays(self):
         self._populate_subdiv_by_public_holidays()
-        self._populate_subdiv_by_catholic_holidays()
+
+        # Assumption Day.
+        self._add_assumption_of_mary_day(tr("Mariä Himmelfahrt"))
 
         # Augsburg Peace Festival.
         self._add_holiday_aug_8(tr("Augsburger Hohes Friedensfest"))
--- a/tests/countries/test_germany.py
+++ b/tests/countries/test_germany.py
@@ -232,7 +232,7 @@ class TestGermany(CommonCountryTests, TestCase):
         self.assertNoHolidayName(name)
         for subdiv, holidays in self.subdiv_holidays.items():
             # Saarland.
-            if subdiv == "SL":
+            if subdiv in ("SL", "Augsburg"):
                 self.assertHolidayName(
                     name, holidays, (f"{year}-08-15" for year in self.full_range)
                 )
--- a/tests/countries/test_germany.py
+++ b/tests/countries/test_germany.py
@@ -232,7 +232,7 @@ class TestGermany(CommonCountryTests, TestCase):
         self.assertNoHolidayName(name)
         for subdiv, holidays in self.subdiv_holidays.items():
             # Saarland.
-            if subdiv in ("SL", "Augsburg"):
+            if subdiv in {"SL", "Augsburg"}:
                 self.assertHolidayName(
                     name, holidays, (f"{year}-08-15" for year in self.full_range)
                 )
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -56,6 +56,7 @@ Filip Bednárik
 Firas Kafri
 Gabriel L Martinez
 Gabriel Trabanco
+Gerhard Schmidt
 Giedrius Mauza
 Gordon Inggs
 Greg Rafferty


holidays/countries/germany.py
```diff
@@ -266,6 +266,7 @@ def _populate_subdiv_th_catholic_holidays(self):
 
     def _populate_subdiv_augsburg_public_holidays(self):
         self._populate_subdiv_by_public_holidays()
+        self._populate_subdiv_by_catholic_holidays()
```
PPsyrius: Maybe something like this instead since `CATHOLIC` holidays should not be included in Augsburg's `PUBLIC` category list, but move `_populate_subdiv_augsburg_catholic_holidays` after `_populate_subdiv_augsburg_public_holidays`
```suggestion
    def _populate_subdiv_augsburg_catholic_holidays(self):
        self._populate_subdiv_by_catholic_holidays()

    def _populate_subdiv_augsburg_public_holidays(self):
        self._populate_subdiv_by_public_holidays()
```

tests/countries/test_germany.py
```diff
@@ -232,7 +232,7 @@ def test_assumption_day(self):
         self.assertNoHolidayName(name)
         for subdiv, holidays in self.subdiv_holidays.items():
             # Saarland.
-            if subdiv == "SL":
+            if subdiv in ("SL", "Augsburg"):
```
PPsyrius: Add this after Bayern/Bavaria ones below instead
```python
self.assertSubdivAugsburgCatholicHolidayName(name, (f"{year}-08-15" for year in self.full_range))
```

PPsyrius (commented):
Feel free to add your name to the [`CONTRIBUTORS`](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTORS) file as well 👍

holidays/countries/germany.py
Estartu: I beg differ. The catholic category is there because some some catholic holidays are only in some regions of Bavaria. There is no region in Augsburg that is not catholic. So in the whole city of Augsburg all catholic holidays apply.

PPsyrius: Good catch 👍

PPsyrius (dismissed):
LGTM 🇩🇪

holidays/countries/germany.py
KJhellico:
```suggestion

        # Assumption Day.
        self._add_assumption_of_mary_day(tr("Mariä Himmelfahrt"))
```
I think it's better to specify this holiday explicitly than to call `..._catholic_...` from `..._public_...`.

Estartu: right now it's only one holiday. But if others are added to the catholic holidays they will also apply to Augsburg, because Augsburg is a catholic city and all holidays for predominantly catholic regions in Bavaria apply to Augsburg as well. 

So adding all from the catholic category reduces redundancy.

KJhellico: I'm considering this from the perspective of implementing the concepts outlined in #1712 in v1, where the Holiday instance will contain information about its category.

Estartu: Adding it to the Augsburg subdivision would make it a city holiday like the "Hohes Friedensfest" which exists only in Augsburg. 

But it's not. It's still a holiday for predominantly catholic regions in Bavaria and should be categorized as such. 

it's added to the Augsburg subdivision because the whole area of Augsburg is catholic and therefore gets all catholic holidays in Bavaria as well.

KJhellico: > Adding it to the Augsburg subdivision would make it a city holiday like the "Hohes Friedensfest" which exists only in Augsburg.

No, just as its presence in `_populate_subdiv_sl_public_holidays` doesn't make it a "holiday which exists only in Saarland".

Estartu: But its marked as a holiday by the law of the subdivision SL (Saarland) 

Augsburg's "Hohes Friedensfest" is a holiday because it's stated so in the City Laws of Augsburg. 

"Mariä Himmelfahrt" is a holidays because Bavarian Laws states that it is in any region in Bavaria that is predominantly catholic. Augsburg is such a region and therefore get the Holiday.

In `_populate_subdiv_augsburg_public_holidays` only contain holidays that are specified by the subdivision laws (here city laws) 

in `_populate_subdiv_by_catholic_holidays` contains all holidays that are for all predominantly catholic regions in Bavaria as stated in Bavarian law. 

in `_populate_subdiv_by_public_holidays` contains all holidays specified for all regions stated by Bavarian law. 

Following your argumentation, `_populate_subdiv_by_public_holidays` shouldn't be included as well and all the holidays added to `_populate_subdiv_augsburg_public_holidays`

 If we add the single holiday, we lose the information, it's there because Augsburg is a catholic region and not because the city declared it a holiday.

KJhellico: > Following your argumentation, `_populate_subdiv_by_public_holidays` shouldn't be included as well and all the holidays added to `_populate_subdiv_augsburg_public_holidays`

No, this is consistent with "categories" concept: _public_ and _public_.

Estartu: It's a public holiday in Augsburg. But it's public there because Augsburg is a catholic region.

The category catholic it not there to categorize it as catholic holiday but categorize it as only apply to catholic regions. 

I think the categories are abused here. Maybe there should be two subdivisions for Bavaria:  BY and BY_Catholic. 
or some sub sub divisions BY/protestant and BY/catholic. Than Augsburg would include only `_populate_subdiv_by_catholic_public_holidays`. 
 
I think your suggestion doesn't add the concepts outlined in #1712 because the category catholic isn't really a category as discussed there. Many of the holidays specified in the public category can also be specified as catholic and as protestant and as christian and as religious.  

I think your suggestion add the risk of future errors. If somebody adds or removes a holiday that are only in catholic regions (There are discussions to re add the "Josefstag" (march 19th)) might miss to add them to Augsburg because Augsburg only really maters to the people living or working there. Including all the catholic holidays will include it automatically.

Estartu: Sorry. Your Version fails the SonarCube test. How do we fix this?

KJhellico: That's ok.

tests/countries/test_germany.py
KJhellico:
```suggestion
            if subdiv in {"SL", "Augsburg"}:
```

KJhellico (approved):
LGTM.

PPsyrius (approved):

arkid15r (commented):
LGTM, thanks @Estartu 👍

Issue #2739, opened 2025-07-21
Revisit Peru holidays in full and bring it up to modern implemented standard


PPsyrius:
While Peru holidays list itself should be more or less up-to-date thanks to PR #2730, its current implementation is very barebone compared to other updated entries like [Argentina's](https://github.com/vacanza/holidays/blob/58c2d08c0d4a0d99077d085c522ced8510229606/holidays/countries/argentina.py#L33) and [Brazil's](https://github.com/vacanza/holidays/blob/58c2d08c0d4a0d99077d085c522ced8510229606/holidays/countries/brazil.py#L22) i.e. there's no `start_year` parameter yet or historical implementations, or that its test cases coverage could use some improvements i.e. #2720 or #2737 

( Issue initially raised by @KJhellico  )

Usable Sources:
- https://www.peruweek.pe/antecedentes-de-los-feriados-en-el-peru/
- [Ley N° 10908](https://docs.peru.justia.com/federales/decretos-leyes/10908-dec-3-1948.pdf) (There's the earlier [Ley N° 3010](https://www.leyes.congreso.gob.pe/Documentos/Leyes/03010.pdf) as well but I can't find the specific list of "Civic Holidays" (Fiesta Cívica) mentioned)
- https://busquedas.elperuano.pe/ (AFAIK this archive only got law listing in full from Jan 1996 onwards, need alternative source for older ones)
- https://lpderecho.pe/ (we already use this source for our 2 new holidays from 2023)

virendiran:
Hi @PPsyrius I want to work on this issue if still available, kindly assign me for this issue and i will update you soon.
FYI, while drafting proposal for woc 5.0 I noticed this issue.

KJhellico:
@virendiran, please note: this issue requires more research into sources than code writing.

aniketgit-hub101:
Hi, I'm a GSoC 2026 applicant interested in working on this. I'd like to tackle the Peru holidays refactor as my first contribution. Is this still available?

HarshYadav1711:
Hello there mentors,
I am Harsh Yadav from India and one of the contributors for GSOC 2026. I am experienced and proficient in Python, C, C++, MERN Stack, Databases, react native, Android and iOS dev etc. and have a good amount of experience working on projects in these domains. I would love to contribute to holidays and gain more experience with the open-source scene. Looking forward to working with you. Kindly guide me on how to start.

HarshYadav1711:
[...]

PR #3332, opened 2026-03-16
Add holidays for Peru (PE)

### LINKED ISSUE #1234: Add Nepal holidays
Links:
  - https://en.wikipedia.org/wiki/Public_holidays_in_Nepal
  - https://www.timeanddate.com/holidays/nepal/

mrudu06:
This PR adds the 16 national public holidays for Peru for 2026, alongside their unit tests. It also registers Peru in the global and country namespaces. All tests and pre-commit checks pass.

Resolves #1234


holidays/countries/peru.py
```diff
@@ -17,81 +17,69 @@
...
 
 class Peru(HolidayBase, ChristianHolidays, InternationalHolidays):
-    """Peru holidays.
-
+    """
     References:
-        * <https://web.archive.org/web/20250414165243/https://www.gob.pe/feriados/>
-        * <https://es.wikipedia.org/wiki/Anexo:Días_feriados_en_el_Perú>
-        * [Ley N° 31788](https://web.archive.org/web/20250716164223/https://img.lpderecho.pe/wp-content/uploads/2023/06/Ley-31788-LPDerecho.pdf)
-        * [Ley N° 31822](https://web.archive.org/web/20250716164455/https://img.lpderecho.pe/wp-content/uploads/2023/07/Ley-31822-LPDerecho.pdf)
```
PPsyrius, line 26: Why are we suddenly replacing all relevant references with a single Wikipedia one and removing a supported language?

Not to mention the issue number referenced here is likely a placeholder one

PPsyrius (changes requested):

KJhellico:
Perhaps this PR was referring to issue #2739? But in any case, the current version is unacceptable.

PR #3348, opened 2026-03-19
Refactor EntityLoader: add namespace validation

RinZ27:
Lazy loading mechanism in registry.py utilizes importlib.import_module to resolve holiday entities dynamically based on a provided path. While these paths are currently managed internally, the lack of explicit scope validation poses a risk of importing modules outside the intended holidays namespace if the registry logic is ever extended or modified incorrectly.\n\nTransitioning to a strict prefix check in EntityLoader.__init__ ensures that only modules within the holidays package can be loaded. This improvement follows defensive programming principles, making the registry system more robust against unexpected input and ensuring clear boundaries for dynamic imports without affecting the existing lazy-loading performance.

### DIFF
--- a/holidays/registry.py
+++ b/holidays/registry.py
@@ -318,6 +318,12 @@ class EntityLoader:
                 "`from holidays.financial import Entity`."
             )
 
+        if not path.startswith("holidays."):
+            raise ValueError(
+                f"Invalid entity path: {path}. "
+                "Entities must be in the holidays namespace."
+            )
+
         entity_path = path.split(".")
 
         self.entity = None
--- a/holidays/registry.py
+++ b/holidays/registry.py
@@ -308,16 +308,7 @@ class EntityLoader:
     __slots__ = ("entity", "entity_name", "module_name")
 
     def __init__(self, path: str, *args, **kwargs) -> None:
-        """
-        Set up a lazy loader.
-
-        :param path:
-            The path to the entity to be lazy-loaded.
-        :param args:
-            Additional arguments.
-        :param kwargs:
-            Additional keyword arguments.
-        """
+        """Set up a lazy loader."""
         if args:
             raise TypeError(
                 "This is a holidays entity loader class. "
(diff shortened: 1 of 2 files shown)


PPsyrius (commented):
You might want to add a test case to simulate the issue as well to fix test coverage issue

holidays/registry.py
```diff
@@ -308,7 +308,16 @@ class EntityLoader:
     __slots__ = ("entity", "entity_name", "module_name")
 
     def __init__(self, path: str, *args, **kwargs) -> None:
-        """Set up a lazy loader."""
+        """
```
KJhellico: These changes are unnecessary.

RinZ27: @KJhellico Reverted the docstring expansion and the namespace validation check, also removed the corresponding test case to keep the changes minimal

KJhellico: Seems you actually revert _all_ your changes, completely.

RinZ27: @KJhellico Apologies for the confusion. I might have over-interpreted your previous comment. Since the core goal was adding namespace validation, should I restore just that specific check (without the docstring expansion) or would you prefer to close this PR if you feel any validation here is redundant? I'd like to ensure the changes are strictly aligned with your vision for the registry.

KJhellico: > These changes are unnecessary.

This refers to docstring change.

RinZ27: @KJhellico Restored the namespace validation check and the test case. I skipped the docstring changes this time to keep things focused on the core logic

holidays/registry.py
```diff
@@ -5,8 +5,8 @@
 #  specific date is a holiday as fast and flexible as possible.
 #
 #  Authors: Vacanza Team and individual contributors (see CONTRIBUTORS file)
-#           dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
```
KJhellico, line 8: There's something wrong with the formatting here and below.

RinZ27: @KJhellico Fixed the unintended escaping (backslash triple quotes and Unicode-escaped brackets) in `registry.py` and the test file. The namespace validation check and its test case are now restored with clean formatting.

arkid15r (commented):
Hi @RinZ27 thanks for submitting this.

I checked the existing [use cases](https://github.com/vacanza/holidays/blob/d05ef3d2c1c227f9f32390adf537b69c4cc116ce/holidays/registry.py#L408) and it seems the check is a redundant one as it's always starts w/ `holidays.` atm. In general, defensing approach is good however in this case I'd prefer not it introduce an extra check that's not currently needed and highly unlikely to be in use in future.

PR #3357, opened 2026-03-21
Add holiday classification (PUBLIC, UNOFFICIAL, WORKDAY) for India

### LINKED ISSUE #3231: Adding multilingual in date feature
**Describe the bug**
we have official 22 languages in India officially. now its supported for en-us,hi,gu as per libraries ,we  can modify the country class (India) to read the new language, and we can extend it to full multilingual holiday calendar generator.Classification of holidays

**Expected behavior**
Holiday Classification
Dates are classified into:
Official holidays → government-declared public holidays
Unofficial holidays → cultural/regional holidays
Workdays → weekdays that are not holidays
Optional: categories like OPTIONAL, UNOFFICIAL, WORKDAY are correctly applied.
Multilingual Support-Holiday names are shown in multiple languages as per state Users can switch languages or see multiple columns at once. For unsupported holidays, fallback to English.
Optional: visual calendar (using Matplotlib) highlights holidays.

Aditya-ha11:
### Add holiday classification support for India calendar

This PR introduces a classification system for holidays in the India calendar.

---

### Features

* Classifies holidays into:

  * `PUBLIC` (official government holidays)
  * `UNOFFICIAL` (mapped from `OPTIONAL` / cultural holidays)
  * `WORKDAY` (non-holiday weekdays)

* Uses `__setitem__` interception to integrate without modifying existing holiday generation logic

* Stores metadata in `_holiday_meta`

* Adds helper method:

  ```python
  get_holiday_info(date)
  ```

---

### Workday logic

* Workdays are defined as weekdays (Mon–Fri) that are not holidays

---

### Design

* Context-based classification using `_current_category`
* Intercepts final holiday insertion to ensure compatibility with all helper methods
* Fully backward compatible
* No changes to existing holiday generation logic

---

### Example

```python
import holidays
from datetime import date
from holidays.constants import PUBLIC, OPTIONAL

ind = holidays.India(years=2024, categories=(PUBLIC, OPTIONAL))

ind.get_holiday_info(date(2024, 1, 26))

# {'type': 'public', 'name_key': 'Republic Day'}
```

---

Closes #3231


KJhellico:
The idea of keeping a category attribute for each holiday is reasonable, and it's one of the changes [planned for implementation in v1](https://github.com/vacanza/holidays/issues/1712). If we are going to try to implement it within the existing data model, it makes sense to do so globally rather than on a country basis.

Aditya-ha11:
Thanks for the clarification!

That makes sense — implementing this at a global level is the right approach rather than within a single country class.

I’ll avoid introducing a country-specific implementation and will explore designing this at the framework level instead.

Would it be helpful if I open a separate PR or proposal for a global implementation of holiday categories?

KJhellico:
It seems to me that implementing this within the current data model would unnecessarily complicate the code (in fact, HolidayBase would need to maintain a secondary data set in addition to the main one and update it whenever changes are made).

shaonahamed0:
@KJhellico Should I try to understand this issue by going through the commit history of the PR and then attempt to make changes or solve it step by step?

KJhellico:
@shaonahamed0, I didn't quite understand your question.

shaonahamed0:
> @shaonahamed0, I didn't quite understand your question.

I mentioned that the person who opened the PR doesn’t seem to be actively working on it anymore. I’m thinking of reviewing what they’ve already done or intended to do, understanding their approach, and then trying to complete the remaining work myself.

KJhellico:
As I mentioned above, I doubt that it would be reasonable to implement this before the transition to v1.

PR #3398, opened 2026-03-24
add Hindi (hi_IN) translation file

bhargavikunduru8387-star:
Added IN.po file for Hindi (hi_IN) translations.
Includes basic holiday translations.


PPsyrius (commented):
But the current Hindi l10n support for India is `hi`, not `hi_IN` (and does already exists) - I'm not sure what this PR is trying to achieve here

KJhellico:
Hindi localization is already supported (as `hi` language).

PR #3403, opened 2026-03-24
Add Shanghai Stock Exchange (SSE) holidays

mengChengxi:
## Proposed change

Add Shanghai Stock Exchange market holidays support.

- adds ShanghaiStockExchange` with XSHG as the primary MIC and SSE as an alias.
- keeps the implementation minimal by reusing `China` holiday logic and limiting the market start year to 2001.
- adds XSHG localization stubs so supported locales fall back cleanly to the parent translations.
- adds tests for market metadata and localized 2024 holidays.
- adds the minimal README market table entry required by the docs checks.

References:
- https://english.sse.com.cn/start/trading/schedule/
- https://www.sse.com.cn/disclosure/announcement/general/c/c_20231226_5733939.shtml
- https://big5.sse.com.cn/site/cht/www.sse.com.cn/disclosure/announcement/general/c/c_20200127_4991582.shtml

## Type of change

## Checklist


holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,42 @@
...
+class ShanghaiStockExchange(ChildEntity, China):
+    """Shanghai Stock Exchange (SSE) holidays.
+
+    This class provides Shanghai Stock Exchange-specific market holidays.
+    Market holidays are days when the stock exchange is closed for trading.
+
+    References:
+        * <https://english.sse.com.cn/start/trading/schedule/>
+        * <https://www.sse.com.cn/disclosure/announcement/general/c/c_20231226_5733939.shtml>
+        * <https://big5.sse.com.cn/site/cht/www.sse.com.cn/disclosure/announcement/general/c/c_20200127_4991582.shtml>
```
KJhellico: It would be good to add links to closure announcements from the past few years (they seem to be available), similar to the one cited for 2024.

holidays/locale/en_US/LC_MESSAGES/XSHG.po
```diff
@@ -0,0 +1,29 @@
...
+#           dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
+#           ryanss <ryanssdev@icloud.com> (c) 2014-2017
+#  Website: https://github.com/vacanza/holidays
+#  License: MIT (see LICENSE file)
+#
+# Shanghai Stock Exchange (SSE) holidays en_US localization.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Holidays 0.81\n"
```
KJhellico:
```suggestion
"Project-Id-Version: Holidays 0.94\n"
```
(same for rest of .po files)

mengChengxi:
Thank you @KJhellico. I have addressed these 2 points.

- I added official SSE holiday schedule references for the past few years.
- Also, I corrected the version in all 4 files.

Are there any other problems I should address?

holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,47 @@
...
+
+    References:
+        * <https://english.sse.com.cn/start/trading/schedule/>
+        * <https://www.sse.com.cn/disclosure/announcement/general/c/c_20241223_10767108.shtml>
+        * <https://www.sse.com.cn/disclosure/announcement/general/c/c_20231226_5733939.shtml>
+        * <https://www.sse.com.cn/disclosure/announcement/general/c/c_20221227_5714458.shtml>
+        * <https://www.sse.com.cn/disclosure/announcement/general/c/c_20211220_5662606.shtml>
+        * <https://www.sse.com.cn/disclosure/announcement/general/c/c_20201224_5286949.shtml>
+        * <https://www.sse.com.cn/disclosure/announcement/general/c/c_20191220_4969627.shtml>
+        * <https://www.sse.com.cn/star/media/news/c/c_20200127_4991583.shtml>
```
KJhellico:
```suggestion
        * [2020](https://www.sse.com.cn/disclosure/announcement/general/c/c_20191220_4969627.shtml)
        * [2020 changes](https://www.sse.com.cn/star/media/news/c/c_20200127_4991583.shtml)
        * [2021](https://www.sse.com.cn/disclosure/announcement/general/c/c_20201224_5286949.shtml)
        * [2022](https://www.sse.com.cn/disclosure/announcement/general/c/c_20211220_5662606.shtml)
        * [2023](https://www.sse.com.cn/disclosure/announcement/general/c/c_20221227_5714458.shtml)
        * [2024](https://www.sse.com.cn/disclosure/announcement/general/c/c_20231226_5733939.shtml)
        * [2025](https://www.sse.com.cn/disclosure/announcement/general/c/c_20241223_10767108.shtml)
```
I recommend updating the links to WM archived versions (I think the easiest way would be to run `python scripts/archive_links.py holidays/financial/shanghai_stock_exchange.py`).

holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,47 @@
...
+#           ryanss <ryanssdev@icloud.com> (c) 2014-2017
+#  Website: https://github.com/vacanza/holidays
+#  License: MIT (see LICENSE file)
+
+from holidays.constants import PUBLIC
+from holidays.countries.china import China
+from holidays.mixins.child_entity import ChildEntity
+
+
+class ShanghaiStockExchange(ChildEntity, China):
```
KJhellico:
```suggestion


class ShanghaiStockExchange(China):
```
Same as changes in XJPX in #3405.

mengChengxi:
Thank you @KJhellico. I have addressed these 2 points.

- switched the SSE references to WM archived links and reformatted the yearly links
- updated `ShanghaiStockExchange` to inherit from `China`, matching the `XJPX` change in #3405

Are there any thing I can further improve? If it is ok, I am planning to work on HKSE holidays.

tests/financial/test_shanghai_stock_exchange.py
```diff
@@ -0,0 +1,131 @@
...
+class TestShanghaiStockExchange(CommonFinancialTests, TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass(ShanghaiStockExchange)
+
+    def test_code(self):
+        self.assertTrue(hasattr(self.holidays, "market"))
+        self.assertIsNone(getattr(self.holidays, "country", None))
+        self.assertEqual(self.holidays.start_year, 2001)
+        self.assertNoHoliday("2000-01-01")
```
KJhellico:
```suggestion
```

KJhellico (commented):
Last fix. And you can also add your name to CONTRIBUTORS file.

mengChengxi:
Thank you @KJhellico. I have addressed this point.

- I deleted these 2 lines and add my name to CONTRIBUTORS file.

Thank you again for reviewing, and I am very happy to contribute to this project.

KJhellico (dismissed):
LGTM.

arkid15r (commented):
Nice addition, thank you @mengChengxi !

mengChengxi:
Hello @arkid15r @KJhellico. The CI/CD test failed during the merge process. Sorry for the inconvenience. I found the cause.

The merge-queue failure on `#3403` came from inherited `China` localizations changing on the base branch after `#3411`.

`XSHG` currently falls back to `China` for those localized names, but my SSE test was asserting concrete localized strings directly. Once the inherited `China` translations changed, the SSE l10n test became stale in the queue environment.

When I was writing the test, I did not consider for future changes in the parent `China` localizations. I will be more careful about inherited behavior and test robustness in future changes.

I pushed a follow-up test fix to align the expected `XSHG` localized output with the current `China` translations instead of hardcoding those inherited strings.

Can you review them again? Thank you so much.

tests/financial/test_shanghai_stock_exchange.py
```diff
@@ -0,0 +1,70 @@
...
+            "2024-10-02",
+            "2024-10-03",
+            "2024-10-04",
+            "2024-10-07",
+        )
+        kwargs = {"years": 2024}
+        if language:
+            kwargs["language"] = language
+        holidays = China(**kwargs)
+        return tuple((dt, holidays.get(dt)) for dt in dates)
```
PPsyrius: I'm not yet clear on the new test format, so I'm leaving this to @arkid15r and @KJhellico 's discretion for now.

Interesting pattern nonetheless, we can probably reuse this for all the various French/US dependency entities elsewhere as well 👀

pareshjoshij:
@PPsyrius Just a quick thought while reviewing the implementation: won't inheriting directly from the standard `China` public holiday class cause some issues with the actual trading calendar?

For instance, in 2024, the SSE was officially closed on Feb 9 (Lunar NYE) and didn't observe the weekend makeup workdays on Feb 4 & 18. If we use the base class as-is, it seems it will flag those as open market days.

tests/financial/test_shanghai_stock_exchange.py
KJhellico: It’s an interesting idea. But there are different scenarios here: while this approach can be used for French territories (since they are an official subdivisions of the parent country, all their holiday implementation is actually included in the country code, and any changes must be made there), we need to be careful when it applies to financial markets. And this case is a good example: a change in China’s holidays forced us to take another look at the exchange’s holidays, and I believe it would be appropriate to do the same for any future changes to China’s holidays.

mengChengxi:
I checked the official SSE holiday notices more closely, and I agree the current `ShanghaiStockExchange(China)` implementation is too optimistic.

The pattern seems to be:

- SSE generally follows the China holiday schedule for the main closure ranges.
- However, national makeup work weekends do not become trading days on SSE. For example, the 2024 SSE notice keeps `2024-02-04` and `2024-02-18` as weekend closures, even though the State Council holiday arrangement treats them as workdays. The 2025 SSE notice does the same for `2025-01-26` and `2025-02-08`.
- 2024 also appears to require an explicit exchange-specific closure on Lunar New Year’s Eve (`2024-02-09`).

So I do not think the current thin wrapper over `China` is sufficient anymore. I will rework the implementation to keep SSE aligned with the official exchange notices instead of assuming the country calendar can be inherited unchanged.

Sources:
- SSE 2024 notice: https://www.sse.com.cn/disclosure/announcement/general/c/c_20231226_5733939.shtml
- SSE 2025 notice: https://www.sse.com.cn/disclosure/announcement/general/c/c_20241223_10767108.shtml
- State Council 2024 holiday arrangement: https://app.www.gov.cn/govdata/gov/202310/25/508678/article.html

mengChengxi:
Hello @pareshjoshij @arkid15r @KJhellico @PPsyrius I pushed a follow-up rework for SSE.

It now keeps China’s main holiday ranges as the base, but no longer inherits China `weekend_workdays`, so makeup-work weekends remain closed on SSE. I also added the explicit `2024-02-09` closure from the SSE 2024 notice and updated the tests accordingly.

holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,65 @@
...
+    parent_entity = China
+    supported_categories: tuple[str, ...] = (PUBLIC,)  # type: ignore[assignment]
+    start_year = 2001
+
+    def _populate_common_holidays(self):
+        super()._populate_common_holidays()
+
+        # SSE keeps national makeup work weekends closed and publishes its own
+        # holiday calendar instead of inheriting China weekend workdays.
+        self.weekend_workdays.clear()
```
KJhellico, line 60: You don't need to clear all `weekend_workdays`; it's enough to delete only the relevant values for 2024.

holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _populate_public_holidays(self):
+        super()._populate_public_holidays()
+
+        # The 2024 SSE trading schedule explicitly includes Lunar New Year's Eve
+        # as a market closure.
+        if self._year == 2024:
+            self._add_chinese_new_years_eve(tr("农历除夕"))
```
KJhellico: This should be done as special holidays.
And you'll need to update the .po files (`make l10n`).

pareshjoshij:
> Hello @pareshjoshij @arkid15r @KJhellico @PPsyrius I pushed a follow-up rework for SSE.

Ok ! BDW Overall, you did an excellent job  🦅

holidays/financial/shanghai_stock_exchange.py
mengChengxi: I checked the official SSE notices more broadly, and the pattern seems consistent beyond 2024.

In every verified SSE notice I checked from 2020 through 2026, national makeup-work weekends are still explicitly listed as weekend closures for the exchange rather than trading days. So from the exchange-calendar perspective, inheriting `China.weekend_workdays` does not seem correct as a general rule.

And 2024 is very special. That year's Chinese new years eve is not a national holiday, but the SSE was still closed.

KJhellico: > national makeup-work weekends are still explicitly listed as weekend closures for the exchange rather than trading days

Yes, you're right. For some reason, I was reading about non-working days but thinking about working days. 🤷‍♂️

mengChengxi:
I have moved Chinese new years eve in 2024 to special holidays and add its' translation to .po files.

holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class ShanghaiStockExchangeStaticHolidays(ChinaStaticHolidays):
+    substituted_date_format = ChinaStaticHolidays.substituted_date_format
+    substituted_label = ChinaStaticHolidays.substituted_label
+    special_public_holidays = {
+        **ChinaStaticHolidays.special_public_holidays,
+        2024: (
+            *ChinaStaticHolidays.special_public_holidays[2024],
+            (FEB, 9, tr("农历除夕")),
```
PPsyrius, line 77:
```suggestion
            *ChinaStaticHolidays.special_public_holidays[2024],
            #. Chinese New Year's Eve.
            (FEB, 9, tr("农历除夕")),
```

holidays/locale/en_US/LC_MESSAGES/XSHG.po
```diff
@@ -0,0 +1,32 @@
...
+"Language: en_US\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Lingua 4.15.0\n"
+"X-Generator: Poedit 3.4.2\n"
+"X-Source-Language: zh_CN\n"
+
+msgid "农历除夕"
+msgstr "Chinese New Year's Eve"
```
PPsyrius, line 33: The same goes for other `.po` files
```suggestion

#. Chinese New Year's Eve.
msgid "农历除夕"
msgstr "Chinese New Year's Eve"
```

tests/financial/test_shanghai_stock_exchange.py
```diff
@@ -0,0 +1,88 @@
...
+class TestShanghaiStockExchange(CommonFinancialTests, TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass(ShanghaiStockExchange)
+
+    def test_code(self):
+        self.assertTrue(hasattr(self.holidays, "market"))
+        self.assertIsNone(getattr(self.holidays, "country", None))
+
+    @staticmethod
```
PPsyrius: Let's test that 1 special holiday specific to SSE separately as well
```suggestion
    def test_special(self):
        self.assertHoliday("2024-02-09")

    @staticmethod
```

PPsyrius (commented):
Last fixes (unless other reviewers spot other market close days for SSE not include in the main China holidays calendar) 🎉

holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    country = None  # type: ignore[assignment]
+    market = "XSHG"
+    parent_entity = China
+    supported_categories: tuple[str, ...] = (PUBLIC,)  # type: ignore[assignment]
+    start_year = 2001
+
+    def __init__(self, *args, **kwargs):
+        ChineseCalendarHolidays.__init__(self)
```
KJhellico, line 48: You should use the existing members of the parent class instead of initializing everything from scratch.

holidays/financial/shanghai_stock_exchange.py
```diff
@@ -0,0 +1,81 @@
...
+
+class SSE(ShanghaiStockExchange):
+    pass
+
+
+class ShanghaiStockExchangeStaticHolidays(ChinaStaticHolidays):
+    substituted_date_format = ChinaStaticHolidays.substituted_date_format
+    substituted_label = ChinaStaticHolidays.substituted_label
+    special_public_holidays = {
+        **ChinaStaticHolidays.special_public_holidays,
```
KJhellico, line 73: I think it would be better to implement general support for extending the inherited `special_holidays` set with the corresponding data from the subclass.

tests/financial/test_shanghai_stock_exchange.py
```diff
@@ -0,0 +1,88 @@
...
+            )
+            for dt in dates
+        )
+
+    def test_makeup_work_weekends_remain_closed(self):
+        for dt in ("2024-02-04", "2024-02-18", "2025-01-26", "2025-02-08"):
+            self.assertNoHoliday(dt)
+            self.assertFalse(self.holidays.is_working_day(dt))
+
+    def test_l10n_default(self):
```
KJhellico, line 41: I believe a standard l10n tests implementation is necessary here.
https://github.com/vacanza/holidays/pull/3403#discussion_r2994727316

mengChengxi:
I looked into this more carefully.

Right now, `China.__init__()` initializes `StaticHolidays` with `ChinaStaticHolidays` directly. Because of that, `ShanghaiStockExchange` cannot just inherit the parent behavior and add its own `special_public_holidays` entry automatically. The SSE-specific `2024-02-09` closure would not be applied that way.

To implement general support, I am not sure if I understand you correctly, I think I need to change the framework-level `StaticHolidays` behavior. That looks more complex, and it may have side effects on other classes that already use `special_public_holidays`.

For this commit, I kept the narrower SSE-specific implementation because I was concerned that the broader framework change could affect other entities. If needed, I can look at that as a separate follow-up.

Other points are addressed.

tests/financial/test_shanghai_stock_exchange.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def test_code(self):
+        self.assertTrue(hasattr(self.holidays, "market"))
+        self.assertIsNone(getattr(self.holidays, "country", None))
+
+    def test_special(self):
+        self.assertHoliday("2024-02-09")
+
+    def test_makeup_work_weekends_remain_closed(self):
+        for dt in ("2024-02-04", "2024-02-18", "2025-01-26", "2025-02-08"):
```
KJhellico:
```suggestion
    def test_no_weekend_workdays(self):
        for dt in (
            "2024-02-04",
            "2024-02-18",
            "2025-01-26",
            "2025-02-08",
        ):
```

KJhellico:
> need to change the framework-level `StaticHolidays` behavior

That's right. So I agree - let's leave it at that for now. But we'll need to implement it later.

mengChengxi:
Thank you.
I have renamed the function and improved the coding style.

KJhellico (dismissed):
LGTM.

PPsyrius (dismissed):

PPsyrius (dismissed):

mengChengxi:
Sorry, I added my name to translator as required and it needs another approve.
Thank you

KJhellico (dismissed):

PPsyrius (approved):

arkid15r (commented):
@mengChengxi thanks for implementing this 👍

PR #3425, opened 2026-03-26
Add support for filtering only observed holidays

shaonahamed0:
## Proposed change

Currently, when `observed=True`, the API returns both statutory and observed holidays. This can feel like duplication for certain use cases where only observed dates are needed.

There is no built-in way to retrieve only observed holidays, requiring users to manually filter holiday names, which is not reliable.

This PR introduces an optional `only_observed` check in `_add_holiday()` to allow filtering only observed holidays.

- Uses `getattr` to safely handle cases where `observed_label` may not be defined
- Does not change existing behavior unless explicitly used
- Avoids manual string filtering at the user level

Would appreciate feedback on whether this aligns with the intended design.

---

## Type of change

---

## Checklist

### DIFF
--- a/holidays/holiday_base.py
+++ b/holidays/holiday_base.py
@@ -235,6 +235,7 @@ class HolidayBase(dict[date, str]):
         years: YearArg | None = None,
         expand: bool = True,
         observed: bool = True,
+        only_observed: bool = False,
         subdiv: str | None = None,
         prov: str | None = None,  # Deprecated.
         state: str | None = None,  # Deprecated.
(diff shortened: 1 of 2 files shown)


holidays/holiday_base.py
```diff
@@ -800,7 +802,10 @@ def _add_holiday(self, name: str, *args) -> date | None:
 
         if dt.year != self._year:
             return None
-
+        if getattr(self,"only_observed",False):
+            label = getattr(self,"observed_label",None)
```
KJhellico, line 806: This code doesn't do what you need.

PR #3429, opened 2026-03-26
pytest9 fix: use [tool.pytest.ini_options] instead of [tool.pytest] with dotted keys

hrnciar:
Pytest 9 rejects using [tool.pytest] with ini_options.* dotted keys. Use the standard [tool.pytest.ini_options] section instead.

Made-with: Cursor

## Proposed change

I am working on updating pytest to 9.x.x in Fedora and holidays fail. This PR addresses changes in pytest to make it compatible with 9, it also works with pytest 8.

## Type of change

- [x] Dependency update (version deprecation/pin/upgrade)

## Checklist

pyproject-fmf check failed and reverted my changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,14 +119,14 @@ overrides = [
   { module = "holidays.groups.*", disable_error_code = [ "attr-defined" ] },
 ]
 
-[tool.pytest]
-ini_options.addopts = [
+[tool.pytest.ini_options]
+addopts = [
   "--cov-fail-under=100",
 ]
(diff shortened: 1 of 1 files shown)


KJhellico:
> Pytest 9 rejects using [tool.pytest] with ini_options.* dotted keys

We currently use pytest 9.0.2, and the tests on Linux/Windows/macOS are working fine.

hrnciar:
You are right, I missed downstream only `sed` that modified `pyproject.toml` in a way new pytest didn't like. Sorry for the noise.

PR #3170, opened 2025-12-26
Add Bombay Stock Exchange (BSE) holidays

pareshjoshij:
## Proposed change

This PR adds support for the **Bombay Stock Exchange (BSE)**, market code `XBOM`.

Closes #3149

**Implementation Details:**
As discussed in the linked issue, the BSE shares an identical holiday calendar with the National Stock Exchange (NSE) as both are regulated by SEBI.

Following the feedback from @arkid15r regarding structural consistency and long-term perspective, I have implemented this as a **separate class** that inherits from `NationalStockExchangeOfIndia`, rather than just adding an alias. This ensures the correct directory structure while avoiding code duplication for the holiday logic.

## Type of change

- [x] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] I've run `make check` locally; all checks and tests passed.

PR #3204, opened 2026-01-09
Refactor PO generator for robust metadata and improved class discovery

pareshjoshij:
## Proposed change

Closes #3180

This PR refactors the `generate_po_files.py` script to significantly improve the robustness and standardization of our localization workflow.

* **Standardized Metadata:** Added `POT-Creation-Date`, `MIME-Version`, `Report-Msgid-Bugs-To`, and other standard gettext headers to ensure better compatibility with translation tools (like Weblate/Crowdin).
* **License Headers:** Added a mechanism to automatically inject license headers from `docs/file_header.txt` into generated `.po` files.
* **Descriptive Comments:** The script now extracts docstrings to add helpful comments (e.g., `# United States holidays.`) to the `.po` files.
* **Strict No-Change Policy:** Improved logic to minimize "noise" in git diffs by only updating timestamps if actual content has changed.

**Testing & Notes:**
* I have tested this locally by modifying `holidays/countries/india.py` (adding/changing holidays) and creating dummy blank `.po` files.
* I ran `make check` locally and all checks passed.
* **Draft Status:** I am opening this as a draft because I am aware this refactor may affect `scripts/l10n/l10n_helper.py`. I plan to make necessary changes there accordingly.
* I would appreciate feedback on whether this approach aligns with the project's coding style.

## Type of change

- [ ] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [x] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] I've run `make check` locally; all checks and tests passed.

PR #3453, opened 2026-03-30
Add Shenzhen Stock Exchange (SZSE) holidays

mengChengxi:
## Proposed change

Add Shenzhen Stock Exchange (SZSE) market holidays support.

The implementation follows the thin-wrapper approach used in #3170. SZSE publishes the same annual market-closure schedule as Shanghai Stock Exchange, so this PR reuses the Shanghai exchange calendar logic while adding SZSE-specific market codes, locale files, tests, registry entries, and README documentation.

## Type of change

## Checklist


tests/financial/test_shenzhen_stock_exchange.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def test_l10n_default(self):
+        holidays = self._get_holidays()
+        self.assertHolidayName("元旦", holidays, "2024-01-01")
+        self.assertHolidayName("农历除夕", holidays, "2024-02-09")
+        self.assertHolidayName("春节（补假）", holidays, "2024-02-13")
+        self.assertHolidayName("休息日（由 2024-02-04 调休）", holidays, "2024-02-15")
+        self.assertHolidayName("清明节", holidays, "2024-04-04")
+        self.assertHolidayName("中秋节", holidays, "2024-09-17")
+        self.assertHolidayName("国庆节", holidays, "2024-10-01")
```
PPsyrius: Please use the standardized formatting we did for SSE

tests/financial/test_shenzhen_stock_exchange.py
```diff
@@ -0,0 +1,75 @@
...
+        super().setUpClass(ShenzhenStockExchange)
+
+    @staticmethod
+    def _get_holidays(language=None):
+        return ShenzhenStockExchange(years=2024, language=language)
+
+    def test_code(self):
+        self.assertTrue(hasattr(self.holidays, "market"))
+        self.assertEqual(self.holidays.market, "XSHE")
+        self.assertIsNone(getattr(self.holidays, "country", None))
```
PPsyrius: We can probably ignore this test case as SZSE's parent class is SSE, also a `market` 
```suggestion
```

holidays/financial/shenzhen_stock_exchange.py
```diff
@@ -0,0 +1,40 @@
...
+    References:
+        * [2026](https://www.szse.cn/disclosure/notice/t20251222_618087.html)
+
+    Historical data:
+        * [2024](https://www.szse.cn/disclosure/notice/t20231226_605108.html)
+        * [2025](https://www.szse.cn/www/disclosure/notice/general/t20241223_611283.html)
+    """
+
+    market = "XSHE"
+    start_year = 2001
```
PPsyrius: `start_year` is the same as Shanghai's so we can probably left this out
```suggestion
    parent_entity = ShanghaiStockExchange
```

holidays/financial/shenzhen_stock_exchange.py
```diff
@@ -0,0 +1,40 @@
...
+    Shenzhen Stock Exchange publishes the same annual holiday schedule as
+    Shanghai Stock Exchange, so this class reuses the Shanghai calendar
+    implementation and provides SZSE-specific market codes and translations.
+
+    References:
+        * [2026](https://www.szse.cn/disclosure/notice/t20251222_618087.html)
+
+    Historical data:
+        * [2024](https://www.szse.cn/disclosure/notice/t20231226_605108.html)
+        * [2025](https://www.szse.cn/www/disclosure/notice/general/t20241223_611283.html)
```
PPsyrius: FYI in case Internet Archive's Wayback Machine cannot archive this, `archive.today` can be use for this as well
```suggestion
    References:
        * [2024](http://archive.today/2026.03.30-044431/https://www.szse.cn/disclosure/notice/t20231226_605108.html)
        * [2025](http://archive.today/2026.03.30-044217/https://www.szse.cn/www/disclosure/notice/general/t20241223_611283.html)
        * [2026](http://archive.today/2026.03.30-043931/https://www.szse.cn/disclosure/notice/t20251222_618087.html)
```

holidays/financial/shenzhen_stock_exchange.py
mengChengxi: I tested adding parent_entity = ShanghaiStockExchange, but there will be a bug:

With the current translation initialization logic, that changes the gettext fallback from `XSHE -> China` to `XSHE -> XSHG` only, and it does not continue recursively to `China`.Therefore, holidays in `en_US`, `th`, and `zh_TW` language will fall back to `XSHG` only and display Chinese rather than the translated language.

tests/financial/test_shenzhen_stock_exchange.py
mengChengxi: I don't think we can delete `self.assertEqual(self.holidays.market, "XSHE")`

holidays/financial/shenzhen_stock_exchange.py
KJhellico: Yes, the implementation did not foresee the possibility of more than two levels of inheritance. :)

mengChengxi:
Thank you @PPsyrius 
I have addressed these points except adding parent_entity.

mengChengxi:
Hello, maintainers.
I have updated my code, and are there any other possible improvements to it?
When you have time, could you please take another look?
Thank you.

tests/financial/test_shenzhen_stock_exchange.py
```diff
@@ -0,0 +1,132 @@
...
+from tests.common import CommonFinancialTests
+
+
+class TestShenzhenStockExchange(CommonFinancialTests, TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass(ShenzhenStockExchange)
+
+    def test_code(self):
+        self.assertEqual(self.holidays.market, "XSHE")
```
KJhellico:
```suggestion
        self.assertTrue(hasattr(self.holidays, "market"))
        self.assertIsNone(getattr(self.holidays, "country", None))
```

mengChengxi: Sorry, I may misunderstand your point. Do you mean keep 3 tests(add 2 more and keep the original one)? Because deleting self.assertEqual(self.holidays.market, "XSHE") will cause test to fail.

KJhellico: > will cause test to fail

Why?

mengChengxi: You are right — I mixed up replacing the assertion with removing the local override entirely.

Keeping the local test_code() override and changing it to your suggested checks passes. The failure only happens if the whole override is removed and the test falls back to CommonFinancialTests.test_code().
And I have update the code accordingly.

KJhellico (approved):
LGTM.

arkid15r (commented):
@mengChengxi thanks for adding this 👍

arkid15r:
It seems `make l10` isn't in place again @KJhellico @PPsyrius

PPsyrius:
~~Let's see if we can get the fix for that up alongside #3204 - that PR is almost ready IMO~~

Edit: NVM, those .PO files should've been blanked, will open a PR for them shortly

Issue #2344, opened 2025-03-09
Localize untranslated `.po` files


arkid15r:
### **Contribute Translations for Untranslated `.po` Files Based on Your Native Language Knowledge**  

**Description:**  
The project supports localized holiday names for over **150+ countries and regions**. However, many countries still have **untranslated holiday names** in their respective `.po` files, especially for local markets. This task invites contributors to **help translate these holidays** based on their **native language knowledge** to improve the project's localization (l10n) coverage.  

This is a high-impact task as it directly benefits users from different countries by ensuring they see holiday names in their native language.  

---

### **Which Languages Need Help?**  
Here are a few languages that still have **many untranslated holidays**:  
- 🇸🇦 Arabic (`ar.po`)  
- 🇧🇷 Brazilian Portuguese (`pt_BR.po`)  
- 🇨🇳 Chinese (`zh_CN.po`)  
- 🇩🇪 German (`de.po`)  
- 🇮🇳 Hindi (`hi.po`)  
- 🇮🇹 Italian (`it.po`)  
- 🇯🇵 Japanese (`ja.po`)  
- 🇫🇷 French (`fr.po`)  
- 🇪🇸 Spanish (`es.po`)  

You can also browse all languages in the `locales/` folder to see where your language knowledge could contribute.  

---

### **Acceptance Criteria**  
- All untranslated holidays in the `.po` file have been translated to the target native language.  
- Translations are accurate and appropriate for the region.  
- A pull request is submitted with clear information about the language and country you worked on.  
- Optional: If you find mistakes in existing translations, feel free to correct them in the same PR.  

---

### **Reference**  
Please refer to our localization contribution guidelines for detailed instructions:  
[https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.rst#localization](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.rst#localization)  

---

### **Why This Task Matters**  
By contributing translations, you are:  
- **Improving localization (l10n)** for users from your region.  
- Helping businesses, developers, and applications provide more accurate and localized holiday data.  
- Supporting an open-source project that serves millions of developers worldwide.  

If you have any questions or need clarification, feel free to comment on this issue or ask in our community channel.

⚠ https://github.com/vacanza/holidays/issues/1658

Nithin3003:
I wish to work on this issue.

arkid15r:
> I wish to work on this issue.

[...]

Issue #3152, opened 2025-12-19
Missing l10n for non-English official language countries


KJhellico:
Here are the countries that remain non-localized and have an official language other than English:

| Country       | Code | Official language(s) | Issue | Status |
|---------------|------|----------------------|------|--------|
| Bhutan        | BT   | Dzongkha (dz) | #3389 | ✅ |
| Burkina Faso  | BF   | French (fr) | #3175 | ✅ |
| Chad          | TD   | Arabic (ar), French (fr) |||
| Comoros       | KM   | Shikomoro (swb), Arabic (ar), French (fr) |||
| Eritrea       | ER   | Tigrinya (ti), Arabic (ar), English (en) |||
| Gabon         | GA   | French (fr) | #3176 | ✅ |
| Kyrgyzstan    | KG   | Kyrgyz (ky), Russian (ru) | #3301 | ✅ |
| Maldives      | MV   | Dhivehi (dv) | #3110 | ✅ |
| Mauritania    | MR   | Arabic (ar) | #3660 | ✅ |
| Somalia       | SO   | Somali (so), Arabic (ar) |||

aman-coder03:
can i work on this?

KJhellico:
@aman-coder03, Which countries are you interested in? Do you speak some of the languages listed?

aman-coder03:
Thanks for the info! I’d love to help with this. I can start by working on French (fr) localizations for:

Burkina Faso (BF)
Gabon (GA)
and where applicable Chad (TD) and Comoros (KM)

I don’t speak all the listed languages natively (e.g., Dzongkha, Tigrinya, Dhivehi, Somali, Kyrgyz, Russian, Arabic), but I can work from official sources and CLDR/local standards to add accurate localized strings, and the results can be reviewed by native speakers later.

If that plan sounds good, I’ll draft a PR with those French localizations first, then we can expand to the other languages/countries. Thanks! @KJhellico

KJhellico:
@aman-coder03, it's sounds pretty good. It would be better to keep this issue as a summary and open separate ones for each country.

aman-coder03:
I’ll start by opening individual issues for:
- **Burkina Faso (BF)** — French (fr)
- **Gabon (GA)** — French (fr)

and then proceed with PRs linked to those issues.

aman-coder03:
PR opened for Burkina Faso #3177

aman-coder03:
PR opened for Gabon #3178

[...]

PR #3472, opened 2026-03-31
l10n: add en_US translations for Ethiopia (ET)

ay-jay-9:
Adds English (en_US) translations for all untranslated msgstr entries 
in the Ethiopia (ET) locale file, as part of issue #2344.


PPsyrius:
@ay-jay-9 For country without l10n support yet - see https://github.com/vacanza/holidays/issues/3152

ay-jay-9:
> @ay-jay-9 For country without l10n support yet - see #3152

Thank you for the direction @PPsyrius! I'll look into #3152 and pick a country to work on.

PR #3493, opened 2026-04-02
Update dependabot configuration

arkid15r:
## Proposed change

Changes
 - 21-day cooldown for version updates on both the uv and github-actions ecosystems. New releases are only proposed after they have been out for three weeks. Cooldown applies to version updates only; it does not apply to Dependabot security updates.
 - Grouped version updates via a version-updates group with applies-to: version-updates and pattern '*', so routine dependency bumps are combined into one PR per ecosystem instead of many small ones. Security updates stay on their own PRs unless you add a separate security group later.
 - YAML style: top-level keys after package-ecosystem are ordered alphabetically (cooldown, directory, groups, schedule, target-branch); nested schedule keys use day then interval; string literals use single quotes.

## Type of change

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,15 +1,29 @@
 version: 2
 updates:
-  - package-ecosystem: uv
+  - package-ecosystem: github-actions
+    cooldown:
+      default-days: 21
     directory: /
+    groups:
+      version-updates:
+        applies-to: version-updates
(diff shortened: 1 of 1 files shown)


PPsyrius (approved):
LGTM 🛠️ 

This is pretty warranted considering all the recent supply-chain attacks

PR #3506, opened 2026-04-04
Update documentation build: proper notes rendering on GitHub and ReadTheDocs

KJhellico:
## Proposed change

ProperDocs admonitions are not supported by the GitHub renderer:
image

GitHub Flavored Markdown notes are just ignored by ProperDocs:

image

instead of

image

So the best solution seems to be keeping all notes in GFM format and converting them to admonitions format when generating documentation in ProperDocs.

## Type of change

## Checklist

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -11,9 +11,10 @@ True
 False
 ```
 
-!!! tip "Tip"
-    Don't do this! It is not efficient because it is initializing a new Holiday object and
-    generating a list of all the holidays in 2014 during each comparison.
+> [!tip]
+>
+> Don't do this! It is not efficient because it is initializing a new Holiday object and
+> generating a list of all the holidays in 2014 during each comparison.
 
 It is more efficient to create the object only once:
 
@@ -524,13 +525,14 @@ True
 1
 ```
 
-!!! tip "Tip"
-    Although the iCalendar specification supports a wide range of language formats as outlined
-    in [RFC 5646](https://datatracker.ietf.org/doc/html/rfc5646), the Holidays object requires
-    that the `language` attribute adhere to the [ISO 639-1 or
-    ISO 639-2](https://www.loc.gov/standards/iso639-2/php/code_list.php) language codes,
-    such as `en` or `pap-AW`. Additionally, if no `language` is specified for a holiday, but a
-    `default_language` is set for the Holiday object, the default language will be used instead.
+> [!tip]
+>
+> Although the iCalendar specification supports a wide range of language formats as outlined
+> in [RFC 5646](https://datatracker.ietf.org/doc/html/rfc5646), the Holidays object requires
+> that the `language` attribute adhere to the [ISO 639-1 or
+> ISO 639-2](https://www.loc.gov/standards/iso639-2/php/code_list.php) language codes,
+> such as `en` or `pap-AW`. Additionally, if no `language` is specified for a holiday, but a
+> `default_language` is set for the Holiday object, the default language will be used instead.
 
 ``` python
 >>> from holidays import country_holidays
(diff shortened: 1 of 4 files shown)


scripts/docs/gen_index.py
```diff
@@ -12,31 +12,84 @@
...
+
+        note_type = GFM_MKDOCS_MAPPING.get(match.group(1).lower(), "note")
+        i += 1
+
+        title = ""
+        if i < len(lines):
+            match = re.match(r"^>\s*\*\*(.+?)\*\*", lines[i])
+            if match:
+                title = f' "{match.group(1)}"'
+                i += 1
```
PPsyrius, line 64: Could've been a Walrus operator:
```suggestion
        if i < len(lines) and (match := re.match(r"^>\s*\*\*(.+?)\*\*", lines[i])):
            title = f' "{match.group(1)}"'
            i += 1
```

scripts/docs/gen_index.py
```diff
@@ -12,31 +12,84 @@
...
+}
+
+FILES_REPLACE_MAPPING = {
+    "index.md": (
+        (
+            "[CONTRIBUTING.md](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md)",
+            "[Contributing](contributing.md)",
+        ),
+    )
+}
```
PPsyrius, line 42: There's another `ICalExporter` reference that could've been accessible on Github version as well:
```suggestion
FILES_REPLACE_MAPPING = {
    "examples.md": (
        (
            "[ICalExporter](https://github.com/vacanza/holidays/blob/main/holidays/ical.py)",
            "[ICalExporter][holidays.ical.ICalExporter]",
        ),
    ),
    "index.md": (
        (
            "[CONTRIBUTING.md](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md)",
            "[Contributing](contributing.md)",
        ),
    ),
}
```

Then in `docs\examples.md`:
```python
<!-- markdownlint-disable-next-line MD052 -->
[ICalExporter][holidays.ical.ICalExporter] facilitates the creation and export of iCalendar files
in compliance with [RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545).
```
```python
[ICalExporter](https://github.com/vacanza/holidays/blob/main/holidays/ical.py) facilitates the creation
and export of iCalendar files in compliance with [RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545).
```

PPsyrius (commented):
Great execution IMO

I was originally thinking of https://github.com/PGijsbers/admonitions, but this PR avoids introducing new dependencies, so it's a win for me

docs/examples.md
```diff
@@ -492,7 +493,8 @@ True
 ## Generate iCalendar content and export to `.ics`
 
 <!-- markdownlint-disable-next-line MD052 -->
```
PPsyrius: This `markdownlint` rule is no longer needed with our recent change here
```suggestion
```

docs/examples.md
```diff
@@ -11,9 +11,10 @@ True
 False
 ```
 
-!!! tip "Tip"
-    Don't do this! It is not efficient because it is initializing a new Holiday object and
-    generating a list of all the holidays in 2014 during each comparison.
+> [!tip]
+>
+> Don't do this! It is not efficient because it is initializing a new Holiday object and
+> generating a list of all the holidays in 2014 during each comparison.
```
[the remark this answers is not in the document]
PPsyrius: @coderabbitai This is intended, doing so would be duplicating the label on GitHub side of things

PPsyrius (approved):
LGTM 🗒️

docs/examples.md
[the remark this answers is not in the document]
KJhellico: @coderabbitai, "Tip" is default title for "tip" note.

arkid15r (approved):

PR #3468, opened 2026-03-30
docs: fix minor English spelling errors in 12 country modules

sreeramabhavyasreechowdary:
## Proposed change

While performing a preliminary audit for my GSoC 2026 proposal using `codespell`, I identified and fixed minor English spelling errors in the documentation and comments across 12 modules.

**Modules Updated:**
- `france.py`: Corrected 'Chanel', 'Assomption', 'Readded'.
- `laos.py`: Corrected 'annualy', 'Indepedence'.
- `thailand.py`: Corrected 'hilight'.
- `brunei.py`: Corrected 'upcomming'.
- `united_states.py`: Corrected 'som'.
- `vietnam.py`: Corrected 'thay'.
- (And others in `bermuda.py`, `guyana.py`, `sweden.py`, `christian.py`, and `international.py`).

These are non-breaking, documentation-only changes. I intentionally excluded cultural/native holiday names (e.g., 'Whit', 'Te', 'Boun') as my proposal focuses on a multi-language `cspell` system to handle those specifically.

## Type of change

- [ ] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [x] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [ ] I've run `make check` locally; all checks and tests passed.

PR #3509, opened 2026-04-07
Standardize and fix docstring references

PPsyrius:
## Proposed change

- Clean up broken docstring references as seen when rendered on the ReadTheDocs version.
- Sort year-type references from old (top) to new (bottom) for China and Tanzania (other countries/markets already uses this pattern).
- Remove start year support notice for Cambodian and Laotian Lunar Calendar Holidays (this is irrelevant for their current implementation).
- Add a short notice for country entities that use `NoHolidayBase`.
image

Also includes a documentation typo fix for Lao holidays, carried over from the abandoned PR #3468 and migrates dict merging to [PEP 584](https://peps.python.org/pep-0584/).

## Type of change

## Checklist


KJhellico (dismissed):
LGTM! 👍Lots of small but essential things!

arkid15r:
@PPsyrius @KJhellico please check if those make sense to you.

PPsyrius:
Considering the l10n comment change in Thailand, IMO it might be the best to update their `en_US` translations as well

I did double-check just to be sure, and the non-Adjective forms, i.e. "New Year's Day (in lieu)", are already correct and can be left as-is

holidays/locale/en_US/LC_MESSAGES/TH.po
```diff
@@ -293,9 +293,9 @@ msgstr "Thai National Flag Day"
 msgid "วันลอยกระทง"
 msgstr "Loy Krathong"
 
-#. Special In Lieu Holiday.
+#. Special in-lieu holiday.
 msgid "วันหยุดชดเชย"
-msgstr "Special In Lieu Holiday"
+msgstr "Special in-lieu Holiday"
```
KJhellico: l10n comment and en_US translation need to be aligned here (I'm not sure which version to use).

PPsyrius: Let's go with capitalised but hyphenated ones, "Special In-Lieu Holiday" (that's what 3 out of 3 LLM says)

KJhellico (approved):
LGTM! 👍

arkid15r (commented):
👍

PR #3511, opened 2026-04-08
Update documentation: surface API for `ObservedHolidayBase` and `NoHolidayBase`

PPsyrius:
## Proposed change

- Improve documentation linkage for existing files
Screenshot 2026-04-08 160251

- Surface `ObservedHolidayBase` API file, add/update docstring references for both public and private methods (the latter isn't visible to ReadTheDocs, but a nice QoL change for first-time contributors for sure)
image
image
image

- Surface  `NoHolidayBase` to API file, remove `HolidayBase` base class boilerplate for readability.
image image

## Type of change

## Checklist


holidays/holiday_base.py
```diff
@@ -100,8 +100,8 @@ class HolidayBase(dict[date, str]):
 
         >>> assert date(2015, 1, 2) not in us_holidays
 
-    The `HolidayBase` class also recognizes strings of many formats
-    and numbers representing a POSIX timestamp:
+    The [`HolidayBase`][holidays.holiday_base.HolidayBase] class also recognizes strings
```
KJhellico, line 103: I'm not sure how useful it's to link to a class inside methods of that same class. 🙄

PPsyrius: Fair point, but IMO it's probably better for consistency's sake, even if it’s only a short scroll in this case.

KJhellico: > probably better for consistency's sake

Hmm, I agree.

KJhellico (dismissed):
LGTM! Documentation is something I've always been bad at. 🤷‍♂️

holidays/holiday_base.py
```diff
@@ -383,16 +380,24 @@ def __init__(
...
-        [HolidaySum][holidays.holiday_base.HolidaySum] object.
+        [`HolidaySum`][holidays.holiday_base.HolidaySum] object.
 
         Args:
             other:
                 The dictionary of public holiday to be added.
 
         Returns:
-            A `HolidayBase` object unless the other object cannot be added, then `self`.
+            A [`HolidaySum`][holidays.holiday_base.HolidaySum]
```
arkid15r, line 394: We should probably unify it to always return a HolidaySum

arkid15r (approved):
LGTM

KJhellico (approved):

PR #3549, opened 2026-04-21
Add static type checking support for `countries` and `financial` aliases

### LINKED ISSUE #1490: Pylance: "IT" is not exported from module "holidays.countries"
Hi,

i started using Pylance in VSCode and noted that it complains that:

`from holidays.countries import IT` 

"IT" is not exported from module "holidays.countries"
  Import from "holidays.countries.italy" instead

Is this a false positive? Because my code actually works.

Thanks!

### LINKED ISSUE #2682: Type check error
**Describe the bug**
When using vscode with pylance extension,the basic example like:
```python
import sys
sys.path.append(".")
print(sys.path)
from datetime import date
import holidays
print(holidays.__path__)
China_holidays = holidays.CN()
```
will invoke 红色波浪线 with CN (CN is unknow).
I think this is caused by CN is added to globals() by using udpdate dynamically.Is there a possible way to add a type definition file to solve this. I am a novice to this lib and python,so maybe i missed something,please forgive me.

**To Reproduce**

code is listed above

**Expected behavior**
I think there should be any type error.

**Screenshots**

Image

**Environment**

- OS: WSL 2 + Pylance extension
- Python version:3.12.3
- holidays version:0.76

**Additional context**

PPsyrius:
## Proposed change

This should solve the issue with `pylance` ever since we introduced lazy-loading in 2023 - users should now be able to import via aliases without triggering the `reportAttributeAccessIssue` false-positive error

Before:
Screenshot 2026-04-21 192419
After:
image
image
image

Resolves #1490
Resolves #2682

## Type of change

## Checklist


arkid15r (commented):
If import order doesn't matter here I'd prefer to keep them all on top.

LGTM

KJhellico (approved):
LGTM.

PR #2676, opened 2025-06-25
Update l10n for RTL languages

KJhellico:
## Proposed change

Update l10n for RTL languages (Arabic, Farsi, Hebrew).

Thanks to @Wasif-Shahzad for bringing this issue to my attention.

## Type of change

- [ ] New country/market holidays support (thank you!)
- [x] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've followed the [contributing guidelines][contributing-guidelines]
- [x] I've successfully run `make check`, all checks and tests are green

[contributing-guidelines]: https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md
[docs]: https://github.com/vacanza/holidays/tree/dev/docs/source

PR #3551, opened 2026-04-22
Update Arabic localization: fix "estimated" and "observed" labels translation

### LINKED ISSUE #3550: Arabic translation for “estimated” label is inaccurate
This was changed in #2676 
Current: `المقدرة`
Suggested: `تقدير`, (or `مقدرة`)
(المقدرة) → sounds like you’re naming a thing (“the estimated one”)

I would appreciate feedback from other Arabic speakers or contributors familiar with Arabic.

dimssu:
## Proposed change

Fixes the Arabic translation used for the `"%s (estimated)"` label. The current string `المقدرة` is the *definite* feminine passive participle ("the estimated one") and reads awkwardly as a parenthetical tag next to a date. This PR replaces it with `مقدرة` — the *indefinite* participle, which drops the leading definite article `ال` and matches English "estimated" as a parenthetical adjective/participle.

The companion string `المقدرة، ملاحظة` (used for `observed_estimated_label`) is updated the same way to `مقدرة، ملاحظة`.

The original Arabic string was introduced in #2676. Issue reporter ([@abh31000](https://github.com/abh31000)) suggested either `مقدرة` or `تقدير`; `مقدرة` is the smaller semantic change and keeps the label a participle rather than a noun.

**Scope:** 77 files — 19 country modules (`estimated_label` / `observed_estimated_label`), the `msgid` entries in every matching `.po` catalog across `ar`, `ar_EG`, `ar_QA`, `ar_SD`, `en_US`, `fr`, `es`, `bn`, `th`, and 10 test files that assert on the rendered Arabic string. `msgstr` values in non-Arabic catalogs are unchanged.

**⚠️ User-visible change:** downstream code that matches against the old Arabic string (`... (المقدرة)`) will break. Any fix to this issue necessarily has that property; flagging for release-notes consideration.

Closes #3550

## Type of change

## Checklist

- [x] I've run `make check` locally; all checks and tests passed (7058 passed, 100% coverage).


holidays/countries/egypt.py
```diff
@@ -45,11 +45,11 @@ class Egypt(
     country = "EG"
     default_language = "ar_EG"
     # %s (estimated).
-    estimated_label = tr("%s (المقدرة)")
+    estimated_label = tr("%s (مقدرة)")
     # %s (observed).
     observed_label = tr("%s (ملاحظة)")
     # %s (observed, estimated).
-    observed_estimated_label = tr("%s (المقدرة، ملاحظة)")
+    observed_estimated_label = tr("%s (مقدرة، ملاحظة)")
```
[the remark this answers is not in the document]
KJhellico: @coderabbitai, do you consider that this is RTL language?

holidays/countries/lebanon.py
```diff
@@ -38,11 +38,11 @@ class Lebanon(ObservedHolidayBase, ChristianHolidays, InternationalHolidays, Isl
     country = "LB"
     default_language = "ar"
     # %s (estimated).
-    estimated_label = tr("%s (المقدرة)")
+    estimated_label = tr("%s (مقدرة)")
     # %s (observed).
     observed_label = tr("%s (يُحتفل به)")
     # %s (observed, estimated).
-    observed_estimated_label = tr("%s (يُحتفل به، المقدرة)")
+    observed_estimated_label = tr("%s (يُحتفل به، مقدرة)")
```
KJhellico: These also need to be replaced with the version used in other countries, as I understand.

KJhellico (commented):
Looks solid. @abh31000, what do you think of this solution?

holidays/countries/lebanon.py
dimssu: @KJhellico thanks for the quick review! To confirm — do you want the Lebanon cleanup (replacing `يُحتفل به` with `ملاحظة` in `observed_label` / `observed_estimated_label` + matching .po catalogs + tests) folded into this PR, or handled as a separate follow-up? It's slightly outside the scope of #3550 (which is only about the `estimated` label), so happy to go either way — just wanted to check before expanding the diff.

dimssu:
Thanks for the quick look, @KJhellico! 🙏

@abh31000 — I went with `مقدرة` (the indefinite participle) rather than `تقدير` (noun) because dropping just the definite article `ال` is the smallest semantic change and keeps the word functioning as an adjective/participle, matching how English "(estimated)" reads as a parenthetical tag. Happy to switch to `تقدير` if you feel that reads more naturally — easy find-replace either way. Would appreciate your confirmation on the chosen word.

holidays/countries/lebanon.py
KJhellico: > do you want the Lebanon cleanup ... folded into this PR

Yes.

dimssu: Thanks for the quick answer! Folded it in as 4fab9b15 — `يُحتفل به` → `ملاحظة` across `lebanon.py`, the three LB .po catalogs, and `test_lebanon.py`. The combined form is now `مقدرة، ملاحظة` to match the shared msgid used in other countries.

`make check` clean locally (7058 tests pass).

abh31000:
@dimssu I thought about it and `تقديري` is more accurate, since "estimated" refers to the date being estimated, not the holiday itself (@KJhellico correct me if I'm wrong).
For "observed", `ملاحظ` doesn’t seem correct here. Something like `يوم بديل` (substitute day) or `يوم تعويضي` would better reflect the meaning, I suppose.

dimssu:
Thanks for thinking it through, @abh31000! That's a good point about agreement — if the implicit noun is `التاريخ` (the date, masc), then `تقديري` is more accurate than the feminine `مقدرة`. Happy to switch — it's a trivial find-replace across the same ~77 files. Waiting on @KJhellico to confirm before I push the change.

On the `observed` concern: you raise a real semantic issue — `ملاحظة` (a note/remark) doesn't quite capture "observed" in the holiday sense. `يوم بديل` / `يوم تعويضي` are much closer to the actual meaning (a replacement/compensatory weekday when the holiday falls on a weekend). That affects all ~18 MENA country modules though, so I'd lean toward splitting it off rather than expanding this PR's scope further — unless @KJhellico prefers to fold it in here to avoid a fragmented l10n history.

@KJhellico — two questions for you:
1. OK to swap `مقدرة` → `تقديري` in this PR per @abh31000's reasoning?
2. For the `observed` rewrite (`ملاحظة` → `يوم بديل` or `يوم تعويضي`): include here, or should I open a separate issue for it? I'm happy to open the follow-up issue and a subsequent PR myself either way.

KJhellico:
> since "estimated" refers to the date being estimated, not the holiday itself

Absolutely right!

> two questions for you:

1. Yes, تقديري seems more suitable.
2. Yeah, I think it's better to do all of it at once so we don't have to update same files again.

dimssu:
Thanks @abh31000 and @KJhellico! Pushed c9219776 with both changes:

- **Estimated**: `مقدرة` → `تقديري` (masc. adjective, agreeing with implicit `التاريخ`)
- **Observed**: `ملاحظة` → `يوم تعويضي` ("compensatory day" — matches the replacement-weekday semantics)
- **Combined** (`observed_estimated_label`): `مقدرة، ملاحظة` → `يوم تعويضي تقديري` as a single noun phrase ("estimated compensatory day"), which reads more naturally than two comma-separated tags with mismatched grammatical categories.

Scope: 84 files — all MENA country modules, every `.po` catalog (`ar`, `ar_EG`, `ar_QA`, `ar_SD`, `en_US`, `fr`, `es`, `bn`, `th`), matching tests, and the Arabic `msgstr` translations in `CA.po` / `UA.po`.

Local `make check` clean (7058 tests pass, 100% coverage). @abh31000 please let me know if the combined-form phrasing (`يوم تعويضي تقديري`) reads naturally or if you'd prefer a different ordering.

abh31000:
LGTM 👍

PPsyrius (approved):
LGTM 🌐

KJhellico (approved):
LGTM! 👍

arkid15r (commented):
Thanks for the contribution @dimssu 👍

PR #3640, opened 2026-06-20
Update iCalendar generation tool

### LINKED ISSUE #3638: `--years +N` parameter support for relative year ranges
## Is your feature request related to a problem? Please describe.

At present, `holidays-ics` requires explicit, hardcoded ranges for the --years flag (e.g., --years 2026-2036). Because these inputs are static, documentation examples inevitably become outdated, while automated scripts and cron jobs are forced to rely on external wrapper logic to maintain a rolling calendar window.

## Describe the solution you'd like

It would be highly convenient if the `--years` flag supported special dynamic string tokens or relative syntax to automatically calculate the current year or future windows.

Instead of uppercase or hardcoded years, the CLI could accept clean keywords or standard offset notation.

* `--years +10` (automatically infers starting from the current year and running for the next 10 years)

## Describe alternatives you've considered

Albeit bulky, if not dynamic tokens or relative offsets are supported natively in the CLI, I propose adding a dynamic one-liner example to the documentation at https://github.com/vacanza/holidays/blob/dev/docs/examples.md#examples ([https://github.com/vacanza/holidays/blob/dev/docs/examples.md#examples](https://github.com/vacanza/holidays/blob/dev/docs/examples.md#examples)). This ensures users always have a future-proof, copy-pasteable command that automatically handles a rolling 10-year window:

```
python3 -c "locale = 'US'; category = 'public'; import datetime, subprocess; years = f'{datetime.datetime.now().year}-{datetime.datetime.now().year + 10}'; subprocess.run(['python3', '-m', 'holidays.generate_ics', locale, '--categories', category, '--years', years])"
```

Design choices for this snippet:

* High Configurability: User-defined variables (`locale`, `category`) are placed at the absolute front (left-hand side) of the snippet so users can easily modify them without scanning or breaking the underlying execution logic.

## Additional context

While [https://holidays.readthedocs.io/en/latest/downloads/](Download Holiday Calendars) makes it convenient to select coming years by default, the CLI itself still requires explicit inputs.

### LINKED ISSUE #3646: Add `--list-categories` flag to view supported categories per country
#### Is your feature request related to a problem? Please describe.

Currently, there is no explicit or dedicated way to discover which holiday categories are available for a given country code.

While it is possible to find this information via a workaround—deliberately passing an empty/invalid category string to trigger a validation error message—it is unintuitive and difficult to remember.

```
$ holidays-ics US --categories ""
Unknown categories for US: . Supported categories: government, half_day, public, unofficial
```

#### Describe the solution you'd like

I would like an explicit `--list-categories` flag that outputs the supported categories for the specified country directly to the console, then exits cleanly.

**Example Usage:**

```
$ holidays-ics US --list-categories
Supported categories for US: government, half_day, public, unofficial
```

#### Describe alternatives you've considered

Using the workaround mentioned above (passing an empty string `""` or a dummy value to `--categories`). This safely forces the tool to print the supported options without generating an entire calendar or exiting with a disruptive crash, but a dedicated discovery flag would offer a much better user experience.

 ```
holidays-ics US --categories ""
Unknown categories for US: . Supported categories: government, half_day, public, unofficial
``` 

#### Additional context

This would improve CLI discoverability, especially when scripting or exploring available data for unfamiliar country codes.

### PR DESCRIPTION
**Pull request by KJhellico**
## Proposed change

Update iCalendar generation tool:
- add `--years +N / -N` parameter support, when +N means "the current year through the current year plus N", and -N means "the current year minus N through the current year"
- add `--list-subdivisions`, `--list-categories`, `--list-languages` options

Closes #3638, #3646.

## Type of change

## Checklist

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -649,6 +649,12 @@ Generate a holiday calendar for the current year and the next 5 years:
 holidays-ics US --years +5
 ```
 
+Generate a holiday calendar for the current year and the previous 5 years:
+
+```shell
+holidays-ics US --years -5
+```
+
 Generate a holiday calendar for multiple years and save it to a custom file:
 
 ```shell
@@ -661,10 +667,10 @@ Generate a calendar containing only bank holidays:
 holidays-ics AT --years 2025 --categories bank
 ```
 
-Generate a calendar containing public and optional holidays:
+Generate a calendar containing public and unofficial holidays:
 
 ```shell
-holidays-ics CA --years 2025 --categories public,optional
+holidays-ics CA --years 2025 --categories public,unofficial
 ```
 
 Generate a localized subdivision-specific calendar and save it to a custom file:
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -667,10 +667,10 @@ Generate a calendar containing only bank holidays:
 holidays-ics AT --years 2025 --categories bank
 ```
 
-Generate a calendar containing public and unofficial holidays:
+Generate a calendar containing public and optional holidays:
 
 ```shell
-holidays-ics CA --years 2025 --categories public,unofficial
+holidays-ics CA --years 2025 --categories public,optional
 ```
 
 Generate a localized subdivision-specific calendar and save it to a custom file:
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -626,6 +626,8 @@ For installation instructions, see `uv` [documentation](https://docs.astral.sh/u
 
 By default, the generated calendar contains holidays for the current year and is written
 to a file whose name is derived from the selected calendar and year range.
+The `--years` option accepts a single year (2025), a year range (2025-2027), or a
+relative range (+10), which selects the current year through the current year plus 10.
 
 ### Examples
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docs/examples.md [hunk 1]
```diff
@@ -641,6 +643,12 @@ Generate a holiday calendar for the Canton of Zurich for a specific year:
 holidays-ics CH --subdiv ZH --years 2025
 ```
 
+Generate a holiday calendar for the current year and the next 5 years:
+
+```shell
+holidays-ics US --years +5
```
arkid15r, line 649: Do we want to add negative delta support as well, e.g. `-5` ?

KJhellico: Actually, I don't see any point in that, that's why I added a check for positive numbers.

PPsyrius: Hmm, I think there's probably use cases for that still i.e. `--years -5` would output current year and last 5 years for auditing usage and those with data retention policies without the need to update that script yearly

KJhellico: Hmm, that seems logical!

arkid15r (approved):
LGTM

PPsyrius (approved):
LGTM 🗓️

PR #3647, opened 2026-06-23
Update iCalendar generation tool: add output suffix option

### LINKED ISSUE #3645: `--output-suffix` flag to modify default output filename ending
### **Is your feature request related to a problem? Please describe.**

When using `holidays-ics`, the tool automatically generates descriptive default filenames based on the arguments provided (e.g., `holidays-ics CH --subdiv ZH --years 2020` generates `CH_ZH_2020.ics`).

However, if a user wants to append a personal tag or context to that filename, they are forced to override the entire name using `-o/--output`. This breaks the convenience of the auto-generation, forcing the user to manually reconstruct the default string:
`holidays-ics CH --subdiv ZH --years 2020 --output CH_ZH_2020-MYNOTE.ics`

### **Describe the solution you'd like**

Introduce an `--output-suffix` flag that appends a custom text string directly to the end of the automatically generated base filename.

To keep behavior consistent with the existing `-o/--output` flag (which does not automatically append an extension), `--output-suffix` should **not** automatically add `.ics`. The user will explicitly include the extension in the suffix text.

**Example:**
`holidays-ics CH --subdiv ZH --years 2020 --output-suffix -MYNOTE.ics`

**How it should behave internally:**

1. The tool generates the base string: `CH_ZH_2020` (dropping its own default `.ics` extension).
2. It appends the exact string provided to `--output-suffix`.
3. **Resulting filename:** `CH_ZH_2020-MYNOTE.ics`

### **Describe alternatives you've considered**

* **Manual overriding:** Using `--output CH_ZH_2020-MYNOTE.ics`. This is tedious and error-prone, especially when scripting or running queries across multiple years/regions where the base name constantly changes.
* **Shell scripting wrappers:** Using `mv` or `rename` post-generation, which adds unnecessary friction for a high-level CLI tool.

### **Additional context**

Adding a suffix flag preserves the powerful "smart naming" logic already built into the tool while giving users the flexibility to tag their files (e.g., `-personal.ics`, `-work.ics`) without breaking the current CLI design philosophy regarding extensions.

sanmaxdev:
## Proposed change

Adds `--output-suffix` to `holidays-ics` so users can keep the generated smart filename base and append their own suffix without manually reconstructing the whole `--output` value.

Closes #3645.

Implementation details:

- Adds a mutually exclusive `--output-suffix` option alongside `-o/--output`.
- Keeps the existing default `.ics` behavior when no output option is passed.
- Appends the suffix exactly as provided, without automatically adding `.ics`.
- Supports suffix values that start with `-`, matching the issue example: `--output-suffix -MYNOTE.ics`.
- Documents the new option in the `holidays-ics` examples.
- Adds focused tests for generated suffix filenames, no implicit extension, mutual exclusion, and dash-prefixed suffix parsing.

## Type of change

## Checklist


sanmaxdev:
Follow-up pushed after review:

- Tightened `--output-suffix` normalization so it only rewrites single-dash suffix values (for example `-MYNOTE.ics`) and does not consume double-dash options such as `--list-categories`.
- Added regression coverage for normal suffix values, dash-prefixed suffix values, `--output-suffix=<value>`, missing suffix values, and double-dash options after `--output-suffix`.
- Applied the CodeRabbit nested-`with` cleanup.

Re-validated locally:

- `uv run ruff format holidays/generate_ics.py tests/test_generate_ics.py`
- `uv run ruff check holidays/generate_ics.py tests/test_generate_ics.py`
- `uv run pytest tests/test_generate_ics.py -q --cov=holidays.generate_ics --cov-report=term-missing --cov-fail-under=100` -> 40 passed, 100% coverage
- `uv run python -m holidays.generate_ics CH --subdiv ZH --years 2020 --output-suffix -MYNOTE.ics`
- `make test` before the review follow-up -> 7314 passed, 4 skipped, 100% coverage
- `make doc`
- `git diff --check`

sanmaxdev:
Pushed a small follow-up for the `--output-suffix` boundary cases.

Validation:
- `uv run --group tests python -m pytest tests/test_generate_ics.py -q -k 'output_suffix'`
- `uv run --group tests python -m compileall holidays/generate_ics.py tests/test_generate_ics.py`
- `uv run --group tests ruff check holidays/generate_ics.py tests/test_generate_ics.py`
- `git diff --check`

holidays/generate_ics.py
```diff
@@ -196,7 +218,10 @@ def run(self) -> None:
         start_year, end_year = self.args.years
         years_part = f"{start_year}_{end_year}" if start_year != end_year else f"{start_year}"
         subdiv_part = f"_{self.args.subdiv.upper().replace(' ', '_')}" if self.args.subdiv else ""
-        output_path = self.args.output or f"{self.args.code}{subdiv_part}_{years_part}.ics"
+        output_suffix = self.args.output_suffix if self.args.output_suffix is not None else ".ics"
```
KJhellico:
```suggestion
        output_suffix = self.args.output_suffix or ".ics"
```

tests/test_generate_ics.py
```diff
@@ -425,6 +449,46 @@ def test_filename_subdivision(self):
...
+            self.assertTrue((temp_dir / "US_2025-personal").exists())
+            self.assertFalse((temp_dir / "US_2025-personal.ics").exists())
+
+    def test_filename_output_and_suffix_are_mutually_exclusive(self):
+        with (
+            self.argv("US", "--output", "calendar.ics", "--output-suffix", "-personal.ics"),
+            self.assertRaises(SystemExit),
+        ):
+            IcsGenerator()
+
```
KJhellico:
```suggestion
```
Actually, it's a test of `argparse` functionality.

tests/test_generate_ics.py
```diff
@@ -425,6 +449,46 @@ def test_filename_subdivision(self):
...
+    def test_output_suffix_does_not_consume_long_options(self):
+        with self.argv("US", "--output-suffix", "--list-languages"):
+            with self.assertRaises(SystemExit):
+                IcsGenerator()
+
+    def test_output_suffix_missing_value_uses_argparse_error(self):
+        with self.argv("US", "--output-suffix"):
+            with self.assertRaises(SystemExit):
+                IcsGenerator()
+
```
KJhellico:
```suggestion
```
Same for this.

sanmaxdev:
Addressed the latest review notes:
- simplified the suffix fallback
- removed the two parser-behavior tests

Verification:
- `uv run --group tests python -m pytest tests/test_generate_ics.py -q -k 'output_suffix'`
- `uv run --group tests python -m compileall holidays/generate_ics.py tests/test_generate_ics.py`
- `uv run --group tests ruff check holidays/generate_ics.py tests/test_generate_ics.py`
- `git diff --check`

tests/test_generate_ics.py
```diff
@@ -425,6 +449,36 @@ def test_filename_subdivision(self):
...
+            self.argv("US", "--output", "calendar.ics", "--output-suffix", "-personal.ics"),
+            self.assertRaises(SystemExit),
+        ):
+            IcsGenerator()
+
+    def test_output_suffix_equals_form_is_unchanged(self):
+        args = IcsGenerator.normalize_output_suffix_args(["US", "--output-suffix=-personal.ics"])
+
+        self.assertEqual(args, ["US", "--output-suffix=-personal.ics"])
+
```
KJhellico:
```suggestion
```
Duplicate with `test_normalize_output_suffix_args` code.

sanmaxdev:
Removed the duplicate output suffix test from the latest review.

Verification:
- `uv run --group tests python -m pytest tests/test_generate_ics.py -q -k 'output_suffix'`
- `git diff --check`

sanmaxdev:
Pushed a small follow-up for the empty suffix case.

Verification:
- `uv run --group tests python -m pytest tests/test_generate_ics.py -q -k 'output_suffix'`
- `uv run --group tests python -m compileall holidays/generate_ics.py tests/test_generate_ics.py`
- `uv run --group tests ruff check holidays/generate_ics.py tests/test_generate_ics.py`
- `git diff --check`

KJhellico:
@sanmaxdev, thank you for your efforts, but this issue has been resolved by #3679.

PR #3651, opened 2026-06-25
Update `HolidayBase::get_closest_holiday` method

gaoflow:
## What

`get_closest_holiday` looks into the adjacent year when the search would cross a year boundary. It called `_populate(next_year)` / `_populate(previous_year)` but never added the year to `self.years`. This left the object in an inconsistent state:

- `self.years` reported only the originally loaded years even though the adjacent year's data was already in the dict.
- Every subsequent call to `get_closest_holiday` for the same date range saw the adjacent year still missing from `self.years`, so it called `_populate` again unnecessarily.

## Reproduction

```python
import holidays
from datetime import date

us = holidays.US(years=2024, expand=False)
print(us.years)                                         # {2024}

us.get_closest_holiday(date(2024, 12, 31), "forward")
print(us.years)                                         # {2024}  ← BUG: 2025 missing
print(len(us))                                          # 22 (2025 data IS in the dict)

# Second call redundantly calls _populate(2025) again
us.get_closest_holiday(date(2024, 12, 31), "forward")
print(us.years)                                         # {2024}  ← still wrong
```

## Fix

Add `self.years.add(next_year)` / `self.years.add(previous_year)` before the corresponding `_populate` calls, matching the pattern used in `__keytransform__` (the normal expansion path).

## Tests

Added `test_get_closest_holiday_updates_years` to `TestClosestHoliday` asserting that `self.years` is updated in both the forward and backward directions.

---

This pull request was prepared with the assistance of AI, under my direction and review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved closest-holiday searches when expanding into adjacent years.
  * Forward and backward searches now correctly include holidays from the neighboring year.
  * Searches with expansion disabled preserve existing calendar data and state.

* **Tests**
  * Added coverage for neighboring-year searches and state preservation.


holidays/holiday_base.py
```diff
@@ -1137,8 +1137,10 @@ def get_closest_holiday(
 
         dt = self.__keytransform__(target_date or datetime.now().date())
         if direction == "forward" and (next_year := dt.year + 1) not in self.years:
+            self.years.add(next_year)
             self._populate(next_year)
         elif direction == "backward" and (previous_year := dt.year - 1) not in self.years:
+            self.years.add(previous_year)
             self._populate(previous_year)
```
KJhellico: I think, in general, it would be more accurate here to consider the value of `self.expand` rather than simply adding an extra year.

eeshsaxena (commented):
Nice catch - this is a real inconsistency. I pulled the branch and confirmed the behavior on `dev` before the fix:

- `US(years=2024, expand=False)` then `get_closest_holiday(date(2024, 12, 31), "forward")` leaves `self.years == {2024}` even though `2025`'s holidays are already in the dict (`any(d.year == 2025 for d in us.keys())` is `True`). So the object really is in a state where the data and `self.years` disagree. Same thing in the `"backward"` direction with `2023`.

A couple of things I checked that might be useful for the review:

- **Completeness:** these two call sites look like the only `_populate()` calls that load a year which isn't already in `self.years` without recording it. The other callers (`__init__`, the re-populate loop after `clear()`) iterate over `self.years`, and `__keytransform__` already does `self.years.add(...)` right before `_populate(...)`. So the fix covers both affected branches.
- **Ordering:** adding the year before calling `_populate()` is safe here - `_populate()` doesn't consult `self.years`, and this matches the existing `__keytransform__` pattern (`self.years.add(dt.year); self._populate(dt.year)`).
- One thing I was initially worried about but which turned out fine: without the fix, a repeated `get_closest_holiday` call re-runs `_populate()` for the adjacent year (since it's still missing from `self.years`), and `_populate` documents that it "assumes that no holidays for the year have already been populated". In practice `__setitem__` dedupes identical names, so it doesn't corrupt the data - but the fix does avoid that redundant re-population, which is a nice side benefit worth keeping in the description.

One small, optional thought (very much defer to the maintainers): the `self.years.add(year); self._populate(year)` pair now appears in three places, and this bug came from a `_populate()` call that was missing its paired `add`. A tiny helper that does both could keep them in sync and make future call sites harder to get wrong - but that's a maintainability nicety beyond the scope of this fix.

Test looks good and exercises both directions. Thanks for tracking this down!

eeshsaxena (commented):
Nice catch - this looks correct to me, and it lines up with how the rest of the class manages `self.years`.

The convincing bit is that the fix mirrors the canonical expand path in `__keytransform__`:

```python
if self.expand and dt.year not in self.years:
    self.years.add(dt.year)
    self._populate(dt.year)
```

i.e. the established contract is that the *caller* adds the year to `self.years` before calling `_populate` (`_populate` itself only generates holidays and never touches `self.years`). `get_closest_holiday` was the one place calling `_populate` without first doing the `self.years.add(...)`, so the object was left inconsistent exactly as described. Adding the year first - in the same order as `__keytransform__` - is the right fix.

I applied the patch on top of `dev` and ran the `get_closest_holiday` tests: all pass, including the new `test_get_closest_holiday_updates_years` covering both the forward (`+1`) and backward (`-1`) branches. Good that both directions are exercised.

One tiny thing you might optionally note (not a blocker, and it matches existing behaviour): if the adjacent year falls outside `[start_year, end_year]`, `_populate` early-returns, so the year gets added to `self.years` with no data - but that's already exactly what the `__keytransform__` expand path does too, so this stays consistent with the current contract.

LGTM from me - deferring to the maintainers for the merge decision.

binggao1230:
Thanks for verifying both directions. On the optional point: the out-of-`[start_year, end_year]` case is intentional. Adding the year before `_populate` early-returns leaves `self.years` holding a data-less year, but that's exactly the existing `__keytransform__` expand contract — so I kept it consistent rather than special-casing `get_closest_holiday`. Happy to add a range guard if the maintainers would prefer, though that would diverge from `__keytransform__`.

sanmaxdev (commented):
Verified the fix locally. All 7 TestClosestHoliday tests pass, including the new regression test.

The inconsistency is real: before this PR, `_populate(next_year)` was called but the year was never added to `self.years`, so repeated calls to `get_closest_holiday` would re-populate the same year every time.

Regarding the `expand` point: the canonical pattern in `__keytransform__` (line 650) gates both `self.years.add()` and `self._populate()` on `self.expand`. However, `get_closest_holiday` already calls `_populate()` unconditionally (the pre-existing code did not check `self.expand`). So the method intentionally crosses year boundaries regardless of `expand`.

If the maintainer prefers keeping `self.years` fixed when `expand=False`, a middle ground would be to guard only the `self.years.add()` calls on `self.expand`, accepting redundant `_populate()` calls for the `expand=False` case. If `get_closest_holiday` is expected to work across year boundaries regardless of `expand` (as the pre-existing code suggests), the current approach of always tracking populated years is the right one.

KJhellico:
> If the maintainer prefers keeping `self.years` fixed when `expand=False`

Yes, I think we missed this point when adding this method, and we need to fix it. If someone sets `expand = False`, he obviously expects that `HolidayBase` object will not change its state while being accessed.

binggao1230:
Agreed — pushed 33d2910c. The adjacent-year population is now gated on `self.expand`, mirroring `__keytransform__`, so with `expand=False` the object state (years and keys) stays untouched and the search only covers the populated years. Added a test asserting both the unchanged state and the backward result within the populated range.

holidays/holiday_base.py
```diff
@@ -1136,10 +1136,14 @@ def get_closest_holiday(
             raise AttributeError(f"Unknown direction: {direction}")
 
         dt = self.__keytransform__(target_date or datetime.now().date())
-        if direction == "forward" and (next_year := dt.year + 1) not in self.years:
-            self._populate(next_year)
-        elif direction == "backward" and (previous_year := dt.year - 1) not in self.years:
-            self._populate(previous_year)
+        # Automatically expand for `expand=True` cases only, like `__keytransform__`.
```
KJhellico:
```suggestion
```

tests/test_holiday_base.py
```diff
@@ -1477,6 +1477,38 @@ def test_get_closest_holiday_unsorted_calendars(self):
             (date(2024, 1, 15), "Martin Luther King Jr. Day"),
         )
 
+    def test_get_closest_holiday_updates_years(self):
+        # Searching forward should add the next year to self.years so that
+        # repeated calls do not re-populate the adjacent year.
```
KJhellico:
```suggestion
```

tests/test_holiday_base.py
```diff
@@ -1477,6 +1477,38 @@ def test_get_closest_holiday_unsorted_calendars(self):
...
+
+        us.get_closest_holiday(date(2024, 12, 31), "forward")
+        self.assertIn(2025, us.years)
+
+        # Searching backward should add the previous year to self.years.
+        us2 = US(years=2024)
+        self.assertEqual(us2.years, {2024})
+
+        us2.get_closest_holiday(date(2024, 1, 1), "backward")
+        self.assertIn(2023, us2.years)
```
KJhellico:
```suggestion
        us_2 = US(years=2024)
        self.assertEqual(us_2.years, {2024})

        us_2.get_closest_holiday(date(2024, 1, 1), "backward")
        self.assertIn(2023, us_2.years)
```

tests/test_holiday_base.py
```diff
@@ -1477,6 +1477,38 @@ def test_get_closest_holiday_unsorted_calendars(self):
...
+
+    def test_get_closest_holiday_expand_false_keeps_state(self):
+        # With expand=False the object must not change state while accessed.
```
KJhellico:
```suggestion
```

binggao1230:
Applied all four suggestions in d5966a5.

KJhellico (approved):
LGTM.

PPsyrius (approved):
LGTM 🛠️

arkid15r (approved):

arkid15r (commented):
Thanks @gaoflow 👍

PR #3639, opened 2026-06-20
Add Bolsa Mexicana de Valores (BMV) holidays

pareshjoshij:
## Proposed change

This PR adds the holidays for  Mexicana de Valores (XMEX) financial calendar and introduces localization (l10n) support.

Part of GSoC (https://github.com/vacanza/holidays/issues/3424)

- [X] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] I've run `make check` locally; all checks and tests passed.

PR #3617, opened 2026-06-11
Add  Chicago Mercantile Exchange  (XCME) holidays

pareshjoshij:
This PR adds the holidays for the XCME financial calendar and introduces localization (l10n) support.

Part of GSoC (#3424)

- [x] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] i've run `make check` locally; all checks and tests passed.

PR #3654, opened 2026-06-29
Add London Stock Exchange (LSE) holidays

eeshsaxena:
### Proposed change

Adds the **London Stock Exchange** (`XLON`) to the supported financial markets. The LSE is one of the world's largest exchanges and was not yet covered, while several smaller exchanges already are.

The LSE is closed on weekends and on the bank holidays observed in England and Wales:

- New Year's Day (observed)
- Good Friday and Easter Monday
- Early May Bank Holiday ("May Day")
- Spring Bank Holiday (with the Jubilee-year moves in 2002, 2012 and 2022)
- Late Summer Bank Holiday
- Christmas Day and Boxing Day (observed)

plus the one-off royal bank holidays: the 2002 Golden Jubilee, the 2011 royal wedding, the 2012 Diamond Jubilee, the 2022 Platinum Jubilee, the 2022 State Funeral of Queen Elizabeth II, and the 2023 Coronation of Charles III.

`start_year` is 2000 (the modern electronic-trading era).

### Verification

- New `tests/financial/test_london_stock_exchange.py` (per-year assertions, aliases, and a no-holidays check) - all pass, 100% coverage of the new module.
- The market's output was cross-checked against the library's existing `UnitedKingdom(subdiv="ENG")` bank-holiday data for every year 2000-2030 (exact match on both dates and names).
- `tests/test_docs.py` and `tests/test_registry.py` pass; README financial-markets table and registry updated; `ruff check`/`format` clean.

### Aliases

`LondonStockExchange`, `XLON`, `LSE`.


KJhellico:
If holidays are the same as in UK, why not just inherit from `UnitedKingdom` class?

eeshsaxena:
Good question, and I did try that first. The blocker is that the LSE follows specifically the **England and Wales** bank holidays, which in `UnitedKingdom` are a subdivision (`subdiv="ENG"`), not the entity's default output.

I prototyped `class LondonStockExchange(UnitedKingdom)` (mirroring `JapanExchange(Japan)` / `HongKongStockExchange(HongKong)`), but those work cleanly because Japan, Hong Kong and China have no subdivisions. To get England here I have to force `subdiv="ENG"`, and that makes the market entity retain `country`/`subdiv` and pull in UK's subdivisions and localization - which fails the financial `test_code` (`assertFalse(hasattr(self.holidays, "country"))`) and changes the supported-languages table.

To make sure the standalone version is faithful, I cross-checked its output against `UnitedKingdom(subdiv="ENG")` for every year 2000-2030 and they match exactly (dates and names), which is why the special days (jubilees, the 2022 state funeral, the 2023 coronation) are all covered.

Happy to go whichever way you prefer:
1. keep the standalone implementation as-is, or
2. inherit from `UnitedKingdom` if there's an accepted pattern for a subdivision-specific market that I'm missing - if so, point me at it and I'll switch.

arkid15r:
> 2\. inherit from `UnitedKingdom` if there's an accepted pattern for a subdivision-specific market that I'm missing - if so, point me at it and I'll switch.

Is [this](https://github.com/search?q=repo%3Avacanza%2Fholidays%20parent_entity_subdivision_code&type=code) what you are looking for?

eeshsaxena:
Thanks @arkid15r, that's exactly the pointer I needed. I tried it:

```python
class LondonStockExchange(ChildEntity, UnitedKingdom):
    country = None
    market = "XLON"
    parent_entity = UnitedKingdom
    parent_entity_subdivision_code = "ENG"
    supported_categories = (PUBLIC,)
    start_year = 2000
```

The output matches `UnitedKingdom(subdiv="ENG")` exactly for 2000-2030 (dates and names), so functionally this is the right approach and removes all the duplicated logic.

The one thing I'm stuck on is `CommonFinancialTests.test_code`, which asserts `assertFalse(hasattr(self.holidays, "country"))`. With `ChildEntity` the instance ends up with `subdiv="ENG"` and a (None) `country` attribute, so `hasattr(..., "country")` is `True` and the test fails - whereas the existing no-subdivision markets like `JapanExchange(Japan)` pass it. (The README supported-languages column also then needs `en_GB, en_US, th` from the parent, which is easy.)

Is there an additional step to make a subdivision-backed market not expose `country` for `test_code`, or should `test_code` be relaxed for `ChildEntity`-based markets? Happy to push the `ChildEntity` version once I know how you'd like `test_code` handled. I've kept the standalone version here in the meantime so CI stays green.

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,112 @@
...
+
+    References:
+        * <https://en.wikipedia.org/wiki/London_Stock_Exchange>
+        * <https://en.wikipedia.org/wiki/Bank_holiday>
+        * [LSE business days](https://www.londonstockexchange.com/equities-trading/business-days)
+    """
+
+    market = "XLON"
+    observed_label = "%s (observed)"
+    start_year = 2000
```
pareshjoshij, line 41: @eeshsaxena the LSE also observes early closures/half days (like Christmas Eve and New Year's Eve). Here is the [reference link](https://www.londonstockexchange.com/equities-trading/business-days)

 ​We will need to add half-day support to make this fully accurate. I've already done the research on this, so I am happy to help you implement this section if you'd like! Just let me know

eeshsaxena:
Thanks @pareshjoshij! Added half-day support - Christmas Eve and New Year's Eve early closes are now exposed under the `HALF_DAY` category, following the existing exchange half-day pattern (e.g. Hong Kong). Added a `test_2024_half_day` case and the full LSE suite passes (17 tests). Let me know if you'd also like weekend dates filtered out of the half-day set.

pareshjoshij:
@eeshsaxena  how will the end user know that these dates are half-days?
We need to make sure a label is added so it's clearly visible to them.


> Let me know if you'd also like weekend dates filtered out of the half-day set.

take a look at the recent bot comment on this PR. For the best approach, I suggest reviewing #3617 to see exactly how we implemented half-day holidays previously, and let's align with that pattern. :)

pareshjoshij (commented):
In my opinion, it would be highly beneficial to cross-reference your output with historical LSE calendars. For example, you can use the Wayback Machine to look at past versions of their business days page: [LSE Business Days (Archive)](https://web.archive.org/web/20260000000000*/http://www.londonstockexchange.com/about-the-exchange/company-overview/business-days/business-days.htm).

Here is why I suggest doing this:
* **Accuracy Check:** Matching the official historical holidays against your generated output will give you a clear idea of whether your current implementation is accurate, or if inheriting from `UnitedKingdom` might actually be a better approach.
* **Easier Reviews:** If you add these historical links to the `References` section in the docstring, it significantly reduces the workload for maintainers when verifying the dates. 

I highly recommend digging up a few of these historical links yourself—it is a great way to learn the intricacies of the exchange's calendar! If you need a good research strategy for finding them, check out this trick shared by @PPsyrius  in a previous review: [PR #3639 Research Tip](https://github.com/vacanza/holidays/pull/3639#pullrequestreview-4544491111). 

Just my two cents—hope this helps!

KJhellico:
> The one thing I'm stuck on is `CommonFinancialTests.test_code`

Look at similar markets that derive from countries - JPX, SSE etc.

eeshsaxena:
Thanks both!

@pareshjoshij - done. Following the #3617 (XCME) pattern, the half-day sessions are now labelled via `_format_holiday_name`, so under the `HALF_DAY` category they render as:

```
2024-12-24  Christmas Eve (early close)
2024-12-31  New Year's Eve (early close)
```

so end users can clearly tell these are shortened trading days (12:30 close) rather than full market holidays.

@KJhellico - on inheriting from `UnitedKingdom`: I re-checked empirically and the clean JPX/SSE pattern (`country = None`, `parent_entity = ...`, no `subdiv`) doesn't reproduce the LSE calendar. JPX/SSE work because Japan/China have no subdivisions, so the entity default *is* the market calendar. For the UK the entity default is **not** England - e.g. for 2024 it omits Easter Monday (Apr 1) and the Late Summer Bank Holiday (Aug 26), both of which the LSE observes:

```
UnitedKingdom()            -> {Jan1, GoodFri, MayDay, SpringBH, Christmas, Boxing}
UnitedKingdom(subdiv=ENG)  -> + Easter Monday, + Late Summer BH   # what LSE follows
```

So the LSE specifically needs `subdiv="ENG"`. The `ChildEntity` + `parent_entity_subdivision_code="ENG"` route does give the right dates, but it keeps `subdiv`/`country` on the instance and fails `CommonFinancialTests.test_code` (`assertFalse(hasattr(self.holidays, "country"))`). The standalone implementation here reproduces `UnitedKingdom(subdiv="ENG")` exactly for 2000-2030 and passes all common financial tests (17 tests green, no `country` attr). Happy to switch to inheritance instead if there's a market precedent that pins a specific subdivision while still satisfying `test_code` - I couldn't find one, but point me at it and I'll adopt it.

KJhellico:
> point me

I was talking exactly about this test, not the implementation.

Of course, for the implementation itself, this
```python
    parent_entity_subdivision_code = "ENG"
```
is needed.

Also, for the HALF_DAY category to work properly, you need to add
```diff
diff --git a/holidays/countries/united_kingdom.py b/holidays/countries/united_kingdom.py
--- a/holidays/countries/united_kingdom.py	(revision b0d20a82c4b8fd2001cb3b3fb9fabfd8f491b263)
+++ b/holidays/countries/united_kingdom.py	(date 1783183559104)
@@ -13,6 +13,7 @@
 from gettext import gettext as tr
 
 from holidays.calendars.gregorian import APR, MAY, JUN, JUL, SEP, DEC
+from holidays.constants import PUBLIC
 from holidays.groups import ChristianHolidays, InternationalHolidays, StaticHolidays
 from holidays.observed_holiday_base import (
     ObservedHolidayBase,
@@ -104,7 +105,7 @@
             self._add_whit_monday(tr("Whit Monday"))
 
     def _populate_subdiv_holidays(self):
-        if self.subdiv != "SCT":
+        if PUBLIC in self.categories and self.subdiv != "SCT":
             if self._year >= 1975:
                 # New Year's Day.
                 self._add_observed(self._add_new_years_day(tr("New Year's Day")))

```

BTW, since the United Kingdom supports l10n, it would be advisable to implement it here as well.

eeshsaxena:
Thanks @KJhellico - refactored to the inheritance approach as you suggested. Summary of the changes:

**1. Inherit from `UnitedKingdom`.** `LondonStockExchange` is now `class LondonStockExchange(ChildEntity, UnitedKingdom)` with `parent_entity_subdivision_code = "ENG"`, mirroring the JPX/SSE pattern. All the duplicated England bank-holiday logic is gone - the public output is now an exact alias of `UnitedKingdom(subdiv="ENG")` (verified equal for 2000-2030), including the one-off royal bank holidays.

**2. `test_code`.** Followed JPX/SSE - the test now asserts `market` is present and `country` is `None`.

**3. HALF_DAY guard.** Applied your `united_kingdom.py` diff exactly:
```diff
-        if self.subdiv != "SCT":
+        if PUBLIC in self.categories and self.subdiv != "SCT":
```
so requesting only `HALF_DAY` no longer pulls in New Year's Day / Christmas / Boxing Day. All 30 UK tests and Isle of Man still pass.

**4. l10n.** Added localization for the half-day strings (source `en_GB`, plus `en_US` and `th`), reusing the repo's existing `%s (half-day closing)` label so the strings are consistent with other markets. Rendered output:
```
en_GB/en_US:  Christmas Eve (half-day closing) | New Year's Eve (half-day closing)
th:           วันคริสต์มาสอีฟ (ปิดครึ่งวัน)   | วันสิ้นปี (ปิดครึ่งวัน)
```
README updated with XLON's languages + `HALF_DAY` category.

`test_l10n`, `test_docs`, the LSE/UK/Isle of Man suites and ruff all pass. Let me know if you'd like any adjustments.

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,66 @@
...
+class LondonStockExchange(ChildEntity, UnitedKingdom):
+    """London Stock Exchange holidays.
+
+    The London Stock Exchange (LSE) is closed on the bank holidays observed in
+    England and Wales, so its public holidays are an alias of the England
+    (``ENG``) subdivision of the United Kingdom, including the one-off royal
+    bank holidays.
+
+    On Christmas Eve and New Year's Eve the exchange runs a shortened trading
+    session (a half-day closing), available under the ``HALF_DAY`` category.
```
KJhellico:
```suggestion
    """London Stock Exchange (LSE) holidays.

    The London Stock Exchange is closed on the bank holidays observed in
    England and Wales.

    On Christmas Eve and New Year's Eve the exchange runs a shortened trading
    session.
```

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,66 @@
...
+        # %s (half-day closing).
+        half_day_closing_label = tr("%s (half-day closing)")
+
+        # Christmas Eve.
+        christmas_eve = self._format_holiday_name(half_day_closing_label, tr("Christmas Eve"))
+        self._add_christmas_eve(christmas_eve)
+
+        # New Year's Eve.
+        new_years_eve = self._format_holiday_name(half_day_closing_label, tr("New Year's Eve"))
+        self._add_new_years_eve(new_years_eve)
```
KJhellico:
```suggestion
        # %s (markets close at 12:30pm).
        close_12_30pm_label = tr("%s (markets close at 12:30pm)")

        self._add_christmas_eve(
            # Christmas Eve.
            self._format_holiday_name(close_12_30pm_label, tr("Christmas Eve"))
        )

        self._add_new_years_eve(
            # New Year's Eve.
            self._format_holiday_name(close_12_30pm_label, tr("New Year's Eve"))
        )
```
And update .po files accordingly.

holidays/locale/en_GB/LC_MESSAGES/XLON.po
```diff
@@ -0,0 +1,41 @@
...
+#  specific sets of holidays on the fly. It aims to make determining whether a
+#  specific date is a holiday as fast and flexible as possible.
+#
+#  Authors: Vacanza Team and individual contributors (see CONTRIBUTORS file)
+#           dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
+#           ryanss <ryanssdev@icloud.com> (c) 2014-2017
+#  Website: https://github.com/vacanza/holidays
+#  License: MIT (see LICENSE file)
+#
+# London Stock Exchange holidays.
```
KJhellico:
```suggestion
# London Stock Exchange (LSE) holidays.
```
As in other markets .po files (same for rest of files).

tests/financial/test_london_stock_exchange.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+class TestLondonStockExchange(CommonFinancialTests, TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass(LondonStockExchange)
+
+    def test_code(self):
+        # LondonStockExchange derives from UnitedKingdom, so (like JPX/SSE) it
+        # keeps a ``country`` attribute set to None rather than removing it.
```
KJhellico:
```suggestion
```

tests/financial/test_london_stock_exchange.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertTrue(hasattr(self.holidays, "market"))
+        self.assertIsNone(getattr(self.holidays, "country", None))
+
+    def test_market_aliases(self):
+        self.assertAliases(LondonStockExchange, XLON, LSE)
+
+    def test_no_holidays(self):
+        self.assertNoHolidays(LondonStockExchange(years=1999))
+
```
KJhellico:
```suggestion
```

tests/financial/test_london_stock_exchange.py
```diff
@@ -0,0 +1,158 @@
...
+    def test_2002(self):
```
KJhellico: Need to keep only one test from "one-year" and add:
- Christmas Eve and New Year's Eve standard tests
- l10n tests

eeshsaxena:
Thanks @KJhellico - applied all of it:

1. **Docstring** - simplified to your suggested version.
2. **Half-day label** - renamed to \`%s (markets close at 12:30pm)\`, restructured the two \`_add_christmas_eve\`/\`_add_new_years_eve\` calls with inline comments as suggested, and updated the msgid across all three .po files (en_GB, en_US, th) accordingly.
3. **.po header comments** - updated to \`# London Stock Exchange (LSE) holidays[...]\` to match the docstring/other markets' convention.
4. **Tests** - trimmed to one representative year (2022, which still exercises the Platinum Jubilee + State Funeral one-offs), added dedicated \`test_christmas_eve\`/\`test_new_years_eve\` half-day tests, and \`test_l10n_default\`/\`test_l10n_en_us\`/\`test_l10n_th\`. Year-by-year calendar correctness stays covered by the United Kingdom test suite, per the alias relationship.

Also added \`-> None\` to \`_populate_half_day_holidays\` per CodeRabbit's nitpick. Left the CodeRabbit "missing 2012 Jubilee test" suggestion out since it conflicts with consolidating to one yearly test - happy to add it back if you'd rather keep both.

Ruff, \`test_l10n\`, \`test_docs\`, the LSE/UK suites (14 + 43 tests) all green locally.

pareshjoshij:
@eeshsaxena 

```diff

- "Project-Id-Version: Holidays 0.100\n"

+ "Project-Id-Version: Holidays 0.101\n"

```

tests/financial/test_london_stock_exchange.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def test_no_holidays(self):
+        self.assertNoHolidays(LondonStockExchange(years=1999))
+
+    def test_2022(self):
+        # Exercises the royal one-off bank holidays alongside the regular
+        # England & Wales calendar; year-by-year correctness of the underlying
+        # calendar itself is covered by the United Kingdom test suite.
+        self.assertHolidays(
+            LondonStockExchange(years=2022),
```
KJhellico:
```suggestion
        self.assertHolidaysInYear(
            2022,
```
An place this test before l10n tests.

eeshsaxena:
Thanks both - done:

- @pareshjoshij: bumped `Project-Id-Version` to `Holidays 0.101` in all three XLON `.po` files (and regenerated the `.mo` files).
- @KJhellico: switched the 2022 test to `assertHolidaysInYear(2022, ...)`; it already sits above the l10n tests.

14 LSE tests plus `test_docs`/`test_registry` pass locally and ruff is clean.

tests/financial/test_london_stock_exchange.py
KJhellico: @eeshsaxena, the suggestion includes removing the redundant comment.
And "before" means "right before". ;)

tests/financial/test_london_stock_exchange.py
KJhellico: @eeshsaxena, this remains relevant.

pareshjoshij (commented):
- **Unshifted Half-Days:** When Christmas Eve (Dec 24) and New Year's Eve (Dec 31) fall on a weekend, the LSE shifts the half-day early close to the preceding Friday. The current script hardcodes these to the 24th and 31st, resulting in market opens being reported on Saturdays or Sundays.
- Because the class inherits from the  `UnitedKingdom` calendar, it emits standard weekend holidays (e.g., actual Christmas Day on a Sunday) alongside the observed weekday substitutes. IMO it should only output affected business days .

**Comparison (Using [2011 as an example](https://web.archive.org/web/20110201180705/http://www.londonstockexchange.com/about-the-exchange/company-overview/business-days/business-days.htm))**
In 2011, Dec 24 and Dec 31 fell on Saturdays. The official LSE schedule shifted the early closes to Friday, Dec 23 and Friday, Dec 30.

| Current Output  | Official LSE Schedule  |
|---|---|
| `2011-01-01` -> New Year's Day | *(Removed - Weekend)* |
| `2011-01-03` -> New Year's Day (observed) | `2011-01-03` -> New Year's Day (observed) |
| `2011-04-22` -> Good Friday | `2011-04-22` -> Good Friday |
| `2011-04-25` -> Easter Monday | `2011-04-25` -> Easter Monday |
| `2011-04-29` -> Wedding of William and Catherine | `2011-04-29` -> Wedding of William and Catherine |
| `2011-05-02` -> May Day | `2011-05-02` -> May Day |
| `2011-05-30` -> Spring Bank Holiday | `2011-05-30` -> Spring Bank Holiday |
| `2011-08-29` -> Late Summer Bank Holiday | `2011-08-29` -> Late Summer Bank Holiday |
| `2011-12-24` -> Christmas Eve (markets close at 12:30pm) | **`2011-12-23` -> Christmas Eve (markets close at 12:30pm)** |
| `2011-12-25` -> Christmas Day | *(Removed - Weekend)* |
| `2011-12-26` -> Boxing Day | `2011-12-26` -> Boxing Day |
| `2011-12-27` -> Christmas Day (observed) | `2011-12-27` -> Christmas Day (observed) |
| `2011-12-31` -> New Year's Eve (markets close at 12:30pm) | **`2011-12-30` -> New Year's Eve (markets close at 12:30pm)** |

eeshsaxena:
Thanks @pareshjoshij, this is a great catch and your 2011 reference is spot on. I reproduced both issues on the current branch:

```
HALF_DAY 2011:
  2011-12-24 Sat  Christmas Eve (markets close at 12:30pm)   <- Saturday, market shut
  2011-12-31 Sat  New Year's Eve (markets close at 12:30pm)  <- Saturday, market shut

PUBLIC 2011:
  2011-01-01 Sat  New Year's Day        <- weekend
  2011-12-25 Sun  Christmas Day         <- weekend
  ... (observed substitutes on 01-03 and 12-27 are present and correct)
```

There are really two separate things here, and I want to get the second one right rather than guess:

**1. Half-day early closes falling on a weekend.** This one is clearly a bug - an early close on a Saturday is meaningless. I can fix it by shifting the half-day to the preceding Friday, which matches your 2011 evidence (Dec 24 Sat -> Dec 23 Fri, Dec 31 Sat -> Dec 30 Fri). The library already has `SAT_SUN_TO_PREV_FRI` (`{SAT: -1, SUN: -2}`) in `observed_holiday_base.py`, so I would move the two `_add_christmas_eve` / `_add_new_years_eve` calls through that rule. Happy to push that.

**2. Weekend-falling public holidays.** I agree a market calendar arguably shouldn't list Christmas Day when it lands on a Sunday (the market is already closed, and the observed substitute is what actually matters). The wrinkle is that these come from the inherited `UnitedKingdom(subdiv="ENG")` population via `ChildEntity` - which is the design @arkid15r and @KJhellico pointed me toward specifically so LSE reuses the UK bank-holiday logic rather than reimplementing it. Suppressing the weekend-falling ones means overriding/post-filtering that inherited output, which pulls a little against that clean-inheritance approach (e.g. XNYS gets this "for free" only because it builds its own calendar from scratch).

So before I implement #2: @KJhellico @pareshjoshij, what is the direction you'd prefer - filter out inherited holidays that fall on a weekend (drop the raw Sat/Sun dates, keep the observed substitutes), or keep the full inherited set as-is? I'll implement whichever you both agree on. I can push the half-day fix (#1) right away since that one seems uncontroversial.

pareshjoshij:
> So before I implement #2: @KJhellico @pareshjoshij, what is the direction you'd prefer - filter out inherited holidays that fall on a weekend (drop the raw Sat/Sun dates, keep the observed substitutes), or keep the full inherited set as-is? I'll implement whichever you both agree on. I can push the half-day fix (#1) right away since that one seems uncontroversial.


Don't worry buddy, we're in the same boat! Take a look at my open PR #3692  below and just read through the [mentor's suggestions](https://github.com/vacanza/holidays/pull/3692#discussion_r3571311781)

eeshsaxena:
Thanks @pareshjoshij - that pointer was exactly what I needed, and thanks @KJhellico for the guidance on #3692. Both of your points are now implemented.

**1. Weekend bank holidays are dropped.** Following the approach @KJhellico prescribed on #3692, the inherited England and Wales dates that land on a Saturday or Sunday are suppressed, so only the observed weekday substitutes remain:

```python
def _populate(self, year: int) -> None:
    super()._populate(year)

    if self.observed:
        for dt in tuple(self.keys()):
            if dt.year == year:
                self._move_holiday(dt, rule=SAT_TO_NONE + SUN_TO_NONE, show_observed_label=False)
```

I used the composed `SAT_TO_NONE + SUN_TO_NONE` rather than adding `SAT_SUN_TO_NONE` here, so this does not collide with #3692 - once that lands and the constant exists, this is a one-line swap.

**2. Half-days move to the preceding Friday.** The shortened Christmas Eve and New Year's Eve sessions now shift back via `SAT_SUN_TO_PREV_FRI` when the eve falls on a weekend.

2011 now matches the official schedule you linked, exactly:

| Date | Category | Name |
|---|---|---|
| `2011-01-03` | PUBLIC | New Year's Day (observed) |
| `2011-04-22` | PUBLIC | Good Friday |
| `2011-04-25` | PUBLIC | Easter Monday |
| `2011-04-29` | PUBLIC | Wedding of William and Catherine |
| `2011-05-02` | PUBLIC | May Day |
| `2011-05-30` | PUBLIC | Spring Bank Holiday |
| `2011-08-29` | PUBLIC | Late Summer Bank Holiday |
| **`2011-12-23`** | HALF_DAY | Christmas Eve (markets close at 12:30pm) |
| `2011-12-26` | PUBLIC | Boxing Day |
| `2011-12-27` | PUBLIC | Christmas Day (observed) |
| **`2011-12-30`** | HALF_DAY | New Year's Eve (markets close at 12:30pm) |

`2011-01-01` (Sat) and `2011-12-25` (Sun) are gone, and the two early closes sit on the preceding Fridays.

**Tests:** updated `test_2022` (the Sat/Sun raw dates are no longer emitted), extended the Christmas Eve / New Year's Eve tests to pin the shifted cases (2011 Sat, 2017 Sun, 2024 weekday), and added `test_weekend_holidays_are_not_trading_days`. 15 LSE tests pass at 100% coverage; `tests/financial/` + the United Kingdom suite are green (293 passed); `test_docs` and `test_registry` pass; ruff clean.

Agreed with @KJhellico that a cleaner mechanism for country-backed markets would be nice down the line - happy to help with that once the pattern settles.

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,80 @@
...
+                self._format_holiday_name(close_12_30pm_label, tr("Christmas Eve"))
+            ),
+            self._add_new_years_eve(
+                # New Year's Eve.
+                self._format_holiday_name(close_12_30pm_label, tr("New Year's Eve"))
+            ),
+        ):
+            # The shortened session moves to the preceding Friday when the eve falls
+            # on a weekend, since the exchange does not trade on Saturday or Sunday.
+            self._move_holiday(dt, rule=SAT_SUN_TO_PREV_FRI, show_observed_label=False)
```
KJhellico:
```suggestion
        self._move_holiday(
            self._add_christmas_eve(
                # Christmas Eve.
                self._format_holiday_name(close_12_30pm_label, tr("Christmas Eve"))
            ),
            rule=SAT_SUN_TO_PREV_FRI,
            show_observed_label=False,
        )

        self._move_holiday(
            self._add_new_years_eve(
                # New Year's Eve.
                self._format_holiday_name(close_12_30pm_label, tr("New Year's Eve"))
            ),
            rule=SAT_SUN_TO_PREV_FRI,
            show_observed_label=False,
        )
```
It's advisable to simply expand on the existing comment about Christmas Eve and New Year's Eve in docstring.

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _populate(self, year: int) -> None:
+        super()._populate(year)
+
+        # The exchange is closed at weekends, so a bank holiday that falls on a
+        # Saturday or Sunday is not a trading day in its own right; only its
+        # observed weekday substitute is.
```
KJhellico: I'm not sure whether such a detailed comment is necessary, but I hope that in the end we'll get completely rid of this code fragment.

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,76 @@
...
+            rule=SAT_SUN_TO_PREV_FRI,
+        )
+
+    def _populate(self, year: int) -> None:
+        super()._populate(year)
+
+        if self.observed:
+            for dt in tuple(self.keys()):
+                if dt.year == year:
+                    self._move_holiday(dt, rule=SAT_SUN_TO_NONE, show_observed_label=False)
```
KJhellico:
```suggestion
    def _populate_public_holidays(self):
        super()._populate_public_holidays()

        for dt in tuple(self.keys()):
            if self._is_weekend(dt):
                self.pop(dt)
```
The approach used in other similar markets.
(And move this method up)

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,76 @@
...
+    The London Stock Exchange is closed on the bank holidays observed in
+    England and Wales.
+
+    On Christmas Eve and New Year's Eve the exchange runs a shortened trading
+    session.
+
+    References:
+        * <https://en.wikipedia.org/wiki/London_Stock_Exchange>
+        * <https://en.wikipedia.org/wiki/Bank_holiday>
+        * [LSE business days](https://www.londonstockexchange.com/equities-trading/business-days)
```
KJhellico:
```suggestion
        * [LSE business days](https://web.archive.org/web/20260617134450/https://www.londonstockexchange.com/equities-trading/business-days)
```

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,75 @@
...
+    supported_categories = (HALF_DAY, PUBLIC)
+    start_year = 2000
+
+    def _populate_public_holidays(self):
+        super()._populate_public_holidays()
+
+        for dt in tuple(self.keys()):
+            if self._is_weekend(dt):
+                self.pop(dt)
+
```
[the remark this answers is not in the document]
PPsyrius: "ENG" subdivision-specific holidays are always on weekdays anyway - so this is a non-issue: 
```python
    def _populate_common(self):
        # Easter Monday.
        self._add_easter_monday(tr("Easter Monday"))

        if self._year >= 1971:
            # Late Summer Bank Holiday.
            self._add_holiday_last_mon_of_aug(tr("Late Summer Bank Holiday"))
        else:
            # Pentecost Monday.
            self._add_pentecost_monday(tr("Whit Monday"))
```
```python
    def _populate_subdiv_eng_public_holidays(self):
        self._populate_common()
```

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,75 @@
...
+    parent_entity_subdivision_code = "ENG"
+    supported_categories = (HALF_DAY, PUBLIC)
+    start_year = 2000
+
+    def _populate_public_holidays(self):
+        super()._populate_public_holidays()
+
+        for dt in tuple(self.keys()):
+            if self._is_weekend(dt):
+                self.pop(dt)
```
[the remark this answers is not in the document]
PPsyrius: "ENG" subdivision-specific holidays are always on weekdays anyway - so this is a non-issue: 
```python
    def _populate_common(self):
        # Easter Monday.
        self._add_easter_monday(tr("Easter Monday"))

        if self._year >= 1971:
            # Late Summer Bank Holiday.
            self._add_holiday_last_mon_of_aug(tr("Late Summer Bank Holiday"))
        else:
            # Pentecost Monday.
            self._add_pentecost_monday(tr("Whit Monday"))
```
```python
    def _populate_subdiv_eng_public_holidays(self):
        self._populate_common()
```

holidays/financial/london_stock_exchange.py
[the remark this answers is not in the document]
KJhellico: Looks like the bot is right. New Year's Day, Christmas Day, and Boxing Day, which are added in `_populate_subdiv_holidays` (i.e., after `_populate_public_holidays`), are not affected by removal. It's a UK implementation detail that I missed.

holidays/financial/london_stock_exchange.py
```diff
@@ -0,0 +1,76 @@
...
+    supported_categories = (HALF_DAY, PUBLIC)
+    start_year = 2000
+
+    def _populate(self, year: int) -> None:
+        super()._populate(year)
+
+        if self.observed:
+            for dt in tuple(self.keys()):
+                if dt.year == year:
+                    self._move_holiday(dt, rule=SAT_SUN_TO_NONE, show_observed_label=False)
```
KJhellico:
```suggestion
    def _populate_subdiv_holidays(self) -> None:
        super()._populate_subdiv_holidays()

        for dt in tuple(self.keys()):
            if self._is_weekend(dt):
                self.pop(dt)
```
That seems simpler.
But we definitely need to think of some universal approach, since this is typical of practically all financial markets...

KJhellico (approved):
LGTM. 👍

holidays/countries/united_kingdom.py
```diff
@@ -104,7 +105,7 @@ def _populate_common(self):
             self._add_pentecost_monday(tr("Whit Monday"))
 
     def _populate_subdiv_holidays(self):
-        if self.subdiv != "SCT":
+        if PUBLIC in self.categories and self.subdiv != "SCT":
```
arkid15r, line 108: Would it make more sense to move this into `_populate_public_holidays` ?

KJhellico: Basically, yes. Traditionally, we move code that somehow uses `self.subdiv` into `_populate_subdiv_holidays` (e.g., USA).

arkid15r: The `if PUBLIC in self.categories` part didn't look familiar enough to me, so I wanted to check. It might also look a bit cleaner that way.

arkid15r (approved):

PR #3605, opened 2026-06-05
Update Maldives holidays: add l10n support

aadesh-shrivastava:
## Proposed change
Resolves #3110
Added localisation files for country maldives 
official source info url : https://presidency.gov.mv/Press/Article/13097
I tried to find more related articles but was only able to find one trusted source 

## Type of change

- [ ] New country/market holidays support (thank you!)
- [x] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] I've run `make check` locally; all checks and tests passed.

PR #3389, opened 2026-03-24
Update Bhutan holidays: add l10n support

Devika9705:
## Proposed change

* Adds initial localization support for Bhutan using `gettext` (`tr`)
* Wraps selected holiday names to enable future Dzongkha (dz) translations
* Keeps changes minimal and aligned with existing localization patterns in the repository

Related to: #3152

## Type of change

* [ ] New country/market holidays support (thank you!)
* [x] Supported country/market holidays update (calendar discrepancy fix, localization)
* [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
* [ ] Dependency update (version deprecation/pin/upgrade)
* [ ] Bugfix (non-breaking change which fixes an issue)
* [ ] Breaking change (a code change causing existing functionality to break)
* [ ] New feature (new `holidays` functionality in general)

## Checklist

* [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
* [x] I've run `make check` locally; all checks and tests passed.

PR #3117, opened 2025-12-08
Update Bangladesh holidays: add l10n support

NehaaGaikwad:
## Proposed change

Added Translations for Bangladesh Holidays in Bengali. 

## Type of change

- [ ] New country/market holidays support (thank you!)
- [x] Supported country/market holidays update (calendar discrepancy fix, localization)
- [ ] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] I've run `make check` locally; all checks and tests passed.

PR #3660, opened 2026-07-01
Update Mauritania holidays: add l10n support

TaniaRocks:
## Proposed change

Added Arabic localization for Mauritania (MR) holidays.

### Changes
- Added `holidays/locale/ar/LC_MESSAGES/MR.po`
- Added Arabic translations for Mauritania holiday names
- Updated `README.md` to include the new localization

Related to #3152

## Type of change

## Checklist


KJhellico:
@TaniaRocks, I recommend looking at some examples of adding l10n support: #3091, #3117, #3389, #3605.

holidays/countries/mauritania.py
```diff
@@ -41,32 +44,32 @@ def __init__(self, *args, islamic_show_estimated: bool = True, **kwargs):
 
     def _populate_public_holidays(self):
         # New Year's Day.
-        self._add_new_years_day("New Year's Day")
+        self._add_new_years_day(tr("New Year's Day"))
 
         # Labor Day.
-        self._add_labor_day("Labor Day")
+        self._add_labor_day(tr("Labor Day"))
```
PPsyrius: For Mauritania, since Arabic (`ar`) is their default language, that should've been added here directly instead
```suggestion
        self._add_labor_day(tr("عيد العمال"))
```

holidays/locale/ar/LC_MESSAGES/MR.po
```diff
@@ -0,0 +1,65 @@
...
+#           dr-prodigy <dr.prodigy.github@gmail.com> (c) 2017-2023
+#           ryanss <ryanssdev@icloud.com> (c) 2014-2017
+#  Website: https://github.com/vacanza/holidays
+#  License: MIT (see LICENSE file)
+#
+# Mauritania holidays.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Holidays 0.79\n"
```
PPsyrius:
```suggestion
"Project-Id-Version: Holidays 0.100\n"
```

holidays/locale/ar/LC_MESSAGES/MR.po
```diff
@@ -0,0 +1,65 @@
...
+"Report-Msgid-Bugs-To: l10n@vacanza.dev\n"
+"POT-Creation-Date: 2025-06-29 16:42+0530\n"
+"PO-Revision-Date: 2025-06-29 16:42+0530\n"
+"Last-Translator: Kriti Birda <workwithkritibirda@gmail.com>\n"
```
PPsyrius: This needs to be your own contact info

holidays/countries/mauritania.py
```diff
@@ -24,6 +26,7 @@ class Mauritania(HolidayBase, InternationalHolidays, IslamicHolidays):
     """
 
     country = "MR"
+    default_language = "ar"
     # %s (estimated).
     estimated_label = "%s (estimated)"
```
PPsyrius: `supported_languages` need to be included as well
```suggestion
    estimated_label = "%s (estimated)"
    supported_languages = ("ar", "en_MR", "en_US")
```

README.md
```diff
@@ -1134,10 +1134,10 @@ any) in brackets, available languages and additional holiday categories. All cou
 <td></td>
 </tr>
 <tr>
-<td>Mauritania</td>
-<td>MR</td>
-<td></td>
+<td>Mauritius</td>
+<td>MU</td>
 <td></td>
+<td><strong>en_MU</strong>, en_US</td>
```
PPsyrius: Mauritania's 2-letter ISO code is `MR` here
```suggestion
<td><strong>ar</strong>, en_MR, en_US</td>
```

PPsyrius (commented):
IMO you'll need to include the Arabic source(s) you used for the holiday names as well (any news websites are fine, but government ones are preferred)

I would recommend running `make l10n` after `make setup` since that auto-populates the `.po` files for your l10n work here instead of copying them manually from elsewhere - see https://holidays.readthedocs.io/en/latest/contributing/#running-tests for more details.

For the list of supported languages, I think we can go with these 3 for now (or more if you happen to find more/be a native speaker for certain languages)
- `ar`: Mauritania's national language
- `en_MR`: previous holiday names wording, local to Mauritania
- `en_US`: library-wide fallback standard, for naming see [`egypt.py`](https://github.com/vacanza/holidays/blob/af39ce7e24c47f3b01101b497d697f50a754d95f/holidays/countries/egypt.py#L121) for example - note that holiday names comment above needs to be standardised to `en_US` ones as well.

Aside from that, `test_l10n_default` (for `ar`), `test_l10n_en_mr`, and `test_l10n_en_us` needs to be added to at the very end of the [`test_mauritania.py`](https://github.com/vacanza/holidays/blob/af39ce7e24c47f3b01101b497d697f50a754d95f/tests/countries/test_mauritania.py#L4) as well, see [`test_singapore.py`](https://github.com/vacanza/holidays/blob/af39ce7e24c47f3b01101b497d697f50a754d95f/tests/countries/test_singapore.py#L253) implementation for example - let's use year 2024 for this and remove the old `test_2024` test case

KJhellico:
> * `en_MR`: previous holiday names wording, local to Mauritania

I doubt that English is used to any significant extent in Mauritania. The current English names seem to be just common terms.

TaniaRocks:
> > * `en_MR`: previous holiday names wording, local to Mauritania
> 
> I doubt that English is used to any significant extent in Mauritania. The current English names seem to be just common terms.

Thanks for the clarification. I added the localization following the existing naming pattern in the project. If different English names would be more appropriate for Mauritania, I'd be happy to revise them.

KJhellico:
> I added the localization following the existing naming pattern in the project. If different English names would be more appropriate for Mauritania, I'd be happy to revise them.

I mean, English isn't widely used in Mauritania, so en_MR locale is redundant.

eeshsaxena (commented):
Thanks for adding l10n for Mauritania! The mechanics look solid - `default_language = "ar"` matches Mauritania's official language, both non-default locales (`en_MR`, `en_US`) are fully translated, the `#.` context comments come through for every entry, and the test split into `test_l10n_default` / `test_l10n_en_us` / `test_l10n_en_mr` follows the usual pattern.

One consistency thing I noticed while comparing the two English `.po` files. `en_MR` and `en_US` agree on every holiday except the two Islamic-calendar ones:

| msgid | `en_US` | `en_MR` |
|---|---|---|
| رأس السنة الهجرية | `Islamic New Year` | `Muharram/Islamic New Year` |
| المولد النبوي الشريف | `The Prophet's Birthday` | `Prophet Muhammad's Birthday` |

Since both are English locales and there's no real US-vs-Mauritania English naming difference for these, the split looks unintentional. It might also be worth aligning them with the names already used elsewhere in the library, which would keep these holidays consistent for anyone comparing countries in English:

- `Islamic New Year` - used in 23 `en_US` locale files (vs. `Muharram/Islamic New Year`, which is new here).
- `Prophet's Birthday` - used in 39 `en_US` locale files; this PR uses `The Prophet's Birthday` (`en_US`) and `Prophet Muhammad's Birthday` (`en_MR`), neither of which matches the common form.

So aligning both locales to `Islamic New Year` and `Prophet's Birthday` would match the established convention. Not a blocker, and please disregard if the regional difference is deliberate - just flagging it since the rest of the file lines up so cleanly. Deferring to the maintainers on the naming call.

eeshsaxena (commented):
Thanks for working on this. The Arabic default plus the en_US/en_MR split reads well and the test restructuring into `test_l10n_default`/`test_l10n_en_us`/`test_l10n_en_mr` follows the usual pattern. A couple of things I noticed while reading through:

**1. `en_US` and `en_MR` disagree on two holiday names.** For the same two Arabic msgids, the two English locales translate differently:

| Arabic msgid | en_US | en_MR |
|---|---|---|
| `رأس السنة الهجرية` | Islamic New Year | Muharram/Islamic New Year |
| `المولد النبوي الشريف` | The Prophet's Birthday | Prophet Muhammad's Birthday |

The `en_US` strings match the updated source comments in `mauritania.py` (`# Islamic New Year.`, `# The Prophet's Birthday.`), while `en_MR` keeps the pre-change wording (`Muharram/Islamic New Year`, `Mawlid al-Nabi` -> `Prophet Muhammad's Birthday`). Is the difference intentional (some Mauritania-specific English usage), or should `en_MR` match the source/`en_US` wording? If there's no deliberate regional distinction, aligning them would avoid two English variants of the same holiday drifting apart.

**2. `.po` header comments.** Comparing with existing locales (e.g. `en_US/LC_MESSAGES/DZ.po` -> `# Algeria holidays en_US localization.`):
- `en_US/LC_MESSAGES/MR.po` currently has `# Mauritania holidays.` - for a non-default locale this is usually `# Mauritania holidays en_US localization.`
- `en_MR/LC_MESSAGES/MR.po` has `# Mauritania holidays locale localization.` - the `locale` looks like a leftover template placeholder that should read `en_MR`; there's also a doubled blank `#` line just below it.

The `ar/LC_MESSAGES/MR.po` header (`# Mauritania holidays.`) is correct as-is since `ar` is the default language.

None of this is blocking from my side - point 1 is the one I'd want a maintainer's read on. Deferring to @PPsyrius for the final call.

PPsyrius (approved):
LGTM 🇲🇷

KJhellico (approved):

arkid15r (commented):
Awesome, thanks for the contribution @TaniaRocks 🚀

PR #3641, opened 2026-06-21
Update documentation: enhance iCalendar export examples

davidhedlund:
## Proposed change

* Enhanced existing documentation examples for better clarity.
* Added a practical usage example to the README.md to address a gap in the current documentation.

Note: This PR must be merged first:
* https://github.com/vacanza/holidays/pull/3640

## Type of change

- [ ] New country/market holidays support (thank you!)
- [ ] Supported country/market holidays update (calendar discrepancy fix, localization)
- [x] Existing code/documentation/test/process quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Dependency update (version deprecation/pin/upgrade)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (a code change causing existing functionality to break)
- [ ] New feature (new `holidays` functionality in general)

## Checklist

- [x] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).
- [x] I've run `make check` locally; all checks and tests passed.

PR #3665, opened 2026-07-02
Update CONTRIBUTORS to include myself

davidhedlund:
Joining the contributor list at the request of fellow contributors.

>  @davidhedlund, you can also add your name to CONTRIBUTORS file - not just for this PR, but for your persistent promotion of iCalendar support, which has inspired our progress in this. 👍
* https://github.com/vacanza/holidays/pull/3641

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -43,6 +43,7 @@ Colin Watson
 Dan Gentry
 Daniel Musketa
 Daniël Niemeijer
+David Hedlund
 David Hotham
 Devaraj K
 Devika Shukla


KJhellico:
It would be better to include this in #3641.

PPsyrius (commented):
LGTM 👍 

Though IMO you can probably add this directly in PR #3641 as well

davidhedlund:
> It would be better to include this in #3641.

Got it!

* https://github.com/vacanza/holidays/pull/3641/changes/24efe2d695008973a54e4cd03776a3ffb93c830e

Issue #3670, opened 2026-07-03
README.md: Add `uv`, and `git` sections


davidhedlund:
I propose that we update https://github.com/vacanza/holidays/blob/dev/README.md#install

# Current https://github.com/vacanza/holidays/blob/dev/README.md#install

The latest stable version can always be installed or updated via pip:

``` shell
pip install --upgrade holidays
```

The latest development (dev) version can be installed directly from GitHub:

``` shell
pip install --upgrade https://github.com/vacanza/holidays/tarball/dev
```

All new features are always first pushed to dev branch, then released on main branch upon official
version upgrades.

# Proposed https://github.com/vacanza/holidays/blob/dev/README.md]

## uv
This it applies to all OSes.

### Installation

Refer to https://docs.astral.sh/uv/getting-started/installation/ for how to install uvx.

### Usage
* Evaluate python code: `uvx --with holidays python3`
* Execute python file: `uvx --with holidays python3 file.py`
* Generate calendar file with holidays-ics: `uvx --from holidays holidays-ics US`

#### Installation

```
git clone https://github.com/vacanza/holidays.git
cd holidays/

# Enable language support
python3 -c "import polib; from pathlib import Path; [polib.pofile(str(f)).save_as_mofile(str(f.with_suffix('.mo'))) for f in Path('holidays/locale').rglob('*.po')]"
```

###### Step 1
```
sudo apt update && sudo apt install -y git python3-polib python3-holidays
```

###### Step 2
See the [git](#git) section.

### pip
**Note:** This method is not recommended for Debian, Ubuntu, or their derivatives. While it requires `pip3`, you must use the `--break-system-packages` flag to bypass system restrictions. Avoid using `--break-system-packages` for routine package installation because it can create conflicts with packages managed by apt. [Read more here](https://github.com/vacanza/holidays/blob/dev/docs/PIPWARNING) [add https://github.com/vacanza/holidays/blob/dev/docs/PIPWARNING]

The latest stable version can always be installed or updated via pip:

``` shell
pip install --upgrade holidays
```

The latest development (dev) version can be installed directly from GitHub:

``` shell
pip install --upgrade https://github.com/vacanza/holidays/tarball/dev
```

#### [PIPWARNING](https://github.com/vacanza/holidays/blob/dev/docs/PIPWARNING)
```
$ pip install --upgrade holidays
Command 'pip' not found, but can be installed with:
sudo apt install python3-pip
$ sudo apt install python3-pip
[...]

PR #3677, opened 2026-07-04
Update README.md: Added `uv` section

davidhedlund:
## Proposed change

* There is an inconsistency in the documentation: the [GitHub README](https://github.com/vacanza/holidays/blob/dev/README.md#install) states that pip should be used but does not mention uv. Conversely, the [ReadTheDocs Examples](https://holidays.readthedocs.io/en/latest/examples/) page recommends using uv instead of pip. We should align these to avoid confusing users.
* On Debian and Ubuntu distributions, `pip` is not pre-installed, and `pip3` requires manual installation along with the `--break-system-packages` flag to run. Using `uv` bypasses these hurdles completely with a simple, hassle-free installation.

Partially fixes:
* https://github.com/vacanza/holidays/issues/3670

## Type of change

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,22 +44,65 @@ specific date is a holiday as fast and flexible as possible.
   </tr>
 </table>
 
-## Install
+## Installation & execution
 
-The latest stable version can always be installed or updated via pip:
+The `holidays` package can be installed and run using either `pip` or `uv`.
 
-``` shell
+### Using pip
+
+#### Installation
+
+Install or update the latest stable version via pip:
+
+```shell
 pip install --upgrade holidays
+
 ```
 
-The latest development (dev) version can be installed directly from GitHub:
+To try out the latest features, install the development version directly from GitHub:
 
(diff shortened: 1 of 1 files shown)


KJhellico:
> * There is an inconsistency in the documentation: the [GitHub README](https://github.com/vacanza/holidays/blob/dev/README.md#install) states that pip should be used but does not mention uv. Conversely, the [ReadTheDocs Examples](https://holidays.readthedocs.io/en/latest/examples/) page recommends using uv instead of pip. We should align these to avoid confusing users.

There's no inconsistencies here. You're mixing recommendations for a user who needs a library (developer) with those for a user who just needs a tool to generate iCalendar files. And when [I said](https://github.com/vacanza/holidays/issues/3670#issuecomment-4883712434) that `uv` should be mentioned, I simply meant that it would be helpful to mention `uv add holidays` alongside `pip install holidays` for those who use `uv`.

PR #3704, opened 2026-07-17
Update iCal exporter: fix dropping or duplicating events

chuenchen309:
## Proposed change

`ICalExporter.generate()` merges consecutive same-name holidays into a single VEVENT with a `DURATION`, then advanced the outer date index with `i += days`. But `days` held only the **last** name iterated on a given date, so when several names share a start date with different run lengths the index moved by the wrong amount:

- the longer run sorts first → index under-advances → that holiday is re-emitted on its second day (a duplicate, overlapping VEVENT);
- the longer run sorts last → index over-advances → a holiday starting on the skipped day is dropped entirely.

`HolidayBase.get_list()` returns names sorted alphabetically, so which failure mode occurs is arbitrary.

```python
from holidays.ical import ICalExporter
h = HolidayBase(expand=False)
h["2024-01-01"] = "AAA"; h["2024-01-01"] = "ZZZ"; h["2024-01-02"] = "AAA"
ICalExporter(h).generate().count("SUMMARY:AAA")   # -> 2 (duplicate); expected 1
# and with AAA on Jan 1+2, BBB on Jan 2 -> BBB is dropped entirely
```

The existing `test_single_holiday_multiple_date_continuous_overlap` (UA 2005) has this exact shape, but its Cyrillic ordering happens to place the two-day name last, so it currently masks the bug.

**Fix:** advance one date at a time and skip a name that merely *continues* the previous same-year day's run (it was already emitted with its `DURATION` when the run started). The look-back only reads dates already in the key set, so it never populates an absent date on an expandable `HolidayBase` — which would otherwise mutate the calendar mid-render (caught by `test_save_ics_pathlib_path`). Existing merge behaviour and the year-boundary split (`test_single_holiday_multiple_date_cross_year_boundary`) are unchanged.

Adds a regression test (a two-day run sharing its start date with a single-day holiday), red before the fix and green after. `holidays/ical.py` stays at **100% coverage**.

*Verification:* `ruff check` / `ruff format --check` clean, and `pytest tests/test_ical.py` — 27 passed, `ical.py` 100% coverage. (Ran the affected iCal suite rather than the full multi-thousand-file country suite, which an iCal-engine change doesn't touch.)

*Disclosure:* prepared with AI assistance (Claude Code, Claude Opus 4.8). I reproduced the bug, wrote the failing regression test first, verified the fix (including the no-mutation property and 100% coverage), and reviewed the diff before opening.

## Type of change

## Checklist

- [x] I've run the relevant `make check` steps locally (`ruff` + `tests/test_ical.py` at 100% coverage); all passed. The full country test suite is unaffected by this iCal-engine change.


sanmaxdev (commented):
The date-by-date iteration preserves each distinct event while skipping only same-name continuations already represented by a duration. The regression covers the shared-start case with unequal run lengths.

Passed:
- `uv run --no-sync pytest -q tests/test_ical.py`
- `uv run --no-sync ruff check holidays/ical.py tests/test_ical.py`
- `uv run --no-sync ruff format --check holidays/ical.py tests/test_ical.py`
- `git diff origin/dev...HEAD --check`

PPsyrius (dismissed):
LGTM 🗓️

KJhellico (dismissed):

arkid15r (approved):

arkid15r (commented):
@chuenchen309 thanks for improving this 👍

PPsyrius (approved):

PR #3720, opened 2026-07-25
Update l10n: general unification

KJhellico:
## Proposed change

Various changes as preparation for the migration to "one file per locale" l10n model, including:
- unify en_US holiday names translations (ANZAC Day, Day of Arafah, Diwali, Islamic New Year, National Independence Day)
- unify Arabic holiday names translations (Independence Day, New Year's Day, Yom Kippur)
- unify Dutch holiday names translations
- unify l10n comments ("observed label", "substituted date format")

## Type of change

- [ ] Documentation update
- [ ] Test suite update
- [ ] Development process update (CI/CD, project configuration, internal tooling)

## Checklist


holidays/countries/libya.py
```diff
@@ -118,13 +118,13 @@ def _populate_workday_holidays(self):
             # Italian Forces Evacuation Day.
             self._add_holiday_oct_7(tr("عيد إجلاء الطليان"))
 
-            # Islamic New Year.
+            # Islamic New Year Day.
```
PPsyrius, line 121: Hmm, pretty sure it's just "Islamic New Year" for Palestine's case below - and Chinese/Lunar ones are "{insert} New Year" respectively at the moment.

I do personally prefer the version without " Day"-suffix for the non-International Jan 1st cases as that's how I've seen it written elsewhere i.e. https://en.wikipedia.org/wiki/Islamic_New_Year and https://en.wikipedia.org/wiki/Lunar_New_Year vs https://en.wikipedia.org/wiki/New_Year's_Day  - though I could be wrong here

KJhellico: I used translation "Islamic New Year Day" for "عيد رأس السنة الهجرية", which is used in several countries, as opposed to "رأس السنة الهجرية" (more common "Islamic New Year").

KJhellico: A more literal translation would be "Feast of Islamic New Year", but that didn't seem quite suitable to me. 🤷‍♂️

PPsyrius (dismissed):
LGTM 🌐

arkid15r (dismissed):
💪

arkid15r (approved):

PPsyrius (approved):

PR #3777, opened 2026-08-20
Fix unbounded functools.cache retention in Burmese and Thai calendars

wu-zhao-min:
## Proposed change
`functools.cache` keys on all arguments. `_get_start_date` (Burmese and Thai) and `buddhist_sabbath_dates` (Thai) were `@cache`'d instance methods, so `self` was part of every cache key. That retained each calendar instance in the unbounded cache and added a new entry per instance.

Make these year-only helpers `@staticmethod` + `@cache` so the cache keys only on `year`.

## Type of change

- [ ] Documentation update
- [ ] Test suite update

## Checklist

- [ ] I've read and followed the [contributing guidelines](https://github.com/vacanza/holidays/blob/dev/CONTRIBUTING.md).

### DIFF
--- a/holidays/calendars/burmese.py
+++ b/holidays/calendars/burmese.py
@@ -122,6 +122,8 @@ class _BurmeseLunisolar:
 
         return date(y, m, d)
 
+    # @staticmethod so functools.cache keys only on `year`. As an instance method,
+    # `self` would be part of the key, which retains instances and grows the cache unboundedly.
     @staticmethod
     @cache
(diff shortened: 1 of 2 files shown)


KJhellico:
@wu-zhao-min, where and how does a memory leak occur here?

KJhellico (approved):
LGTM.

PPsyrius (approved):
LGTM 🛠️

Issue #3733, opened 2026-07-28
Multi-Selection Interface for Consolidating and Downloading Holiday Calendars in Bulk


davidhedlund:
**Is your feature request related to a problem?**

Currently, the [Download Holiday Calendars](https://holidays.readthedocs.io/en/latest/downloads/) page only allows users to access one calendar at a time. This is inefficient when users need to aggregate holiday schedules across multiple regions, categories, or years, as it requires repetitive navigation and manual tracking.

**Describe the solution you'd like**

I propose adding a multi-selection interface to the "Download Holiday Calendars" page. Users should be able to select multiple items simultaneously and click a new **"List calendars"** button to generate a consolidated table. This table would display the available formats (ICS, JSON, Webcal) for all selected items, leveraging the index structure proposed in:
* https://github.com/vacanza/holidays/issues/3684

**Example Workflow:**

1. **Selection:** User selects *Denmark, Sweden, United States* (Countries), *Public, Optional* (Categories), and *ICS, JSON, Webcal* (proposed Format listing: #3669)
2. **Action:** User clicks "List calendars."
3. **Output:** A summary table is generated dynamically:

| Language / Region | Public | Optional |
| :--- | :--- | :--- |
| **Denmark** | [ICS](https://vacanza.github.io/holidays/downloads/ics/countries/DK/ALL_da_public.ics) \| [JSON](https://vacanza.github.io/holidays/downloads/ics/countries/DK/ALL_da_public.json) \| Webcal[1] | [ICS](https://vacanza.github.io/holidays/downloads/ics/countries/DK/ALL_da_optional.ics) \| [JSON](https://vacanza.github.io/holidays/downloads/ics/countries/DK/ALL_da_optional.json) \| Webcal[2] |
| **Sweden** | [ICS](https://vacanza.github.io/holidays/downloads/ics/countries/SE/ALL_sv_public.ics) \| [JSON](https://vacanza.github.io/holidays/downloads/ics/countries/SE/ALL_sv_public.json) \| Webcal[3] | [ICS](https://vacanza.github.io/holidays/downloads/ics/countries/SE/ALL_sv_optional.ics) \| [JSON](https://vacanza.github.io/holidays/downloads/ics/countries/SE/ALL_sv_optional.json) \| Webcal[4] |
[...]

PR #3800, opened 2026-09-02
Update calendar download page: add multi-select support

VishalRaj0:
## Proposed change

Adds multi-select support to the Download Holiday Calendars page, closing #3733.

Previously, the downloads page only supported selecting one country/region at a time. This PR reworks the page so users can:
- Select multiple countries/regions via a searchable multi-select dropdown, with removable selection chips
- Select multiple categories (e.g. Public, Hebrew, Bank) across the selected entities
- Click "List calendars" to view a results table (Region × Category) with ICS, JSON, and Webcal links per cell

For ICS and JSON, downloads are scoped to the selected Year Range: the full static index file (covering 2015-2035) is fetched, filtered client-side to the chosen years, and served as a generated download named e.g. `India-public-2025-2026.ics`. Webcal links are intentionally left unscoped by year and continue to point at the full static file, since a live calendar subscription needs a stable URL to keep re-fetching from — a year-filtered result only exists transiently in the browser and can't be subscribed to.

Also fixed along the way:
- The country/region multi-select dropdown closed itself whenever a checkbox inside it was clicked, because `@click.outside` was bound to the trigger button instead of the wrapping container.
- The "unavailable" table cell used `x-if` on a plain `<span>` instead of a `<template>`, so it silently had no effect and always rendered.
- The results table forced unnecessary horizontal scrolling due to a stale `white-space: nowrap` rule left over from the old single-select design, plus a `min-width` on the table itself.
- A long country name (e.g. "Albania") could wrap mid-word once the Region column got squeezed narrow; the column now has a sensible minimum width and disallows mid-word breaks.

## Type of change

- [ ] Documentation update
- [ ] Test suite update
## Checklist


KJhellico:
There is a need to keep the calendar preview feature.

PR #3816, opened 2026-09-11
fix: add Assam Bahag Bihu holidays

Drona-jadhav7:
## Description

This PR updates the Assam holiday calendar to correctly represent **Bahag Bihu** as a three-day public holiday on **April 14, 15, and 16**, while keeping **Magh Bihu** on January 14.

The issue was identified while auditing the `holidays` library against authoritative holiday references using [HolidayLens.](https://github.com/Drona-jadhav7/HolidayLens)

## Issue

The Assam holiday calendar in the `holidays` library does not correctly represent **Bahag Bihu**.

Bahag Bihu is observed as a **three-day public holiday in Assam on April 14, 15, and 16**, but the library did not provide this Assam-specific three-day public-holiday representation.

Instead, `Bahag Bihu` was handled through the national optional-holiday mechanism, which does not accurately represent its status as an Assam public holiday.

This results in incorrect holiday data for users who use:

```python
India(subdiv="AS")
```

The issue is particularly important for applications that depend on the library for accurate state-level holiday calendars.
Official Reference

The implementation is based on the official Government of Assam – Calendar & Holiday List 2026, which lists Bohag Bihu from April 14–16 under Gazetted Holidays.

The calendar states that it is based on Government of Assam Notification No. E-221561/440 dated 17 November 2025 and was compiled by the National Informatics Centre (NIC), Assam.

### Official reference:
[Government of Assam – Calendar & Holiday List 2026](https://tinsukia.assam.gov.in/sites/default/files/public_utility/cal2026.pdf?utm_source=chatgpt.com)

## Implementation

The implementation makes Bahag Bihu Assam-specific.

### Changes made

* Removed `Bahag Bihu` from the national optional holiday handling.
* Added `Bahag Bihu` to Assam's public holidays.
* Added the holiday on:

  * April 14
  * April 15
  * April 16
* Kept the existing Assam `Magh Bihu` holiday on January 14.
* Updated the relevant tests and localization expectations.

The existing holiday helpers and translation mechanism are used, keeping the implementation consistent with the project's architecture.

April 14 can also correctly coexist with other holidays occurring on the same date, such as Dr. B. R. Ambedkar's Jayanti.

## Challenges

The change affected more than the Assam holiday definition itself.

Since `Bahag Bihu` is a translated holiday name, moving it from the optional-holiday path to Assam's public-holiday path affected localization data and localization tests across multiple languages.

Another challenge was ensuring that the change did not unintentionally add Bahag Bihu to India's national calendar while adding the correct dates for Assam.

## Resolution

The issues were addressed by:

* Removing Bahag Bihu from the national optional-holiday handling.
* Adding it directly to Assam's public-holiday population.
* Using the existing `tr()` localization mechanism.
* Regenerating the localization files.
* Updating the affected localization tests.
* Verifying the Assam calendar across multiple years.
* Updating the branch onto the current upstream development base.

## Result

For Assam, the resulting calendar correctly includes:

```text
2026-01-14    Magh Bihu
2026-04-14    Bahag Bihu; Dr. B. R. Ambedkar's Jayanti
2026-04-15    Bahag Bihu
2026-04-16    Bahag Bihu
```

For the national calendar, Bahag Bihu is no longer incorrectly included:

```text
India()       → Bahag Bihu is not added
India(AS)     → Bahag Bihu is added on April 14–16
```

## Testing

The implementation was verified by checking the `India` calendar with the Assam subdivision across multiple years.

The tests verify that:

* `Magh Bihu` remains on January 14.
* `Bahag Bihu` occurs on April 14, 15, and 16 in Assam.
* Bahag Bihu is not added to the national calendar.
* April 14 correctly combines Bahag Bihu with other holidays on that date.
* Localization expectations are updated for the new holiday behavior.


holidays/countries/india.py
```diff
@@ -522,6 +520,13 @@ def _populate_subdiv_ap_public_holidays(self):
     def _populate_subdiv_as_public_holidays(self):
         # Magh Bihu.
         self._add_pongal(tr("Magh Bihu"))
+
+        # Bahag Bihu.
+        name = tr("Bahag Bihu")
+        self._add_holiday(name, date(self._year, APR, 14))
+        self._add_holiday(name, date(self._year, APR, 15))
+        self._add_holiday(name, date(self._year, APR, 16))
+
```
PPsyrius, line 529: Why are we removing this from the countrywide public holidays list? 👀 

If this is only for a single year, consider doing this via `IndiaStaticHolidays` instead:
```
special_as_public_holidays = {
        2026: (
            (APR, 14, name_bahag_bihu),
            (APR, 15, name_bahag_bihu),
            (APR, 16, name_bahag_bihu),
        ),
}
```

If not, then Assam's specific observance should've been added in other way

Drona-jadhav7: @PPsyrius Bohag Bihu is an annual Assamese festival marking the Assamese New Year. The festival itself is recurring, but the relevant point for this PR is the Assam government's declaration of public holidays.

My intention in this PR was to correct the Assam public-holiday coverage based on that official notification. I removed the existing national Bahag Bihu entry because I understood it to be incorrectly classified as a national holiday, and added the Assam-specific Apr 14–16 public holidays.

If the existing national Bahag Bihu entry is intended to represent the restricted/optional India-wide calendar, rather than an Assam public holiday, I agree that it should be handled separately. In that case, I can preserve that entry and implement the Assam public holidays through the appropriate subdivision-specific mechanism.
Please let me know if that is the intended model.

Drona-jadhav7: I agree that using the IndiaStaticHolidays approach is more appropriate for representing the year-specific Assam public-holiday declaration.Should I update the implementation accordingly and adjust the tests?

KJhellico: Looking at the lists of Assam holidays from the past few years, it's obvious that there are three days off - New Year itself, plus the day before and the day after. And of course, New Year’s Day isn’t strictly on Apr 15, due to difference in leap years between the Indian solar calendar and the Gregorian calendar.

Drona-jadhav7: @KJhellico Thanks for the clarification. I looked into the current India implementation and the existing `_add_vaisakhadi()` helper. It uses the Gregorian leap-year rule, which does not always match the actual Assamese New Year date.

For Bahag Bihu, would you prefer that I:

1. keep it under `IndiaStaticHolidays.special_as_public_holidays` and add the official three-day Assam holiday dates year-by-year, or
2. introduce/reuse a calendar-based calculation for the Assamese New Year and derive the day before and after from it?

I’m leaning toward the first option if the project prefers year-specific static data for Assam government holidays, but I’d like to follow the library’s intended architecture and coding style rather than introduce unnecessary calendar logic.

KJhellico: > does not always match the actual Assamese New Year date

In which years are the differences?

Drona-jadhav7: @KJhellico I checked this more carefully, including historical and future calendar sources. I was too quick earlier to call a mismatch.
The important distinction seems to be that `_add_vaisakhadi()` calculates April 14 in Gregorian leap years and April 15 otherwise, while Bahag Bihu is based on 1st Bohag / the Assamese solar calendar. The Assam government holiday lists then make the surrounding days public holidays as well.
I don't want to claim a specific mismatch without authoritative year-by-year data. Would you recommend using `_add_vaisakhadi()` as the New Year date and adding the adjacent two days, or should Bahag Bihu remain year-specific static data based on the Assam government notifications?

KJhellico: Actually, this holiday has already been correctly added in #3784, along with the corresponding methods and tests. So I don't see any point in moving forward with this PR. 🤷‍♂️

PR #3821, opened 2026-09-14
Update Australia holidays: add school holidays support

gheydon:
## Proposed change

Adds the `SCHOOL` category to Australia, covering all eight states and territories.

Australia declared no `SCHOOL` category, so school holidays were unavailable everywhere in the country. This follows the shape Germany already uses: a year-and-subdivision dataset of date ranges in `holidays/calendars/australia_school.py`, expanded into individual days by `_populate_school_holidays`.

Four holiday names are used — `Term 1 school holidays`, `Term 2 school holidays`, `Term 3 school holidays`, `Summer school holidays` — naming each break by the term it follows, which is how Queensland labels its own published list. I'd happily switch to seasonal names (as NSW uses) if you prefer, though those read oddly in the tropics.

### Where the data comes from

Every state and territory publishes its own calendar; there is no national source. Each is linked in the module docstring and was taken from the department's own page — third-party school-holiday sites were checked and found to disagree with the Northern Territory's department and with each other, so none were used.

Six jurisdictions publish term dates only, so their holidays are the gaps between terms. NSW and Queensland publish the holiday periods themselves, and those are used verbatim.

Dates are **student** dates wherever a jurisdiction distinguishes them from staff dates. The NT, Tasmania, Victoria and the ACT all have teachers return before pupils.

### How it was checked

- Every date the source states with a weekday (230 of them) was verified against the calendar. One transcription error was caught this way.
- No break overlaps a term, in any state, in any year.
- Deriving Queensland's holidays from Queensland's term dates reproduces its published holidays exactly — a direct check that the derivation used for the other six is sound.

### Things I'd like your view on

1. **NSW Western division** starts Term 1 about a week later than Eastern and its summer break runs on accordingly. NSW is a single subdivision here, so I've used Eastern throughout, which covers the large majority of schools. The Western dates are transcribed and ready if you'd rather split the subdivision or special-case it the way `BY`/`Augsburg` is handled for Germany.

2. **Thai localization is not filled in.** I don't speak Thai and would rather leave the four new strings to someone who does than guess at user-facing text. This is the one thing keeping the PR in draft — `test_localization` fails on AU `th` coverage until it's done. `en_US` is filled. Happy to take a suggested translation and add it.

3. **No generator script.** Germany has one because the KMK publishes uniform PDFs. Australia has eight jurisdictions with eight page layouts, two of which block automated fetching, so this is committed data with every source URL recorded. Glad to write one if you'd prefer that shape.

4. **Coverage is uneven** because approval horizons differ — the ACT and Tasmania publish to 2027, the NT to 2032. Western Australia's 2030 onwards are published as "preliminary, to be confirmed" and are left out.

### Known limitation

NSW counts the school development days that open each term as term time, so they are not school holidays here, although students do not attend. Its published periods likewise exclude public holidays falling at a term boundary, which the `PUBLIC` category covers. This is documented in the module docstring.

## Type of change

## Checklist

`ruff check`, `ruff format` and the full suite pass — 7,938 tests — with the single exception of `tests/test_l10n.py::TestLocalization::test_localization`, which fails only on the missing Thai strings noted above.


holidays/locale/th/LC_MESSAGES/AU.po
```diff
@@ -59,6 +59,22 @@ msgstr "วันชาติออสเตรเลีย"
...
+msgid "Term 2 school holidays"
+msgstr ""
+
+#. Term 3 School Holidays.
+msgid "Term 3 school holidays"
+msgstr ""
+
+#. Summer School Holidays.
+msgid "Summer school holidays"
+msgstr ""
```
PPsyrius: Thai l10ns
```suggestion
#. Term 1 School Holidays.
msgid "Term 1 school holidays"
msgstr "วันหยุดภาคเรียนที่ 1"

#. Term 2 School Holidays.
msgid "Term 2 school holidays"
msgstr "วันหยุดภาคเรียนที่ 2"

#. Term 3 School Holidays.
msgid "Term 3 school holidays"
msgstr "วันหยุดภาคเรียนที่ 3"

#. Summer School Holidays.
msgid "Summer school holidays"
msgstr "วันหยุดภาคฤดูร้อน"
```

tests/countries/test_australia.py
```diff
@@ -2149,3 +2151,69 @@ def test_l10n_th(self):
...
+            "2026-09-18",
+            "2026-10-05",
+            "2026-12-18",
+        ):
+            self.assertNoHoliday(vic, term_day)
+
+    def test_school_holidays_run_across_new_year(self):
+        # The summer break belongs to both years it touches.
+        self.assertSubdivVicSchoolHolidayName("Summer school holidays", "2026-12-31")
+        self.assertSubdivVicSchoolHolidayName("Summer school holidays", "2027-01-01")
```
PPsyrius: Please also move the new test cases above the existing l10n tests
```suggestion
    def test_school_holidays_run_across_new_year(self):
        self.assertSubdivVicSchoolHolidayName("Summer school holidays", "2026-12-31", "2027-01-01")
```

IMO, following the exisitng Germany school holidays test might be ideal here: https://github.com/vacanza/holidays/blob/0e045800519a59a6ef2bda5dda27ed08ac5e7c46/tests/countries/test_germany.py#L342

tests/countries/test_australia.py
```diff
@@ -2149,3 +2151,69 @@ def test_l10n_th(self):
...
+
+    def test_school_and_public_categories(self):
+        both = Australia(subdiv="VIC", years=2026, categories=(PUBLIC, SCHOOL))
+        self.assertHolidayName("Australia Day", both, "2026-01-26")
+        self.assertHolidayName("Summer school holidays", both, "2026-01-26")
+
+    def test_public_category_excludes_school_holidays(self):
+        public_only = Australia(subdiv="VIC", years=2026, categories=PUBLIC)
+        self.assertHolidayName("Australia Day", public_only, "2026-01-26")
+        self.assertNoHolidayName("Term 1 school holidays", public_only)
```
PPsyrius: These 2 test cases aren't needed IMO
```suggestion
```

holidays/countries/australia.py
```diff
@@ -12,10 +12,16 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
+from datetime import date
+from holidays.calendars.australia_school import (
```
PPsyrius, line 17: Pre-commit format fixes
```suggestion
from datetime import date

from holidays.calendars.australia_school import (
```

holidays/calendars/australia_school.py
```diff
@@ -0,0 +1,312 @@
...
+
+Sources:
+    * [ACT](https://www.act.gov.au/living-in-the-act/public-holidays-school-terms-and-daylight-saving)
+    * [NSW](https://education.nsw.gov.au/schooling/calendars/future-and-past-nsw-term-and-vacation-dates)
+    * [NT](https://nt.gov.au/learning/primary-and-secondary-students/school-term-dates-in-nt)
+    * [QLD](https://education.qld.gov.au/about-us/calendar/future-dates)
+    * [SA](https://www.education.sa.gov.au/parents-and-families/term-dates-south-australian-state-schools)
+    * [TAS](https://www.decyp.tas.gov.au/learning/term-dates/)
+    * [VIC](https://www.vic.gov.au/school-term-dates-and-holidays-victoria)
+    * [WA](https://www.education.wa.edu.au/future-term-dates)
```
PPsyrius: A bit 2 cents on this - the current hardcoded implementation, while not ideal, are fine for now given the number of sources we need to keep track of 👍

holidays/calendars/australia_school.py
```diff
@@ -0,0 +1,312 @@
...
+#  License: MIT (see LICENSE file)
+
+"""Auto-generated Germany school holidays dataset from official KMK sources."""
+
+"""Australian school holidays dataset, transcribed from official sources.
+
+Each state and territory publishes its own school calendar; there is no national
+one. Six publish term dates only, so their holidays are the gaps between terms.
+New South Wales and Queensland publish the holiday periods themselves and those
+are used verbatim — deriving Queensland's from its term dates reproduces its
```
PPsyrius: Another pre-commit standardisation:
```suggestion
are used verbatim - deriving Queensland's from its term dates reproduces its
```

tests/countries/test_australia.py
```diff
@@ -12,6 +12,8 @@
 
 from unittest import TestCase
 
+from holidays.calendars.australia_school import AUSTRALIA_SCHOOL_HOLIDAYS
+from holidays.constants import PUBLIC, SCHOOL
```
PPsyrius:
```suggestion
from holidays.constants import SCHOOL
```

PPsyrius (commented):
Some minor feedbacks:

gheydon:
Thanks both — all three were real, and all three are fixed in 45cf8b6.

**Module docstring was Germany's.** Correct, and a plain copy-paste error: I started the file from `germany_school.py`'s header and took its docstring line with it. Python binds only the first string literal, so `australia_school.__doc__` read "Auto-generated Germany school holidays dataset from official KMK sources" and the Australian docstring below it was an orphaned expression. Stray line removed; `__doc__` now starts "Australian school holidays dataset, transcribed from official sources."

**Queensland's 2029 summer break was missing.** Also correct, and the more serious of the two. Queensland publishes the start — Saturday 8 December 2029 — but leaves the end "to be published when approved by the Minister", and I dropped the whole range rather than the unapproved half. The effect was that every day from 8 December 2029 was reported as a school day. The approved part is now included and stops at 31 December rather than inventing a 2030 return date, as suggested. Covered by a test that also checks 7 December is still a school day, since term 4 ends that Friday.

**Thai.** Now filled. I'd left it deliberately rather than guess at user-facing text in a language I can't check, but that was the wrong call here: this project already carries a merged Thai vocabulary for exactly this concept in the German school holidays, where `Sommerferien` is `ปิดเทอมฤดูร้อน`. "Summer school holidays" now uses that string verbatim, and the three numbered terms follow the same `ปิดเทอม` form. A native speaker's eye on the numbered ones would still be welcome — the summer one I'm confident about because it is your own existing translation.

`test_localization` passes now, so the reason this was in draft is gone. Full suite is 7,943 passing, `ruff check` and `ruff format` clean.

The four questions in the description still stand, particularly the NSW Eastern/Western division one, which is the only design decision here I'd rather not make alone.

gheydon:
Thanks @PPsyrius — all applied in 369c8a4.

**Thai.** Yours is in, replacing what I had. I'd built `ปิดเทอม...` by analogy with the German school holiday strings already in the repo, precisely because I don't speak Thai and didn't want to guess at user-facing text; having it from someone who does is much better. Tests updated to match.

**The two category-interaction tests** are gone, and that leaves `SCHOOL` as the only constant the module imports, so that import is trimmed too.

**The new-year test** is now one assertion with both dates.

**Test ordering** — the school holiday tests now sit above the localisation ones, with the three `*_school` l10n tests after the existing l10n block.

**Pre-commit fixes** (the em dash and the import blank line) went in separately as 21b77d2, before your review landed; the `Code checks` job is green on that.

On the hardcoded dataset — agreed, and thanks for saying so. A generator was the first thing I tried, but with eight jurisdictions on eight different page layouts, two of which block automated fetching outright, it would have been more fragile than the data it produced. Every source URL is recorded in the module docstring so the next person updating it knows exactly where each year came from.

The one open question from the description is still the NSW Eastern/Western division split, if you have a view on it. Eastern is used throughout at the moment; the Western dates are transcribed and ready if you'd rather they were represented.

gheydon:
Settled on the Eastern Division dates for NSW, which is what the dataset already held — fba5312 makes that an explicit, documented decision rather than an unstated one.

The reasoning, now in the module docstring: subdivisions here are states, NSW publishes two sets of term dates, so one has to stand for the whole state, and Eastern covers the great majority of schools. Western Division schools in the far west start Term 1 about a week later and their summer break runs on accordingly.

There is also a test pinning it — Eastern students return on 28 January 2027, where Western would still be on holiday until 3 February — so the choice can't drift quietly if the data is ever regenerated.

I've put it in the docstring rather than only here because the person most affected is a Western Division parent comparing this against their own children's calendar, and they'll be reading the file, not the pull request.

If you'd rather see Western represented properly at some point, the dates are transcribed and I'm happy to follow up — but that needs a subdivision that doesn't currently exist, so it felt wrong to smuggle into this PR.

That was the last open question from the description. Everything from your review is applied, tests are 7,942 passing, and pre-commit is clean.

gheydon:
A note for the record on NSW Western Division, since I said something inaccurate about it earlier in this PR and don't want it left standing.

I wrote that representing Western "needs a subdivision that doesn't currently exist". That was wrong in both directions, and looking at your own code corrected me.

**What the Western Division actually is.** It is not a region. The department publishes it as a roll of 117 named schools across 94 towns, which it calls "late start schools", dated "as of July 2026". It reaches the Queensland border at Mungindi and Boggabilla, the Victorian border at Barham and Deniliquin, and the South Australian border at Broken Hill — three separate corners with plenty of inland NSW in between that is not on it. Whole towns are always listed together, never split, but no boundary is being described and no selection criterion is published. So it genuinely cannot be modelled as geography, which is the stronger reason for using Eastern here than the head-count argument in the docstring.

**But it can still be modelled.** `Augsburg` is the shape: a named entry under `# Cities.` that differs from its parent in one category and inherits it in the other. Augsburg adds `Augsburger Hohes Friedensfest` to PUBLIC while falling back to `BY` for SCHOOL. NSW Western is that inverted — identical PUBLIC holidays to NSW, its own SCHOOL dates. One extra subdivision, not ninety-four, and a Broken Hill or Deniliquin family would select it directly.

**I'm deliberately not doing that here.** This PR is reviewed and I would rather not reopen its scope, and a subdivision that is a school roll rather than a place is a design question you should get to weigh in on separately — every existing extra entry in the library is somewhere you can stand. So I'll raise it as its own issue once this lands, and you can say no to it there without it costing this PR anything.

If you would rather it never happened, say so and I'll drop it — the dates are easy enough to re-derive from the department's page if anyone wants them later.

holidays/calendars/australia_school.py
```diff
@@ -0,0 +1,317 @@
...
+
+Sources:
+    * [ACT](https://www.act.gov.au/living-in-the-act/public-holidays-school-terms-and-daylight-saving)
+    * [NSW](https://education.nsw.gov.au/schooling/calendars/future-and-past-nsw-term-and-vacation-dates)
+    * [NT](https://nt.gov.au/learning/primary-and-secondary-students/school-term-dates-in-nt)
+    * [QLD](https://education.qld.gov.au/about-us/calendar/future-dates)
+    * [SA](https://www.education.sa.gov.au/parents-and-families/term-dates-south-australian-state-schools)
+    * [TAS](https://www.decyp.tas.gov.au/learning/term-dates/)
+    * [VIC](https://www.vic.gov.au/school-term-dates-and-holidays-victoria)
+    * [WA](https://www.education.wa.edu.au/future-term-dates)
```
KJhellico:
```suggestion
    * [ACT](https://web.archive.org/web/20260909031430/https://www.act.gov.au/living-in-the-act/public-holidays-school-terms-and-daylight-saving)
    * [NSW](https://web.archive.org/web/20260726052159/https://education.nsw.gov.au/schooling/calendars/future-and-past-nsw-term-and-vacation-dates)
    * [NT](https://web.archive.org/web/20260906145204/https://nt.gov.au/learning/primary-and-secondary-students/school-term-dates-in-nt)
    * [QLD](https://web.archive.org/web/20260312112035/https://education.qld.gov.au/about-us/calendar/future-dates)
    * [SA](https://web.archive.org/web/20251222001032/https://www.education.sa.gov.au/parents-and-families/term-dates-south-australian-state-schools)
    * [TAS](https://web.archive.org/web/20260610091723/https://www.decyp.tas.gov.au/learning/term-dates/)
    * [VIC](https://web.archive.org/web/20260913195820/https://www.vic.gov.au/school-term-dates-and-holidays-victoria)
    * [WA](https://web.archive.org/web/20260911103607/https://www.education.wa.edu.au/future-term-dates)
```

holidays/calendars/australia_school.py
```diff
@@ -0,0 +1,317 @@
...
+uneven: the Northern Territory reaches 2032, the ACT and Tasmania only 2027.
+Western Australia's 2030 onwards are marked preliminary and are left out.
+"""
+
+(
+    TERM_1_HOLIDAYS,
+    TERM_2_HOLIDAYS,
+    TERM_3_HOLIDAYS,
+    SUMMER_HOLIDAYS,
+) = range(4)
```
KJhellico:
```suggestion
(SUMMER_HOLIDAYS, TERM_1_HOLIDAYS, TERM_2_HOLIDAYS, TERM_3_HOLIDAYS) = range(4)
```

holidays/countries/australia.py
```diff
@@ -12,10 +12,17 @@
...
-from typing import TYPE_CHECKING
-
+from datetime import date
+
+from holidays.calendars.australia_school import (
+    AUSTRALIA_SCHOOL_HOLIDAYS,
+    TERM_1_HOLIDAYS,
+    TERM_2_HOLIDAYS,
+    TERM_3_HOLIDAYS,
+    SUMMER_HOLIDAYS,
```
KJhellico:
```suggestion
    SUMMER_HOLIDAYS,
    TERM_1_HOLIDAYS,
    TERM_2_HOLIDAYS,
    TERM_3_HOLIDAYS,
```

holidays/countries/australia.py
```diff
@@ -142,6 +146,47 @@ def _move_holiday(
...
+    @staticmethod
+    def _get_school_holiday_names() -> dict[int, str]:
+        return {
+            # Term 1 School Holidays.
+            TERM_1_HOLIDAYS: tr("Term 1 school holidays"),
+            # Term 2 School Holidays.
+            TERM_2_HOLIDAYS: tr("Term 2 school holidays"),
+            # Term 3 School Holidays.
+            TERM_3_HOLIDAYS: tr("Term 3 school holidays"),
+            # Summer School Holidays.
```
KJhellico:
```suggestion
            # Term 1 school holidays.
            TERM_1_HOLIDAYS: tr("Term 1 school holidays"),
            # Term 2 school holidays.
            TERM_2_HOLIDAYS: tr("Term 2 school holidays"),
            # Term 3 school holidays.
            TERM_3_HOLIDAYS: tr("Term 3 school holidays"),
            # Summer school holidays.
```
And then need to run `make l10n` to update .po files.

tests/countries/test_australia.py
```diff
@@ -2050,6 +2052,78 @@ def test_all_holidays(self):
...
+            self.assertLessEqual(set(year_data), set(Australia.subdivisions), year)
+
+    def test_school_holidays_vic(self):
+        # Term 1 2026 runs 28 January to 2 April, so the days either side of it
+        # are holidays and the days inside it are not.
+        self.assertSubdivVicSchoolHolidayName("Summer school holidays", "2026-01-27")
+        self.assertSubdivVicSchoolHolidayName("Term 1 school holidays", "2026-04-03")
+        self.assertSubdivVicSchoolHolidayName("Term 2 school holidays", "2026-06-27")
+        self.assertSubdivVicSchoolHolidayName("Term 3 school holidays", "2026-09-19")
+        self.assertSubdivVicSchoolHolidayName("Summer school holidays", "2026-12-19")
```
KJhellico: This is actually duplicated with `test_l10n_default_school` test.

tests/countries/test_australia.py
```diff
@@ -2050,6 +2052,78 @@ def test_all_holidays(self):
...
+            "2026-01-28",
+            "2026-04-02",
+            "2026-04-20",
+            "2026-06-26",
+            "2026-07-13",
+            "2026-09-18",
+            "2026-10-05",
+            "2026-12-18",
+        ):
+            self.assertNoHoliday(vic, term_day)
```
KJhellico:
```suggestion
        self.assertNoSubdivVicSchoolHoliday(
            "2026-01-28",
            "2026-04-02",
            "2026-04-20",
            "2026-06-26",
            "2026-07-13",
            "2026-09-18",
            "2026-10-05",
            "2026-12-18",
        )
```

tests/countries/test_australia.py
```diff
@@ -2050,6 +2052,78 @@ def test_all_holidays(self):
...
+        # Each state sets its own calendar; there is no national one.
+        self.assertFalse(Australia(years=2026, categories=SCHOOL))
+
+    def test_school_holidays_nsw_uses_the_eastern_division(self):
+        # NSW runs two sets of term dates and only one can stand for the state.
+        # Eastern students return on 28 January 2027; Western would still be on
+        # holiday until 3 February.
+        self.assertSubdivNswSchoolHolidayName("Summer school holidays", "2027-01-27")
+        nsw = Australia(subdiv="NSW", years=2027, categories=SCHOOL)
+        self.assertNoHoliday(nsw, "2027-01-28", "2027-02-03")
```
KJhellico:
```suggestion
        self.assertNoSubdivNswSchoolHoliday("2027-01-28", "2027-02-03")
```

tests/countries/test_australia.py
```diff
@@ -2050,6 +2052,78 @@ def test_all_holidays(self):
...
+
+    def test_school_holidays_qld_2029_summer(self):
+        # Queensland publishes the start of the 2029 summer break but not its
+        # end, which awaits the Minister. The approved part still counts.
+        self.assertSubdivQldSchoolHolidayName("Summer school holidays", "2029-12-08")
+        self.assertSubdivQldSchoolHolidayName("Summer school holidays", "2029-12-10")
+        self.assertSubdivQldSchoolHolidayName("Summer school holidays", "2029-12-31")
+        # Term 4 2029 ends on 7 December, so that is still a school day.
+        self.assertNoHoliday(Australia(subdiv="QLD", years=2029, categories=SCHOOL), "2029-12-07")
```
KJhellico:
```suggestion
        self.assertSubdivQldSchoolHolidayName(
            "Summer school holidays",
            "2029-12-08",
            "2029-12-10",
            "2029-12-31",
        )
        # Term 4 2029 ends on 7 December, so that is still a school day.
        self.assertNoSubdivQldSchoolHoliday("2029-12-07")
```

KJhellico (commented):
@gheydon, it's a great and solid work! 👍
(And you made the right decision not to create a generator)
Just a few suggestions (mainly on code style):

gheydon:
Thank you @KJhellico — all eight applied in bab599b.

**Archived sources.** Good catch, and more relevant here than usual: of the eight departmental pages, three had already moved or started blocking while I was collecting the data (the NT and ACT pages both redirected, and the Queensland future-dates path I first tried is now a 404). Snapshots are clearly the right call.

**Id ordering.** Summer first now, which reads better anyway — it is the break the school year opens with.

**Translator comments** lowercased to match the names, and `make l10n` re-run so the .po files carry the new comments. The Thai strings are unchanged.

**Tests.** All four suggestions taken. `assertNoSubdivVicSchoolHoliday` and friends are much tidier than building the instance myself, and I hadn't realised those existed — thank you. The Victorian name test is gone: you're right that `test_l10n_default_school` asserts exactly those four names on exactly those dates, so it was doing the same work twice.

Full matrix is green on the previous commit and everything still passes here: 7,941 tests, 100% branch coverage, pre-commit clean.

And thanks for the kind words on the generator decision — it was the first thing I tried, and eight jurisdictions with eight page layouts made it more fragile than the data it produced.

PPsyrius (approved):
LGTM 🇦🇺

Great PR 👍

KJhellico (approved):
LGTM!

arkid15r (commented):
Thanks for adding this @gheydon 👍

PR #3829, opened 2026-09-18
Update iCalendar generation tool: add `{all}` placeholder to output template

### LINKED ISSUE #3732: Introduce `{all}` placeholder for `--output-template`
### **Is your feature request related to a problem? Please describe.**

https://github.com/vacanza/holidays/blob/dev/docs/examples.md demonstrates how `--output-template` can customize filenames using parameters like `--output-template "HOLIDAYS_{code}_{start_year}_{end_year}_{categories}.ics"`.

However, if a user wants to prepend or append a personal word (e.g., `"HOLIDAYS_"`), they are currently forced to hardcode the entire default naming structure (`{code}_{start_year}_{end_year}_{categories}`). Because the default filename scheme may change over time, users must manually update their scripts to stay synchronized with upstream defaults.

### **Describe the solution you'd like**

Introduce an `{all}` placeholder (or similar keyword) that automatically resolves to the current default filename. This would allow users to easily append or prepend custom text without needing to replicate or maintain the full underlying naming convention in their scripts.

**Example usage:**

```bash
holidays-ics US --years +10 --categories unofficial --output-template "PREFIX_{all}_SUFFIX.ics"

```

#### Update the documentation

Update https://github.com/vacanza/holidays/blob/dev/docs/examples.md by adding an example how to use the new `{all}` placeholder by replacing `--output-template "HOLIDAYS_{code}_{start_year}_{end_year}_{categories}.ics"` with `--output-template "HOLIDAYS_{all}.ics"`.

Kanishka-v21:
## Proposed change

Resolves #3732 by adding an {all} placeholder to the holidays-ics --output-template option.

The {all} placeholder expands to the current default output filename without the .ics extension, allowing users to add a custom prefix or suffix without having to duplicate the default filename structure.

For example:
```
holidays-ics US --years +10 --categories unofficial --output-template "HOLIDAYS_{all}.ics"
```

## Testing

- `python -m pytest tests/test_generate_ics.py` — 47 passed
- `python -m pytest` — 7946 passed
- `git diff --check` — passed

## Type of change

- [x] Existing code quality improvement (best practice, cleanup, refactoring, optimization)
- [ ] Documentation update
- [ ] Test suite update

## Checklist


docs/examples.md
```diff
@@ -713,7 +713,7 @@ holidays-ics XNYS
 Spanning the next 10 years, unofficial holidays, saved to a custom file:
 
 ```shell
-holidays-ics US --years +10 --categories unofficial --output-template "HOLIDAYS_{code}_{start_year}_{end_year}_{categories}.ics"
+holidays-ics US --years +10 --categories unofficial --output-template "HOLIDAYS_{all}.ics"
```
KJhellico, line 717: It would be better to add an example rather than replace the existing one.

tests/test_generate_ics.py
```diff
@@ -477,6 +477,19 @@ def test_output_template_default_values(self):
 
             self.assertTrue((temp_dir / "US_ALL_DEFAULT_PUBLIC.ics").exists())
 
+    def test_output_template_all(self):
+        with self.temp_cwd() as temp_dir:
```
KJhellico, line 481: You need to add other arguments besides `--years`.

docs/examples.md
KJhellico: You're still expanding on the existing example, and with a different result.

